PR #247, opened 2022-12-22
Updates to site_selection() function

Rosejoycrocker:
- updates stand-alone site selection function to work with recent changes to the guided site selection algorithms, such as new criteria and input structure.


Rosejoycrocker:
I know I said I'd also look into integrating site_selection into the ecological model, but that might take some larger structural changes so I thought I'd start with this update in a smaller PR

src/Domain.jl
```diff
@@ -392,37 +392,46 @@ end
...
+- scen_ind : scenario id for criteria (if using a scenario DataFrame, otherwise use scen_id = 1).
+- alg_ind : MCDA algorithm to use in site selection.
+
 # Returns
-Matrix : n_reps * sites * 3
-last dimension indicates: site_id, seeding rank, shading rank
+- Matrix : n_reps * sites * 3 (last dimension indicates: site_id, seeding rank, shading rank)
 """
-function site_selection(domain::Domain, criteria::DataFrame, area_to_seed::Float64, ts::Int, n_reps::Int, alg_ind::Int)
+function site_selection(domain::Domain, criteria::DataFrame, area_to_seed::Float64, ts::Int, n_reps::Int, scen_ind::Int, alg_ind::Int)
```
ConnectedSystems, line 407: I'm not clear on what `ts` and `scen_ind` are for? 

They don't seem to be used.

examples/site_selection_example.jl
```diff
@@ -0,0 +1,19 @@
+using Revise, Infiltrator
```
ConnectedSystems, line 1: Please remove unnecessary dev packages

src/Domain.jl
ConnectedSystems: Sorry, just saw `scen_ind`, but still can't see `ts`

Rosejoycrocker: ts is used to index the dhw/wave time step, I added scen_ind in case we wanted to have a scanario dataframe, generated using the sampler, and then run site selectionon scenarios from the dataframe (indicated using the scen_ind index).

ConnectedSystems: Okay I see now - hard to manually look through because a straightforward `find` (ctrl+f) on `ts` would pick up too many false matches.

How is coral state at time $ts$ incorporated, or is that out of scope of concerns in this use case?

I would split the function up so you have one that handles scenario specific cases, and a more general one that accepts the scenario spec.

e.g., something like:

```julia
function site_selection(domain, criteria, ....)
    # some set up code
    # calls the scenario specific site selection function
    for scen_criteria in eachrow(criteria)
        some_result_variable = site_selection(scen_criteria, ......)
        
        # do something with results....
    end
end

function site_selection(criteria::DataFrameRow, ....)
     # do the thing ....
end
```

Every time you use an index you're introducing the risk that the wrong index gets used, and also increase maintenance burden cos every time something changes you have to go back and update all the places that index was used and make sure it was all updated correctly.

In this function alone you have to touch 15+ lines if something changes...

Rosejoycrocker: Ok cool, I'll work on making these changes tomorrow. The current version of site_selection uses whatever is in domain.init_coral_cover for cover, but this should probably be changed to a separate input instead of using domain so it can be easily integrated with ADRIA's time dynamic model...

Rosejoycrocker: I'm just working on implementing this and I'm having issues importing the DMCDA_vars variable type into Domain - would I use "import ADRIA:dMCDA" to import this ? It also makes me think though - should site_selection be in the dMCDA.jl file instead of Domain.jl?

ConnectedSystems: > I'm having issues importing the DMCDA_vars variable type into Domain

I don't quite follow, sorry. Could you show me what you're trying to do? `dMCDA_vars` is already used inside Domain so I'm not sure how you're trying to use it.

> should site_selection be in the dMCDA.jl file instead of Domain.jl?

Probably, particularly if the function is MCDA specific. I think the current structure needs a review - too many hangovers from MATLAB. As I said, I was trying to keep things matching as much as possible to avoid the perception that I made wild changes.

Rosejoycrocker: Sorry, I forget ther's a few things called MCDA ... I mean the variable struct called MCDA_vars, I want to use it so there's less inputs in the function called in the scenario loop. But if you think it's fine to move the site selection functions to dMCDA.jl then I can do that and DMCDA_vars is already defined there so I don't have to import it.

Rosejoycrocker: actually I just realised we were using it before and it wasn't complaining about it being undefined, maybe something else is going wrong... but anyway I can move it to dMCDA.jl

Rosejoycrocker:
Hey @ConnectedSystems , I've made updates to separate site selection into a single scenario funcion and a function which loops through scenarios to do multi-scenario site selection. I couldn't work out a way of doing this without making mcda_vars a mutable struct, but if you see a way of keeping it as an immutable struct I can change it... (not sure if you're back from leave today)

ConnectedSystems:
Hi Rose, happy new year! Hope you had a good break.

Yes, I will be around today - will do a thorough review soon.

Please note that (possibly duplicate) commits for different purposes/issues not related to changes to `site_selection()` seem to be mixed into this PR. Try to keep things neat, tidy and separate please.

Rosejoycrocker:
Happy new year :) , Hope you had a good break too!

Thanks, sorry about the duplicate commits, I had an issue when I rebased to main and had to revert and then reinstate some commits.

examples/site_selection_example.jl
```diff
@@ -0,0 +1,18 @@
+using ADRIA
+using ADRIA: site_selection
```
ConnectedSystems, line 2: Here, you are importing `site_selection` explicitly, so it makes sense that calling the function directly works, but I don't see `site_selection_scens` being similarly imported, nor do I see it being exported by the ADRIA package. 

How does it work without an "undefined" error?

ConnectedSystems (changes requested):

src/sites/dMCDA.jl
```diff
@@ -691,6 +691,131 @@ function vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Union{Float64,Int64},
     return s_order
 end
 
+"""
+    site_selection_scens(domain::Domain, criteria::DataFrame, depth::DataFrame, cover::AbstractArray, area_to_seed::Float64, ts::Int, n_reps::Int)
+
+# Arguments
+- domain : ADRIA Domain type, indicating geographical domain to perform site selection over.
+- criteria : DataFrame of criteria weightings and thresholds (can be a scenario DataFrame).
```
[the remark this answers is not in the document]
ConnectedSystems: Also, a description in the docstring would help.

Rosejoycrocker:
Hey @ConnectedSystems, I think I've addressed your review points, I'm just doing changes to the tests because the function structure has changed. Sorry for holding up other issues with this.

ConnectedSystems:
Hi Rose,

> Sorry for holding up other issues with this.

Don't be sorry, I am tagging issues with dependencies so we have a better sense of priority.

Before you do anything else, please **rebase** your changes onto `main`, and update this PR so it will be merged into `main` rather than `v0.4.4`

ConnectedSystems:
Please **rebase** Rose!

Rosejoycrocker:
I used 'rebase update-site-selection-function' onto 'main' in GitKraken, is that incorrect?

ConnectedSystems:
The commit history shown here is indicating that you merged those changes into your branch. Are you skipping the force commit option?

Rosejoycrocker:
Is this when resolving conflicts?

ConnectedSystems:
No, this should be when you are pushing your commits. Let's go through it after the meeting this afternoon

Rosejoycrocker:
Okay, thanks. Sorry about this, I definitely used 'rebase onto main' in GitKraken so I'm not sure what I've done.

Rosejoycrocker:
Changes now on PR #282

Issue #238, opened 2022-12-13
Adjust definition or sampling of dist_thresh so that the value representing 'don't use distance sorting' is as likely as values representing using distance sorting


Rosejoycrocker:
- currently, distance sorting is not used in site selection if dist_thresh = 1.0, but this value is unlikely to occur with current sampling settings.
- the definition of dist_thresh or its sampling should be adjusted so that there is not a bias towards using the distance sorting algorithm in the sampling.

ConnectedSystems:
I've decided the best course of action here is to use a separate factor to indicate whether the distance sorting is "active" or "inactive". Please treat this as a priority for v0.4.6

Rosejoycrocker:
Ok cool, I'll make this a priority

PR #290, opened 2023-02-08
Add criteria parameter which indicates whether distance sorting is used.

Rosejoycrocker:
- adds new integer parameter 'use_dist' which =1 when distance sorting is to be used and 0 when distance sorting is not to be used
- addresses issue #238

### DIFF
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -106,6 +106,8 @@ Base.@kwdef struct Criteria{P,N} <: EcoModel
         name="Low Area Tolerance", description="Tolerance for low proportional space for seeding deployments.")
     deployed_coral_risk_tol::P = Param(1.0, ptype="real", bounds=(0.75, 1.0), dists="unif",
         name="Risk Tolerance", description="Filters out sites with heat/wave stress above threshold.")
+    use_dist::P = Param(0, ptype="integer", bounds=(0, 1 + 1), dists="unif",
+        name="Use Distance", description="Turns using distance sorting on and off.")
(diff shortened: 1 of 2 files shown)


src/ecosystem/Ecosystem.jl
```diff
@@ -106,6 +106,8 @@ Base.@kwdef struct Criteria{P,N} <: EcoModel
         name="Low Area Tolerance", description="Tolerance for low proportional space for seeding deployments.")
     deployed_coral_risk_tol::P = Param(1.0, ptype="real", bounds=(0.75, 1.0), dists="unif",
         name="Risk Tolerance", description="Filters out sites with heat/wave stress above threshold.")
+    use_dist::N = Param(0, ptype="integer", bounds=(0, 1 + 1), dists="unif",
+        name="Use Distance", description="Turns using distance sorting on and off.")
```
ConnectedSystems, line 110: How about "Distance Threshold" as the name?

For the description, how about: "Turns ~using~ distance sorting on **or** off"

ConnectedSystems (changes requested):
Thanks Rose, requesting a small change regarding the factor name/description

src/ecosystem/Ecosystem.jl
Rosejoycrocker: Sure, sounds good. Are these used for presenting analysis in a more understandable format?

ConnectedSystems: Yes

Rosejoycrocker: Ok, cool

ConnectedSystems (approved):

Issue #271, opened 2023-01-30
Add result interfaces for other RRAP models


ConnectedSystems:
As in title

Specific `ResultSet` objects could be defined to handle these.

PR #291, opened 2023-02-09
Add function `n_locations()` to determine number of sites/reefs/clusters

ConnectedSystems:
Previously the number of sites/reefs/clusters were being inferred by examining the number of rows in a geopackage.

In the future however, there may be several other kinds of ResultSets which interfaces with results from other RRAP models (e.g., #271), which may store this information in some other form.

This PR replaces manual examination of the geopackage with a call to a helper function. How the number of locations are determined can then be adjusted for different situations.


src/sites/dMCDA.jl
```diff
@@ -405,10 +407,10 @@ function guided_site_selection(
         rankingsin = [site_ids zeros(Int64, length(site_ids)) zeros(Int64, length(site_ids))]
     end
 
-    nsites::Int64 = length(site_ids)
+    n_sites::Int64 = length(site_ids)
 
     # if no sites are available, abort
-    if nsites == 0
+    if n_sites == 0
```
Rosejoycrocker, line 413: This isn't really a comment on your PR, but should we have a warning message when this happens?

Rosejoycrocker:
Looks good to me, I'll just run tests locally

src/sites/dMCDA.jl
ConnectedSystems: I would say no in this case because sites can be legitimately filtered out, particularly under high DHW conditions?

Rosejoycrocker: Thats true, so then we are using the previous time step's selection of sites

Rosejoycrocker (approved):
Tests all pass

src/sites/dMCDA.jl
ConnectedSystems: Ah okay. We need to discuss under what circumstances deployments should be stopped then

PR #294, opened 2023-02-10
Fix bug causing implausible juvenile indicator values

### LINKED ISSUE #292: Scenario juvenile metric returning implausible values (?) and values > 1
May be due to the maximum density value (51.8 juveniles / m²) but needs to be investigated

image

ConnectedSystems:
Closes #292

Fixes for the juvenile density indicator and standardization of some function names.


ConnectedSystems:
@Rosejoycrocker tagging you for awareness. Going to merge this directly.

PR #295, opened 2023-02-10
Adjust samples when use of distance thresholds is turned off

ConnectedSystems:
Something I missed when reviewing #290 - generated samples have to be adjusted such that `dist_thresh` is set to 0 when `use_dist` is set to 0.

### DIFF
--- a/src/io/sampling.jl
+++ b/src/io/sampling.jl
@@ -28,13 +28,16 @@ function adjust_samples(d::Domain, spec::DataFrame, df::DataFrame)::DataFrame
     df[df.guided.==-1.0, non_depth] .= 0.0
 
     # If no seeding is to occur, set related variables to 0
-    not_seeded = (df.seed_TA .== 0) .& (df.seed_CA .== 0.0)
+    not_seeded = (df.seed_TA .== 0.0) .& (df.seed_CA .== 0.0)
     df[not_seeded, contains.(names(df), "seed_")] .= 0.0
     df[not_seeded, :a_adapt] .= 0.0
 
     # Same for fogging/shading
     not_fogged = (df.fogging .== 0) .& (df.SRM .== 0)
     df[not_fogged, contains.(names(df), "shade_")] .= 0.0
+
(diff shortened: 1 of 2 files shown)


src/io/sampling.jl
```diff
@@ -28,14 +28,17 @@ function adjust_samples(d::Domain, spec::DataFrame, df::DataFrame)::DataFrame
     df[df.guided.==-1.0, non_depth] .= 0.0
 
     # If no seeding is to occur, set related variables to 0
-    not_seeded = (df.seed_TA .== 0) .& (df.seed_CA .== 0.0)
+    not_seeded = (df.seed_TA .== 0.0) .& (df.seed_CA .== 0.0)
```
Rosejoycrocker, line 31: why are the seeding variables compared to integers?

src/io/sampling.jl
```diff
@@ -28,14 +28,17 @@ function adjust_samples(d::Domain, spec::DataFrame, df::DataFrame)::DataFrame
...
     df[not_seeded, contains.(names(df), "seed_")] .= 0.0
     df[not_seeded, :a_adapt] .= 0.0
 
     # Same for fogging/shading
     not_fogged = (df.fogging .== 0) .& (df.SRM .== 0)
```
Rosejoycrocker, line 36: why are the fogging and SRM variables not compared to a float zero?

ConnectedSystems: Yeah not sure what I was seeing there, I'll change it back

ConnectedSystems: I see what happened - one condition was using an integer and the other was using a float so I just made them to match.

Corrected this so they're both now compared against integers.

Rosejoycrocker (approved):
Cool, looks good to me now :)

Rosejoycrocker:
I'll just pull and run the sampling test to be sure then merge

Rosejoycrocker:
sampling tests passing :)

PR #298, opened 2023-02-13
Additional safety checks for sampling

ConnectedSystems:
Throws error if number of samples is not a power of 2 (required by the Sobol' sampling scheme to maintain convergence properties).

Warns if duplicate scenarios are detected. This may occur due to the flooring trick necessary to map sampled values back to the expected discrete/categorical values, and due to the adjustment to avoid implausible areas of parameter space. Note: I have yet to see duplicate scenarios in practice.


src/io/sampling.jl
```diff
@@ -36,6 +38,11 @@ function adjust_samples(d::Domain, spec::DataFrame, df::DataFrame)::DataFrame
     not_fogged = (df.fogging .== 0) .& (df.SRM .== 0)
     df[not_fogged, contains.(names(df), "shade_")] .= 0.0
 
+    if nrow(unique(df)) < nrow(df)
+        perc = "$(@sprintf("%.3f", (nrow(unique(df)) / nrow(df)) * 100.0))%"
```
Rosejoycrocker, line 42: Should this be (nrow(df)-nrow(unique(df))/nrow(df)*100, because its the percentage of duplicates. For example, in set = [5 5 6 7 8 8 10], the proportion of duplicates is (7-5)/7 = 2/7. Unless I've misinterpreted the stat you want to show...

ConnectedSystems: Thanks for catching this - I'm missing the `1 - ` at the start, e.g:

$(1.0 - (3 / 10)) * 100.0$

where 3 is the number of unique items, and 10 is the total number of items.

I originally wanted to indicate % of "valid" scenarios, but then decided that knowing the proportion of duplicate scenarios was probably more informative.

Rosejoycrocker: Ok cool, looks good :)

Rosejoycrocker: if you're done with changes, I'll merge

ConnectedSystems: Yes all done! Thank you

Rosejoycrocker (approved):

PR #306, opened 2023-02-21
Adjust shelter volume to remove hardcoded index and leverage NamedDimsArray

ConnectedSystems:
Removed hardcoded position indexers to make the metric more generic with regard to the number of coral groups/size classes represented.

This PR also takes the opportunity to remove dependence on DataFrames (which are quite slow and heavy to use for this) and instead uses NamedDimsArray as much as possible.

Note: This PR also includes a fix to produce shelter volumes in m^3 / m^2

Previous implementation applied a log10-log10 relationship, as opposed to a natural log-log relationship

### DIFF
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -389,13 +389,7 @@ function _colony_Lcm2_to_m3m2(inputs::NamedDimsArray)::Tuple{Vector{Float64},Vec
 
     # Extract assumed colony area (in cm^2) for each taxa/size class from scenario inputs
     # Have to be careful to extract data in the correct order, matching coral id
-    local colony_area_cm2::Array{Float64}
-    try
-        colony_area_cm2 = Array{Float64}(inputs[:, cs_p.coral_id.*"_colony_area_cm2"])'
-    catch
-        # Get from NamedDimsArray instead
-        colony_area_cm2 = Array{Float64}(inputs(cs_p.coral_id .* "_colony_area_cm2"))
-    end
+    colony_area_cm2::Array{Float64} = Array{Float64}(inputs(cs_p.coral_id .* "_colony_area_cm2"))
 
     # Colony planar area parameters (see second column of Table 1 in Urbina-Barreto et al., [1])
     # First column is `b`, second column is `a`
@@ -416,11 +410,7 @@ function _colony_Lcm2_to_m3m2(inputs::NamedDimsArray)::Tuple{Vector{Float64},Vec
     log_colony = pa_params[:, 1] .+ pa_params[:, 2] .* log.(colony_area_cm2)
 
     # Maximum colony area for each species and scenario, using largest size class
-    if ndims(colony_area_cm2) == 1
-        max_log_colony = pa_params[n_sizes:n_sizes:end, 1] .+ pa_params[n_sizes:n_sizes:end, 2] .* log.(colony_area_cm2[n_sizes:n_sizes:end])
-    else
-        max_log_colony = vec(pa_params[n_sizes:n_sizes:end, 1] .+ pa_params[n_sizes:n_sizes:end, 2] .* log.(colony_area_cm2[n_sizes:n_sizes:end, :]))
-    end
+    max_log_colony::Vector{Float64} = pa_params[n_sizes:n_sizes:end, 1] .+ pa_params[n_sizes:n_sizes:end, 2] .* log.(colony_area_cm2[n_sizes:n_sizes:end])
 
     colony_litres_per_cm2::Vector{Float64} = 10.0 .^ log_colony
     max_colony_litres_per_cm2::Vector{Float64} = 10.0 .^ max_log_colony
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -381,10 +381,11 @@ Tuple : Assumed colony volume (m³/m²) for each species/size class, theoretical
    https://doi.org/10.1016/j.ecolind.2020.107151
 
 """
-function _colony_Lcm2_to_m3m2(inputs::Union{DataFrame,DataFrameRow})::Tuple
+function _colony_Lcm2_to_m3m2(inputs::Union{DataFrame,NamedDimsArray})::Tuple
     _, _, cs_p::DataFrame = coral_spec()
     n_corals::Int64 = length(unique(cs_p.taxa_id))
     n_species::Int64 = length(unique(cs_p.coral_id))
+    n_sizes::Int64 = length(unique(cs_p.class_id))
 
     # Extract assumed colony area (in cm^2) for each taxa/size class from scenario inputs
     # Have to be careful to extract data in the correct order, matching coral id
@@ -392,8 +393,8 @@ function _colony_Lcm2_to_m3m2(inputs::Union{DataFrame,DataFrameRow})::Tuple
     try
         colony_area_cm2 = Array{Float64}(inputs[:, cs_p.coral_id.*"_colony_area_cm2"])'
     catch
-        # Get from DataFrameRow instead
-        colony_area_cm2 = Array{Float64}(inputs[cs_p.coral_id.*"_colony_area_cm2"])
+        # Get from NamedDimsArray instead
+        colony_area_cm2 = Array{Float64}(inputs(cs_p.coral_id .* "_colony_area_cm2"))
     end
 
     # Colony planar area parameters (see second column of Table 1 in Urbina-Barreto et al., [1])
@@ -541,7 +542,7 @@ shelter volume (a 3D metric).
    Ecological Indicators, 121, 107151.
    https://doi.org/10.1016/j.ecolind.2020.107151
 """
-function _absolute_shelter_volume(X::AbstractArray{T,4}, site_area::Vector{T}, inputs::DataFrame)::AbstractArray{T} where {T<:Real}
+function _absolute_shelter_volume(X::AbstractArray{T,4}, site_area::Vector{T}, inputs::NamedDimsArray)::AbstractArray{T} where {T<:Real}
     nspecies::Int64 = size(X, :species)
 
     # Calculate shelter volume of groups and size classes and multiply with area covered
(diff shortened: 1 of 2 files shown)


Rosejoycrocker (commented):
It looks good to me but I'm getting an error in the Relative Shelter Volume Test:

`caused by: MethodError: no method matching NamedDimsArray(::DataFrame; scenarios=1:1, factors=["RCP", "dhw_scenario", "wave_scenario", "guided", "seed_TA", "seed_CA", "fogging", "SRM", "a_adapt", "n_adapt"  …  "corymbose_acropora_enhanced_3_3_wavemort90", "tabular_acropora_unenhanced_2_6_wavemort90", 
"corymbose_acropora_unenhanced_4_5_colony_area_cm2", "corymbose_acropora_unenhanced_4_2_wavemort90", "small_massives_5_1_bleaching_sensitivity", "tabular_acropora_enhanced_1_6_bleaching_sensitivity", "tabular_acropora_unenhanced_2_5_mb_rate", "tabular_acropora_enhanced_1_4_colony_area_cm2", "tabular_acropora_unenhanced_2_6_fecundity", "small_massives_5_3_fecundity"])
  Closest candidates are:
    NamedDimsArray(::AbstractVector, ::Symbol) at C:\Users\rcrocker\.julia\packages\NamedDims\pu9Bn\src\wrapper_array.jl:46 got unsupported keyword arguments "scenarios", "factors"
    NamedDimsArray(::KeyedArray{T, N, <:NamedDimsArray{L, T, N, A} where A<:AbstractArray{T, N}} where {T, N}, ::Tuple) where L at C:\Users\rcrocker\.julia\packages\AxisKeys\qzXul\src\names.jl:60 got unsupported keyword arguments "scenarios", "factors"
    NamedDimsArray(::AbstractArray; kw...) at C:\Users\rcrocker\.julia\packages\AxisKeys\qzXul\src\names.jl:149
    ...`

Rosejoycrocker:
I'm not sure if the test just needs to be altered or it's due to an error in the relative shelter volume function

ConnectedSystems:
Hmmm, let me double check (soon). Tests ran fine last night so not sure what the issue might be.

Just in case, could you let me know package versions for NamedDims/AxisKeys?

```julia
(ADRIA) pkg> status NamedDims AxisKeys
Project ADRIA v0.4.5
Status `C:\projects\AIMS\ADRIA.jl\Project.toml`
  [94b1ba4f] AxisKeys v0.2.12
  [356022a1] NamedDims v1.2.1
```

Rosejoycrocker:
Weird, just checked package versions and they're the same as the ones above.

ConnectedSystems:
That's good - I just wanted to rule that out as an issue.

ConnectedSystems:
I stashed some changes to switch between branches and forgot about them. 

Tests all pass now.

ConnectedSystems:
Hold off on this Rose, I think I spotted a bug in the math

ConnectedSystems:
Fixed a mismatch between coral taxa and their associated parameters.

To the slide Ken showed in his recent email.

image

It's a bit hard to see, but the legend seems to give "86" as the maximum shelter volume per ha. This is the mean value over 2000-2020 (SSP2 - RCP4.5).

If we assume a single site hits this value, and does so for all six taxa (86 * 6 = 516) then ADRIA seems to overestimate by 2 to 3 times.

Below I'm showing results for the first site, first scenario.
image

@Rosejoycrocker could be the values we used for the shelter volume model (assumed values based on the Urbina-Barretto paper) are inappropriate. Do you have time to have a look?

Rosejoycrocker:
Hey @ConnectedSystems , I'll have a look this afternoon :)

ConnectedSystems:
@Rosejoycrocker  Looking over the paper over lunch - should we be using:

`exp(b + a * log(PA))` to convert the values back to cubic decimeter?

@KRNA01 We use the default value of 2827.43 cm^2 for all coral types and perturbe these +/- 10% (2544.69, 3110.18) over a triangular distribution

Judging by the x-axis scale, are we severely under-representing the range of possible colony sizes?

image

Rosejoycrocker:
Hey @ConnectedSystems, is that figure from Urbina-Barreto 2021? I can't seem to find it...

Rosejoycrocker:
Sorry, ignore- I found it. I forgot there are 2 2021 papers

ConnectedSystems:
> Sorry, ignore- I found it. I forgot there are 2 2021 papers

Great. Also note that the relevant references are listed in the docstrings

Rosejoycrocker:
Hey @ConnectedSystems, I think we do the transformation here : https://github.com/open-AIMS/ADRIA.jl/blob/347b3fc420f688dcee3efd9d43c6d2e8e793d2bc/src/metrics/metrics.jl#L438

But I'm not sure if it should be exp... I imagine the paper is using `log_{10}` because it's not using `ln()` to denote log but hard to tell

Rosejoycrocker:
Just checking- the screenshot above is in m^3/ha ?

ConnectedSystems:
Yes, see screenshot of output underneath it - I've converted values to match

KRNA01:
Hi @ConnectedSystems and @Rosejoycrocker,

Thanks for double-checking 

> Fixed a mismatch between coral taxa and their associated parameters.
> 
> To the slide Ken showed in his recent email:
> 
> image
> 
> It's a bit hard to see, but the legend seems to give "86" as the maximum shelter volume per ha. This is the mean value over 2000-2020 (SSP2 - RCP4.5).
> 
> If we assume a single site hits this value and does so for all six taxa (86 * 6 = 516) - a highly optimistic scenario - then ADRIA seems to overestimate by 3 to 4 times under a non-intervention context.
> 
> Below I'm showing results for the first site (Briggs_BR_C_1), first scenario. image
> 
> @Rosejoycrocker could be the values we used for the shelter volume model (assumed values based on the Urbina-Barretto paper) are inappropriate. Do you have time to have a look?

Note that these figures are produced using ReefMod outputs.

KRNA01:
@ConnectedSystems and @Rosejoycrocker.

Great that you are double checking the math.  I was unsure about the log natural versus log 10 scale.

I even emailed Urbina-Barreto to get confirmation on what scale was used, but never got a reply.

From memory, I passed the challenge to Ryan. 

@Rosejoycrocker, if Ryan has firmed up this decision, can you think of a test we can do to establish which scale is the right one? 

Also agree that the size ranges used in the study very much under-represent our ranges.

This, the scaling issue, and uncertainty around whether vertical relief (space under corals) were captured as part of the study makes me think we really need ecoRRAP to help redo this for at least the six groups in ReefMod.

ConnectedSystems:
Does using `exp()` look reasonable to you Rose?
```
a = 1.50
b = -8.32
colony_area = 3085.1855

b + a * log(colony_area)
# 3.731550596287125

10^(b + a * log(colony_area))
# 5389.526309135325

exp(b + a * log(colony_area))
# 41.7437
```

image

Rosejoycrocker:
Are we missing a conversion here?
https://github.com/open-AIMS/ADRIA.jl/blob/347b3fc420f688dcee3efd9d43c6d2e8e793d2bc/src/metrics/metrics.jl#L443

We convert from dm^3 to m^3 by multiplying by 10^-3 but the RHS unit label is m^3/m^2, while `colony_litres_per_cm2` is in per cm^2 units, so shouldn't be also divide by 10^-4 to convert to per m^2. Have I missed something? (not that this will effect the final size much)

Rosejoycrocker:
The value with exp definitely seems more reasonable, I'll see if I can check it against the plot scale

Rosejoycrocker:
I just realised, `log()` in julia is the natural log, so which ever it is we have it wrong in the code because we are doing the inverse of $log_{10}$ on the natural log

ConnectedSystems:
> Are we missing a conversion here?

That's converting dm^3 to m^3.

dm^3 is equivalent to Litres/cm^2 in this context (because dm^3 = 1 litre and the model expects the planar area to be given in cm^2).

Converting to m^3 means multiplying by 0.001 and doing so gets us the unit per m^2.

That's how I understand it at least.

KRNA01:
@Rosejoycrocker, one way to test this is to produce estimates of SV (in litres or m3) per m2 for the largest corymbose Acropora (and maybe also a large Tabular Acropora) using both the natLog scale and the 10log scale.  Assume 100% coral cover.

I can then quickly gauge which volume is the right one (produced by which scale) by going back to first principles (i.e. knowing corals like the inside of my pocket).

KRNA01:
@Rosejoycrocker.
Good pick-up. Same in MatLab.

ConnectedSystems:
@KRNA01 the values used in the [example above](https://github.com/open-AIMS/ADRIA.jl/pull/306#issuecomment-1439407552) are for Tabular Acropora (largest size class).

Not sure what you think, but these look wrong to me:

```
b + a * log10(colony_area)
# -3.0860780776546584

10^(b + a * log10(colony_area))
# 0.0008202040744216785
```

Rosejoycrocker:
> > Are we missing a conversion here?
> 
> That's converting dm^3 to m^3.
> 
> dm^3 is equivalent to Litres/cm^2 in this context (because dm^3 = 1 litre and the model expects the planar area to be given in cm^2).
> 
> Converting to m^3 means multiplying by 0.001 and doing so gets us the unit per m^2.
> 
> That's how I understand it at least.

> > Are we missing a conversion here?
> 
> That's converting dm^3 to m^3.
> 
> dm^3 is equivalent to Litres/cm^2 in this context (because dm^3 = 1 litre and the model expects the planar area to be given in cm^2).
> 
> Converting to m^3 means multiplying by 0.001 and doing so gets us the unit per m^2.
> 
> That's how I understand it at least.

Right, do you mean 'the model expects planar area to be in cm^2' as in the x value in the interpolation is planar area?

ConnectedSystems:
> Right, do you mean 'the model expects planar area to be in cm^2' as in the x value in the interpolation is planar area?

I'm probably misunderstanding but aren't we saying the same thing? 
By "the model" I'm referring to the model in Urbina-Barreto et al.

KRNA01:
@ConnectedSystems and @Rosejoycrocker 

I managed to find the code for the Urbina-Barreto paper, including the R script.

Here's the link: https://zenodo.org/record/4309245

ConnectedSystems:
Okay, so `ln(x)` to output `ln(y)`

So to transform back to cubic decimeter, we take apply `exp()`?

Rosejoycrocker:
Yeah sounds good, looks like they use natural log

KRNA01:
Good work, team !

ConnectedSystems:
Absolute SV look much more reasonable now

Showing per ha values below:

image

KRNA01:
@ConnectedSystems, @Rosejoycrocker 

Great, are these in units of m3 per ha?

ConnectedSystems:
> Great, are these in units of m3 per ha?

Indeed

ConnectedSystems:
@Rosejoycrocker ready for (final?) review

Rosejoycrocker:
Yep, I can review

Rosejoycrocker (approved):
Looks good to me and all tests passing on my end :) If you've thought of anything else let me know but otherwise I'll merge

ConnectedSystems:
Nope, though speaking of tests I should add one to ensure values are in expected ranges. Best to do this after the improved example dataset is ready though I think, so please go ahead and merge.

Rosejoycrocker:
Ok, cool, sounds good

ConnectedSystems:
@KRNA01 for completeness, could you respond with your approval as well?

KRNA01:
@ConnectedSystems , @Rosejoycrocker 
Great, thank you both for completing this.

PR #310, opened 2023-02-27
Correct issue with size of NamedDims array for storing ranks in `run_site_selection()`

### LINKED ISSUE #308: `run_site_selection()` errors out
Example script I am using

```julia
using Revise, Infiltrator
using ADRIA


dom = ADRIA.load_domain("path to Moore dataset")
scens = ADRIA.sample_site_selection(dom, 8)

area_to_seed = 1.5 * 10^-6  # area of seeded corals in km^2
ts = 5  # time step to perform site selection at

# initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites))
sum_cover = repeat(sum(dom.init_coral_cover, dims=2)', size(scens, 1))

ranks = ADRIA.run_site_selection(dom, scens, sum_cover, area_to_seed, ts)
```

Errors out with a `BoundsError: attempt to access Tuple{Symbol, Symbol} at index [3]`
Seems it is trying to index a non-existent third dimension for `sum_cover`?
Once I correct for that, it runs into another error where it seems to be trying to select all sites from `sum_cover` (L 428 in `guided_site_selection()`)

Is there a way of performing site selection using available data across time, rather than specifying a specific time step?
This would be the preferred approach for a "set and forget" scenario I think (i.e., select sites based on available long-term projections rather than what's forecasted for the year).

The example needs to be updated and moved into the documentation as well.

Rosejoycrocker:
Closes #308, where the storage NamedDimsArray for ranks in `run_site_selection()` was set to n_site_int instead of n_sites (which did not throw an error in tests because the filtered example site set has 5 sites).

Also adds proper indexing for site storage which accounts for filtered sites.

### DIFF
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -794,7 +794,7 @@ function run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::Abst
     ranks_store = NamedDimsArray(
         zeros(nrow(criteria), length(domain.site_ids), 3),
         scenarios=1:nrow(criteria),
-        sites=1:length(domain.site_ids),
+        sites=domain.site_ids,
         ranks=["site_id", "seed_rank", "shade_rank"],
     )
 
@@ -818,7 +818,7 @@ function run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::Abst
             sum_cover[cover_ind, :, :],
             area_to_seed
         )
-        ranks_store[cover_ind, depth_priority, :] = ranks_temp
+        ranks_store(scenarios=cover_ind, sites=domain.site_ids[depth_criteria]) .= ranks_temp
     end
 
     return ranks_store
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -815,7 +815,7 @@ function run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::Abst
             wave_scens[timestep, :, criteria.wave_scenario[cover_ind]],
             dhw_scens[timestep, :, criteria.dhw_scenario[cover_ind]],
             depth_priority,
-            sum_cover[cover_ind, :, :],
+            sum_cover[cover_ind, :],
             area_to_seed
         )
         ranks_store(scenarios=cover_ind, sites=domain.site_ids[depth_criteria]) .= ranks_temp
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -818,7 +818,7 @@ function run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::Abst
             sum_cover[cover_ind, :, :],
             area_to_seed
         )
-        ranks_store[cover_ind, 1:size(ranks_temp, 1), :] = ranks_temp
+        ranks_store[cover_ind, depth_priority, :] = ranks_temp
     end
 
     return ranks_store
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -775,13 +775,13 @@ function vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Union{Float64,Int64},
 end
 
 """
-    run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::AbstractArray, area_to_seed::Float64, time_step::Int64)
+    run_site_selection(domain::Domain, scenarios::DataFrame, sum_cover::AbstractArray, area_to_seed::Float64, time_step::Int64)
 
 Perform site selection for a given domain for multiple scenarios defined in a dataframe.
 
 # Arguments
 - `domain` : ADRIA Domain type, indicating geographical domain to perform site selection over.
-- `criteria` : DataFrame of criteria weightings and thresholds.
+- `scenarios` : DataFrame of criteria weightings and thresholds for each scenario.
 - `sum_cover` : array of size (number of scenarios * number of sites) containing the summed coral cover for each site selection scenario.
 - `area_to_seed` : area of coral to be seeded at each time step in km^2
 - `timestep` : time step at which seeding and/or shading is being undertaken.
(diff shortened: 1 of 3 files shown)


src/sites/dMCDA.jl
```diff
@@ -818,7 +818,7 @@ function run_site_selection(domain::Domain, criteria::DataFrame, sum_cover::Abst
             sum_cover[cover_ind, :, :],
             area_to_seed
         )
-        ranks_store[cover_ind, 1:size(ranks_temp, 1), :] = ranks_temp
+        ranks_store[cover_ind, depth_priority, :] = ranks_temp
```
ConnectedSystems, line 821: Is there a potential for mismatch here as `depth_priority` relies on index position?

Also note that ReefMod doesn't use depth, so when developing the ReefMod interface I simply set depth_priority to `1:nrow(site_data)` - need a better way of handling that case

Rosejoycrocker: Yeah there is, I'd have to change things in site_selection though because I think the set is depth filtered first before being ranked, so the set that comes out has the size of the filtered set rather than the full set. I can look into it.

Rosejoycrocker: On reviewing, I think the main way a mismatch would occur is if the site_data had duplicate values in site_ids, which it shouldn't. I've simplified the definition of depth_priority a bit anyway (using findall instead of the vector of ids), so it should correspond directly to the site data.

ConnectedSystems: I may be in need of a second coffee but I'm not entirely sure how using `findall` helps here. I don't think it hurts, but I can't see how it is advantageous.

Rosejoycrocker: It's just neater, no difference

Rosejoycrocker: But I'm not sure how else to index here, as `depth_priority` is what is used inside `site_selection`. I suppose we could use `ranks_temp[:,1]` but it should be the same

ConnectedSystems: I see what you mean re neatness.

Here's my understanding of how this is structured:

`depth_priority` indexes the full site list by position (the variable name could be clearer, but ignore that for the moment).

Inside `site_selection()`, it is used to subset the full site list again.

This is then used to index `ranks_store`.

This will work (has been working?) so long as nothing gets accidentally reordered, the index positions correctly map back to the full index, and there are no bugs when slicing arrays. But there are several places where arrays are indexed/subsetted etc.

Given `ranks_store` is a NamedDimsArray, you could start by using the actual location names/site ids as the index keys. This will still work with positional indexing. Then, you could change the positional indexing to use location names/site ids. This will protect from mismatches in a more robust way than simply relying on matching array sizes and positions.

Rosejoycrocker: Let me know if what I just pushed is better

ConnectedSystems: Looks like a good start. Does the example run without issue now?

FYI, I was puzzled as to why GitHub wasn't auto-linking the issue you mentioned and it seems "addresses" is not a closing keyword: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

Rosejoycrocker: Sorry I forgot that was the main issue here ... I looked into it again and I think maybe we're working with different definitions here - the way I've set it up is so that coral cover has dimensions (:scenarios,:sites,:species), to allow the possibility of having different initial coral covers for different scenarios. In the example you sent coral_cover has dimensions (:sites,:species)... I can make it a single coral cover layer if you think thats too complicated.

Also, apologies for not having a key word, I didn't realise you had to have that as well as the issue hash.

ConnectedSystems: > apologies for not having a key word, I didn't realise you had to have that as well as the issue hash.

Meanwhile, I thought "addresses" was a keyword

> to allow the possibility of having different initial coral covers for different scenarios. In the example you sent coral_cover has dimensions (:sites,:species)

Sure but the example above is based on your example (in the examples folder, which should eventually be moved into the docs)

Rosejoycrocker: Right, I must have updated the test but not the example (weird, I thought I had). Should I keep it the way it is and have multiple initial coral covers? Or do you reckon one is best?

ConnectedSystems: You could support both, but to keep things simple for now, let's have just one.

ADRIA uses the mean connectivity values (with a view to move towards a stochastic representation).
ReefMod has multiple connectivity years, but I'm just taking the mean of those for now.

Rosejoycrocker: Ok cool, also what was the example script called that you are working from?

ConnectedSystems: `examples/site_selection.jl`

Rosejoycrocker: Did you change the example? it looks different from the one in Main and site-selection-sampling-error
https://github.com/open-AIMS/ADRIA.jl/blob/0808ba70cb4f1e3192f34408dfcfc62a407a5245/examples/site_selection.jl#L16

Rosejoycrocker: Anyway, I'll change it to what you have

ConnectedSystems: Yes, I am using a modified version of the example to work on the ReefMod interface.

In the example you linked, you're creating a matrix of $N \cdot S$, where $N$ is the number of scenarios and $S$ is the number of locations.

This line: `sum_cover = repeat(sum(dom.init_coral_cover, dims=2)', size(scens, 1))`

Does the same, using the initial coral cover data loaded for the domain

Rosejoycrocker: When I infiltrated it seemed like it was scenarios*species (8 by 36), but scenarios was labelled with sites. What I've done is keep it so we can still have different coral cover scenarios, but it sums over species instead.. There was also an issue with it trying to index species, I think previously the summation was done in site_selection() but we took it out (I'm not sure why it didn't throw an error in tests, but I'll add a test for mcda_vars() to check the components are the correct size

Rosejoycrocker: Ok tests are passing and example works now

Rosejoycrocker: I think this is ready for final review when you have time @ConnectedSystems . Tests passing on my end.

ConnectedSystems (approved):

PR #325, opened 2023-03-27
Add capability to fix factors in place

ConnectedSystems:
Allow users to easily fix factors in place (i.e., make them constant).

Note: No resetting is offered yet as what it means for a user to "reset" a domain/model is not entirely clear (revert everything including "sim constants"? Revert just the fixed factors?

For now, if the user wants to start fresh they can simply reload the domain.


ConnectedSystems:
pinging @Rosejoycrocker for awareness, no review necessary.

PR #348, opened 2023-05-17
Add option to use JMcDM functions in ADRIA site selection

### LINKED ISSUE #242: Consider using JMcDM
Provides access to a wide range of MCDA methods

https://github.com/jbytecode/JMcDM

Rosejoycrocker:
Adds option to use either native ADRIA mcda aggregation functions or one of 18 available methods from the JMcDM package. Also adds function `retrieve_ranks` to allow transitioning between JMcDM methods and ADRIA methods while running ADRIA.

Closes #242


Rosejoycrocker:
Note: I have also reverted the wave_stress criteria normalisation back to min-max as the max normalisation was giving such a small range of values, everything was getting filtered out during testing. This is just a placeholder for whatever we decide to do with this.

src/sites/dMCDA.jl
```diff
@@ -182,12 +205,47 @@ function rank_sites!(S, weights, rankings, n_site_int, mcda_func, rank_col)::Tup
...
+    retrieve_ranks(S::Matrix, scores::Vector, rev_val::Bool, site_ids::Vector)
+
+Get location ranks using mcda technique specified in mcda_func, weights and a decision matrix S.
+
+# Arguments
+- `S` : decision matrix containing criteria values for each location (n locations)*(m criteria)
+- `weights` : importance weights for each criteria. 
+- `mcda_func` : function/[function bool] array to use for mcda, specified as an element from method.
+- `site_ids` : array of site ids still remaining after filtering.
+- `scores` : set of scores derived from applying an mcda ranking method.
```
ConnectedSystems, line 220: Orphaned arguments in argument list?

src/sites/dMCDA.jl
```diff
@@ -182,12 +205,47 @@ function rank_sites!(S, weights, rankings, n_site_int, mcda_func, rank_col)::Tup
...
+- `rev_val` : Boolean indicating whether a mcda method is maximising score (true), or minimising (false). 
+
+# Returns
+- `s_order` : [site_ids, criteria values, ranks]
+"""
+function retrieve_ranks(S::Matrix, weights::Vector{Float64}, mcda_func::Function, site_ids::Vector)
+    S = mcda_normalize(S)
+    S .= S .* repeat(weights', size(S, 1), 1)
```
ConnectedSystems, line 228:
```
S = mcda_normalize(S)
S .= S .* repeat(weights', size(S, 1), 1)
```

Can simply be:

`S = mcda_normalize(S) .* weights'`

src/sites/dMCDA.jl
```diff
@@ -485,20 +545,10 @@ function guided_site_selection(
...
-        mcda_func = vikor
-    else
-        error("Unknown MCDA algorithm selected. Valid options are 1 (Order Ranking), 2 (TOPSIS) and 3 (VIKOR).")
-    end
-
     if log_seed && isempty(SE)
         prefseedsites = zeros(Int64, n_site_int)
     elseif log_seed
-        prefseedsites, s_order_seed = rank_seed_sites!(SE, wse, rankings, n_site_int, mcda_func)
+        prefseedsites, s_order_seed = rank_sites!(SE, wse, rankings, n_site_int, mcda_func, 2)
```
ConnectedSystems, line 551: What do the "2" and "3" mean when used as an argument to `rank_sites!()`?

src/sites/dMCDA.jl
```diff
@@ -182,12 +205,47 @@ function rank_sites!(S, weights, rankings, n_site_int, mcda_func, rank_col)::Tup
...
+    S = mcda_normalize(S)
+    S .= S .* repeat(weights', size(S, 1), 1)
+    scores = mcda_func(S)
+
+    return retrieve_ranks(S, vec(scores), true, site_ids)
 end
-function rank_shade_sites!(S, weights, rankings, n_site_int, mcda_func)::Tuple{Vector{Int64},Matrix{Union{Float64,Int64}}}
-    rank_sites!(S, weights, rankings, n_site_int, mcda_func, 3)
+function retrieve_ranks(S::Matrix, weights::Vector{Float64}, mcda_func::Vector{Any}, site_ids::Vector)
+    fns = repeat([maximum], length(weights))
```
ConnectedSystems, line 234: Use `fill(maximum, length(weights))`

Generally, `repeat()` is slower and uses more memory. It's just to make it easier when you want multiple complicated data structures.

src/sites/dMCDA.jl
```diff
@@ -182,12 +205,47 @@ function rank_sites!(S, weights, rankings, n_site_int, mcda_func, rank_col)::Tup
...
+    results = mcdm(MCDMSetting(S, weights, fns), mcda_func[1])
+
+    return retrieve_ranks(S, results.scores, mcda_func[2], site_ids)
 end
+function retrieve_ranks(S::Matrix, scores::Vector, rev_val::Bool, site_ids::Vector)
```
ConnectedSystems, line 239: I wrote the below and then realised, you probably should be using a NamedDimsArray+AxisKeys here, then you don't need `site_ids` and there's less risk that the order is wrong.

---

It's better practice to put boolean arguments at the end.

Tell me how you read this argument list, because to me it looks like:

"retrieve ranks for elements in this matrix, associated with these scores, reverse the value, and oh by the way here are some site ids"

If it's `retrieve_ranks(S, site_ids, scores, rev_val)`, then (to me) it reads like:

"retrieve ranks for elements in `S` in order of these `site_ids`, that have these scores, and reverse the list"

src/sites/dMCDA.jl
```diff
@@ -182,12 +205,47 @@ function rank_sites!(S, weights, rankings, n_site_int, mcda_func, rank_col)::Tup
...
+
+    s_order = Union{Float64,Int64}[Int.(site_ids) scores 1:size(S, 1)]
```
ConnectedSystems, line 241: Use concrete types where possible please: `Int` ->`Int64`

And you can consolidate this line with Line 243 to avoid reallocations.

src/sites/dMCDA.jl
```diff
@@ -3,6 +3,33 @@
 using StatsBase
 using Distances
 using Combinatorics
+using JMcDM
+using ADRIA: order_ranking, adria_vikor, adria_topsis
+
+global method = [
```
ConnectedSystems, line 9: Globals are bad! Mark this a const for the module please.

Also, isn't there a way of getting a list of available methods from JMcDM?

ConnectedSystems (changes requested):
Partial review, will continue on tomorrow.

src/sites/dMCDA.jl
Rosejoycrocker: which one?

src/sites/dMCDA.jl
Rosejoycrocker: These are the column used for storage in rankings for seeding and shading. In later changes (where mcda_vars , make_decision_matrix and make_seed_matrix/make_shade_matrix are replaced with more generic constructors) this is removed because rankings becomes generic. I thought I'd leave this change for the branch where I'm changing mcda_vars etc.

src/sites/dMCDA.jl
Rosejoycrocker: sorry, I recall you saying this now. It's a Matlab habit.

src/sites/dMCDA.jl
Rosejoycrocker: This is another change that I made whern I genericised the construction of S - S is a NamedDims Array instead of a Matrix

src/sites/dMCDA.jl
Rosejoycrocker: The list of available methods wasn't working for me when I tried previously. I just tried again and it's working now, so I'll add it.

Rosejoycrocker: Actually, sorry I remember now why I used the global (which I can replace with a constant), it's because I needed to add something which indicates whether the method gives best sites as the highest score or lowest score (I add a Boolean next to the method) and I wasn't sure how to do this without manually looking at the docs for the method and working it out. I'll have a look again and see if there is something which indicates but otherwise not sure how to replace.

Rosejoycrocker: So looks like the results struture also has a "best index" field, so I can add somethng which checks whether the best index is equal to the max or min score and order the sites accordingly :)

Rosejoycrocker: So I think the reason the list of JMcDM functions wasn't working for me before is it uses the subtypes() function (in subtypes(MCDMMethod)) which is automatically loaded in interactive mode but doesn't seem to be loaded when running ADRIA. Due to this it throws an error saying it can't find the subtypes() function. Is it ok to include the InteractiveUtils module in dMCDA or is this not a good idea? If so, is there a way to just load one function from the module?

src/sites/dMCDA.jl
ConnectedSystems: This isn't a sustainable approach - we don't want you (and more importantly YOU won't want) to come back and add new numbers and `elseifs` every time a new criteria (or model with new criteria) gets added.

I suggest leaving this as is for now, we'll work out a more manageable approach later.

src/sites/dMCDA.jl
ConnectedSystems: Read the argument list and then compare with the function signatures given at the top, and the function signature that is implemented.

src/sites/dMCDA.jl
ConnectedSystems: No, don't add `InteractiveUtils` to our list of dependencies.

This should be fixed in JMcDM.

Could you give me a link to where this is being used in JMcDM please?

src/sites/dMCDA.jl
Rosejoycrocker: Yeah, as I said, in other changes I've made this is removed, but it requires other changes which are probably best done as part of another issue.

src/sites/dMCDA.jl
Rosejoycrocker: This is where it's used- under the mcdm subtitle https://jbytecode.github.io/JMcDM/dev/utility/

It works fine when I use it in the command prompt, but when I brekpoint in dMCDA in ADRIA with "using JMcDM " in the file, it throws an error for subtypes() for me.

src/sites/dMCDA.jl
ConnectedSystems: Okay, great thank you.

src/sites/dMCDA.jl
ConnectedSystems: Hi Rose, spoke to the maintainer of JMcDM and he's of the opinion that use of `subtypes()` was for illustrative purposes only and not a supported core feature of the package.

This means we're forced to add InteractiveUtils as a dependency.
After adding it to the project, please also add the below somewhere, either in the main `ADRIA.jl` file or in the file where it is used. I think the latter is preferable given we don't use it anywhere else.

```
using InteractiveUtils: subtypes
```

Rosejoycrocker: Hi Takuya, thanks for checking up on this. Maybe if we add JMcDM as an extension later then we will only have to add the dependency if the extension is used. For now I'll add to dependecies as suggested.

Rosejoycrocker: Hi Takuya, so I've been working on adding this now and it seems like the type of the elements of `subtypes(MCDMMethod)` are 'DataType` so they aren't actually functions and cause an error if you try and run the JMcDM methods with them. Do you know a way of getting the associated function from the datatype?

ConnectedSystems: Shouldn't matter (beauty of multiple dispatch). Calling the `DataType` (e.g., `SomeDataType()`) should resolve to the corresponding constructor.

If this is an issue with defining an argument type for a function, then I suggest to just use `Union{Function, DataType}`.

ConnectedSystems: Incidentally, if your implementations of TOPSIS/VIKOR/Order Ranking are equivalent to those in JMcDM, then I highly encourage you to submit a PR there to replace their slower implementations. Then we don't need to maintain functionally equivalent code.

Rosejoycrocker: > Shouldn't matter (beauty of multiple dispatch). Calling the `DataType` (e.g., `SomeDataType()`) should resolve to the corresponding constructor.
> 
> If this is an issue with defining an argument type for a function, then I suggest to just use `Union{Function, DataType}`.

It's fine going into the function (I changed the input to DataType), but it errors when used with the JMcDM mcda() function, it complains it's a DataType not a JMcDM mcda function. I couldn't find anything on the JMcDM docs, they don't actually use the list to implement a method in the example. I'm looking at it again this morning, so I'll look at the docs again and see if I can find anything.

Rosejoycrocker: > Incidentally, if your implementations of TOPSIS/VIKOR/Order Ranking are equivalent to those in JMcDM, then I highly encourage you to submit a PR there to replace their slower implementations. Then we don't need to maintain functionally equivalent code in ADRIA.

I would have to make changes because my functions don't allow for combinations of Max and min in the criteria but I can definitely do this at some point.

Rosejoycrocker: When the JMcDM function is used in this line :
https://github.com/open-AIMS/ADRIA.jl/blob/b119fa91911edd4a957cfa859197df82f1219aed/src/sites/dMCDA.jl#L216

It should be something like `mcda_func=VikorMethod()`, but the entries from subtypes(JMcDM.MCDMMethod) error.

Rosejoycrocker: Note, I've also tried `methods(mcda_func)[1]` to extract the method but because it's a general method and the mcda function requires a `JMcDM.MCDMMethod` type it errors.

ConnectedSystems: https://github.com/open-AIMS/ADRIA.jl/blob/ea5762f5b1f050a1f031ef9a538acb3741237329/src/sites/dMCDA.jl#L222C10-L229

That's what you want.

Note that the Promethee method errors because the constructor for it expects additional arguments.
There may be others like it.

Not sure what you want to do with these methods.

Also note the docstring is now out of date.

```julia
ERROR: MethodError: no method matching JMcDM.PROMETHEE.PrometheeMethod()

Closest candidates are:
  JMcDM.PROMETHEE.PrometheeMethod(::Vector{<:Function}, ::Vector{Float64}, ::Vector{Float64})
   @ JMcDM C:\Users\takuy\.julia\packages\JMcDM\Oo88a\src\promethee.jl:22                                                                                                                                                                                                                 |  ETA: 0:00:04        
  JMcDM.PROMETHEE.PrometheeMethod(::Vector{Function}, ::Vector{Union{Nothing, Float64}}, ::Vector{Union{Nothing, Float64}})
   @ JMcDM C:\Users\takuy\.julia\packages\JMcDM\Oo88a\src\promethee.jl:13
  JMcDM.PROMETHEE.PrometheeMethod(::Any, ::Any, ::Any)
   @ JMcDM C:\Users\takuy\.julia\packages\JMcDM\Oo88a\src\promethee.jl:13
```

Rosejoycrocker: Ah man, brackets is all it needed :( ...
Promethee should be in the list of excluded methods because of this- it's more complicated because you get 2 scores are output and based on these you decide whether to have a single or multiple "best choice". I'll add it to the list. The others which I've added should be the only methods which are exceptions, I looked through all th docs to catch any which have non-standard outputs.

I'll update the docstring.

ConnectedSystems: I did say to call the constructor...

https://github.com/open-AIMS/ADRIA.jl/pull/348#discussion_r1201298621

Rosejoycrocker: I didn't realise that's what you meant by your example, I thought you were just referring to the datatype of the inputs because you referred to multile dispatch

Rosejoycrocker:
Hey @ConnectedSystems I think I've resolved your requested changes now. Let me know if there is anything else. I haven't  changed S to be a NamedDims+AxisKeys array because this was a planned change for when replacing mcda_vars which I will do in another PR

ConnectedSystems:
Thanks, will have a look this afternoon

src/sites/dMCDA.jl
```diff
@@ -37,6 +41,23 @@ struct DMCDA_vars  # {V, I, F, M} where V <: Vector
...
+    JMcDM.MEREC.MERECMethod,
+    JMcDM.ELECTRE.ElectreMethod,
+    JMcDM.PROMETHEE.PrometheeMethod
+]
+
+const methods_mcda = [
+    order_ranking,
+    adria_vikor,
+    adria_topsis,
+    setdiff(jmcdm_methods, jmcdm_ignore)...
```
ConnectedSystems, line 58: FYI @Rosejoycrocker thought of a cleaner way to ignore unwanted methods.

src/sites/dMCDA.jl
ConnectedSystems: List of functions in docstring still doesn't match implementations?

src/sites/mcda_methods.jl
```diff
@@ -0,0 +1,124 @@
...
+    R = maximum(sr_arg, dims=2)
+
+    # Compute the VIKOR compromise Q
+    S_s, S_h = maximum(Sr), minimum(Sr)
+    R_s, R_h = maximum(R), minimum(R)
+    Q = @. v * (Sr - S_h) / (S_s - S_h) + (1 - v) * (R - R_h) / (R_s - R_h)
+    Q .= 1.0 .- Q  # Invert rankings so higher values = higher rank
+
+    return Q
+end
```
ConnectedSystems, line 124: You need to add an empty line at the very end to stop the linter from complaining

(I'll explain why leaving an empty line at the end is a common convention in programming later).

src/sites/dMCDA.jl
```diff
@@ -3,6 +3,10 @@
 using StatsBase
 using Distances
 using Combinatorics
+using JMcDM
+using InteractiveUtils: subtypes
+using ADRIA: order_ranking, adria_vikor, adria_topsis
```
ConnectedSystems, line 8: Quick question - was the `adria_` prefix necessary?

Looks a bit tacky in the sense that none of the other methods have the prefix.
At first glance it doesn't seem like `vikor` or `topsis` would have resulted in a clash with JMcDM functions, so wondering if we can save the extra keypresses.

Rosejoycrocker: I originally didn't have the prefix but JMcDM seems to have wrapping functions for the vikor and topsis methods so instead of mcda(A,weights, method()) you can do vikor(A,weights) etc... so I had to have something different

ConnectedSystems: Okay. I still think it looks a bit clunky but leave it as is. It'll be removed in the future if you contribute changes to JMcDM anyway.

src/sites/mcda_methods.jl
Rosejoycrocker: >List of functions in docstring still doesn't match implementations?

I thought I did this in commit e5ea28 , called "Change order of doc input values to match order"... maybe it got overwritten when merged with Main?

src/sites/dMCDA.jl
ConnectedSystems: @Rosejoycrocker I guess you did update them but they're still incorrect.

Corrected, and changed `rev_val` to `maximize` as the latter is more intuitive.

ConnectedSystems (approved):
Thanks for this PR Rose!

PR #349, opened 2023-05-17
Make Aviz into an extension package

ConnectedSystems:
Re-configures the Aviz companion package into an extension package (supported by Julia v1.9 and above).

User now only needs to install/"using" required visualization packages (GLMakie, GeoMakie and GraphMakie).

GraphMakie is currently unused but is intended to be used to display connectivity data.

See updated readme for usage examples.


Rosejoycrocker:
Hi Takuya, just checking I understand this. So now instead of "using Aviz" we use "using GLMakie, GeoMakie, GraphMakie" and the extention will precompile?

ConnectedSystems:
> Hi Takuya, just checking I understand this. So now instead of "using Aviz" we use "using GLMakie, GeoMakie, GraphMakie" and the extention will precompile?

Yes, that's right.

It means if the user only wants to select intervention locations or run ADRIA-mod, but don't care for any of the analysis/visualization stuff, they don't have to install the Makie ecosystem which is quite big/heavy, and slows ADRIA start times.

Rosejoycrocker:
> > Hi Takuya, just checking I understand this. So now instead of "using Aviz" we use "using GLMakie, GeoMakie, GraphMakie" and the extention will precompile?
> 
> Yes, that's right.
> 
> It means if the user only wants to select intervention locations or run ADRIA-mod, and don't care for any of the analysis/visualization stuff, they don't have to install the Makie ecosystem which is quite big/heavy, and slows ADRIA start times.

Thanks, I'm just running it now to check it out then I'll approve.

Rosejoycrocker:
Hi Takuya, does Aviz have to be run in a particular place to work? I just precompiled ADRIA, then GLMakie, GeoMakie and GraphMakie and now I'm running parts of the example from the docs in my sandbox folder but I'm getting the error "viz not defined" when I run 
`ADRIA.viz.scenario(rs, s_tac; axis_opts=Dict(:ylabel => "Example Metric"))` for example. Is there another setup step I'm missing?

ConnectedSystems:
Hi Rose,

1. You've checked out the right branch?
2. You're running Julia v1.9?
3. You've run `instantiate` or `up` for both ADRIA and sandbox environments?

If yes to all three, then it's something else and we'll need to chat online so I can see what's happening.

Rosejoycrocker:
> Hi Rose,
> 
> 1. You've checked out the right branch?
> 2. You're running Julia v1.9?
> 3. You've run `instantiate` or `up` for both ADRIA and sandbox environments?
> 
> If yes to all three, then it's something else and we'll need to chat online so I can see what's happening.

I did 1. and 2., but I updated my ADRIA and sandbox environments instead of instantiate- I'll do this step now

Rosejoycrocker:
> Hi Rose,
> 
> 1. You've checked out the right branch?
> 2. You're running Julia v1.9?
> 3. You've run `instantiate` or `up` for both ADRIA and sandbox environments?
> 
> If yes to all three, then it's something else and we'll need to chat online so I can see what's happening.

I just ran instantiate and I'm getting the same. I can do a quick call now or if you're finishing now happy to chat tomorrow.

Rosejoycrocker (approved):

PR #354, opened 2023-05-22
Update bleaching mortality model to align with published paper

ConnectedSystems:
Previous implementation was based on the pre-print (I think).

Also corrected a bug which did not cap the DHW coefficient to a max of 100, and removed a division by 100 that was being applied unnecessarily.


Rosejoycrocker (approved):
Just went and checked side by side with the paper, looks good to me

ConnectedSystems:
Hi Rose, I think you should go ahead and merge this. Thanks for the review and confirmation.

PR #359, opened 2023-05-25
Remove Copras method due to erroring

### LINKED ISSUE #357: MCDA method errors
@Rosejoycrocker 

Seems we both neglected to run tests for #348.

When running scenarios, the new MCDA module hits an error due to the line below.

```julia
maximize = results.bestIndex == findall(results.scores .== maximum(results.scores))[1]
```

https://github.com/open-AIMS/ADRIA.jl/blob/23af2238ffbcc0843ed1a9ecd0e17a4ab1e0796f/src/sites/dMCDA.jl#L228-L229

The method being used is `JMcDM.COPRAS.CoprasMethod`, however it may occur with others.

When I infiltrate and check the values of results:

```julia
infil> results.scores
128-element Vector{Float64}:
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
 NaN
   ⋮

infil> all(isnan.(results.scores))
true

infil> results.bestIndex
128
```

Not sure what it means if all values are NaN, nor what to do in such cases.

Incidentally, the line to find the index of the maximum score could be replaced with:

```julia
maximize = results.bestIndex == argmax(results.scores)
```

but again, not sure if this makes sense if values are NaN. Where all values are NaN, `argmax(results.scores)` would simply return 1 (i.e., first position), and `results.bestIndex` would be the last position, so `maximize` would be set to `false`.

It avoids the error, but I don't think it's sensible...

Rosejoycrocker:
Closes #357

Copras method was causing issues due to giving all NaNs for the criteria scores, so that rankings were meaningless. On further research it seems the Copras method is a weighted linear combination of the score for criteria to be maximised divided by the maximum criteria score and score for criteria to be minimised divided by minimum criteria score. It seems that maybe because ADRIA only has maximising criteria this was giving NaN scores (all other methods were found to not error with the same decision matrix and the error occured for any decision matrix with the Copras method). To fix this issue, Copras is removed from the available set of methods in this PR.

All other methods were tested for their outputs  to verify that they should not error (given reasonable decision matrix and weightings) and compared to the JMcDM docs.

### DIFF
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -45,7 +45,7 @@ Base.@kwdef struct Intervention{N,P,P2} <: EcoModel
     # Integer values have a +1 offset to allow for discrete value mapping
     # (see `set()` and `map_to_discrete()` methods)
     # Bounds are defined as floats to maintain type stability
-    guided::N = Param(0, ptype="integer", bounds=(-1.0, 20.0 + 1.0), dists="unif",
+    guided::N = Param(0, ptype="integer", bounds=(-1.0, length(methods_mcda) + 1.0), dists="unif",
         name="Guided", description="Choice of MCDA approach.")
     seed_TA::N = Param(0, ptype="integer", bounds=(0.0, 1000000.0 + 1.0), dists="unif",
         name="Seeded Tabular Acropora", description="Number of enhanced Tabular Acropora to seed per deployment year.")
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -45,7 +45,7 @@ Base.@kwdef struct Intervention{N,P,P2} <: EcoModel
     # Integer values have a +1 offset to allow for discrete value mapping
     # (see `set()` and `map_to_discrete()` methods)
     # Bounds are defined as floats to maintain type stability
-    guided::N = Param(0, ptype="integer", bounds=(-1.0, 21.0 + 1.0), dists="unif",
+    guided::N = Param(0, ptype="integer", bounds=(-1.0, 20.0 + 1.0), dists="unif",
         name="Guided", description="Choice of MCDA approach.")
     seed_TA::N = Param(0, ptype="integer", bounds=(0.0, 1000000.0 + 1.0), dists="unif",
         name="Seeded Tabular Acropora", description="Number of enhanced Tabular Acropora to seed per deployment year.")
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -8,6 +8,25 @@ using InteractiveUtils: subtypes
 using ADRIA: order_ranking, adria_vikor, adria_topsis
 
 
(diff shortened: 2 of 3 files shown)


Rosejoycrocker:
Tests all passing on my end

src/ecosystem/Ecosystem.jl
```diff
@@ -45,7 +45,7 @@ Base.@kwdef struct Intervention{N,P,P2} <: EcoModel
     # Integer values have a +1 offset to allow for discrete value mapping
     # (see `set()` and `map_to_discrete()` methods)
     # Bounds are defined as floats to maintain type stability
-    guided::N = Param(0, ptype="integer", bounds=(-1.0, 21.0 + 1.0), dists="unif",
+    guided::N = Param(0, ptype="integer", bounds=(-1.0, length(methods_mcda) + 1.0), dists="unif",
```
ConnectedSystems, line 48: Using the `const` number of available methods instead to avoid repeated editing

src/sites/dMCDA.jl
```diff
@@ -8,6 +8,25 @@ using InteractiveUtils: subtypes
...
 
 
+jmcdm_ignore = [
+    JMcDM.CRITIC.CriticMethod,
+    JMcDM.COPRAS.CoprasMethod,
+    JMcDM.MOOSRA.MoosraMethod,
+    JMcDM.MEREC.MERECMethod,
+    JMcDM.ELECTRE.ElectreMethod,
+    JMcDM.PROMETHEE.PrometheeMethod,
+    JMcDM.Topsis.TopsisMethod,
```
ConnectedSystems, line 18: Rose, I've removed TOPSIS and VIKOR from consideration, but did you intend to keep these in for a reason I wasn't aware of?

ConnectedSystems (approved):
@Rosejoycrocker some minor comments for your review. If all okay then I will merge.

src/sites/dMCDA.jl
Rosejoycrocker: That probably makes sense to avoid overlap. I kept them in incase we wanted a "fast version" because they are quite a bit faster than the JMcDM implementations, but it's not very noticable when running a lot of simulations.

Rosejoycrocker:
All good on my end :)

src/sites/dMCDA.jl
ConnectedSystems: >  it's not very noticable when running a lot of simulations.

That's when it should be very noticeable? I'm noting roughly an extra minute in run times compared to when we were only considering the three methods for 1024 scenarios.

I'll do a more in-depth profiling, but noting that seemingly small increases has a large implication for scenario/sensitivity analysis.

ConnectedSystems:
image

FYI, PSIMethod looks promising, at least for coral cover (14th in the methods list).

Rosejoycrocker:
Oh cool, PSI is based on approximate pareto optimal solutions in the criteria space.

PR #367, opened 2023-06-06
Address ranking errors in mcda outputs

### LINKED ISSUE #366: align_rankings! giving incorrect output rankings
align_rankings! is a function which should assign ranks to the set of site ids using the site ids ordered according to mcda score in s_order.

For example, if s_order outputs 
```julia
[5  0.77,
10 0.75,
22 0.72,
11 0.68,
1 0.66]
```

Then the ranks should be
```julia
[1 5,
5  1,
10 2,
11 4,
22 3]
```

Currently the mcda starts with 
```julia
s_order =
[1 0.66 1
5  0.77 2,
10 0.75 3,
11 0.68 4,
22 0.72 5]
```
And uses the third column to assign ranks during rank calculation. However, the third column is also reordered when sorting occurs, so the ranks are not assigned correctly. For example:

```julia
s_order =
[5  0.77 2,
10 0.75 3,
22 0.72 5,
11 0.68 4,
1 0.66 1]
```

And then ranks are assigned as in https://github.com/open-AIMS/ADRIA.jl/blob/c1bf6852a32e56bc362da6fd293e691ef3bfa616/src/sites/dMCDA.jl#L166

Rosejoycrocker:
Closes #366
Adds test to check that rankings line up with site ordering in s_order.

### DIFF
--- a/src/sites/dMCDA.jl
+++ b/src/sites/dMCDA.jl
@@ -205,7 +205,7 @@ end
 """
     retrieve_ranks(S::Matrix, site_ids::Vector, weights::Vector{Float64}, mcda_func::Function)
     retrieve_ranks(S::Matrix, site_ids::Vector, weights::Vector{Float64}, mcda_func::Type{<:MCDMMethod})
-    retrieve_ranks(S::Matrix, site_ids::Vector, scores::Vector, maximize::Bool)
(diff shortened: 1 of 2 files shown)


ConnectedSystems (approved):
LGTM

PR #380, opened 2023-07-03
Changes to number of species in ADRIA

Rosejoycrocker:
Enhanced corals no longer represented as species in ADRIA. Two enhanced coral species are replaced with Abhorescent Acropora and Corymbose non-acropora. Growth rates, death rates and fecundity updated accordingly. Metrics such as shelter volume and juveniles also updated. ODE in ADRIA no longer includes competition terms between Acropora and small massives.


src/ecosystem/corals/growth_expanded.jl
```diff
@@ -21,48 +21,47 @@ function growthODE_expanded(du::Array{Float64,2}, X::Array{Float64,2}, p::NamedT
     r = @view p.r[:]  # growth rates
     X_mb = X .* p.mb   # current cover * background mortality
 
-    # 0.3 is the competition rate
-    # Tabular Acropora Enhanced
+    # Aborescent Acropora
```
ConnectedSystems, line 24: If this has been updated to reflect the changes in this PR, please delete `slow_ODE()`

src/scenario.jl
```diff
@@ -381,8 +380,8 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     seed_sc_CA::Int64 = first(findall(corymbose_enhanced .& target_class_id))
 
     # Extract colony areas for sites selected and convert to m^2
-    col_area_seed_TA = corals.colony_area_cm2[seed_sc_TA] / 10^4
-    col_area_seed_CA = corals.colony_area_cm2[seed_sc_CA] / 10^4
+    col_area_seed_TA = pi *(corals.mean_colony_diameter_m[seed_sc_TA]/2)^2
```
ConnectedSystems, line 383: Fix spacing here (did auto-format not work?).

Use consistent data types as well (e.g., `/2` should be `/ 2.0`)

ConnectedSystems: The same formula is used below (L 384) - might as well put it in a function for readability.

Rosejoycrocker: I just made a function instead as you suggested and delleted slow_ODE. Let me know if there are other changes needed. Tests pass on my end.

src/ecosystem/Ecosystem.jl
```diff
@@ -207,6 +207,18 @@ function create_coral_struct(bounds::Tuple{Float64,Float64}=(0.9, 1.1))::Nothing
     return
 end
 
+"""
+    colony_areas(colony_diam_means)
+
+Generate colony areas for each size class.
+
+# Arguments
+- `colony_diam_means_m` : mean colony diameter (in meters)
```
ConnectedSystems, line 216: Mismatched argument name?

Rosejoycrocker: Fixed now, let me know if there's anything else

src/ecosystem/Ecosystem.jl
```diff
@@ -207,6 +207,18 @@ function create_coral_struct(bounds::Tuple{Float64,Float64}=(0.9, 1.1))::Nothing
...
+
+Generate colony areas for each size class.
+
+# Arguments
+- `colony_diam_means` : mean colony diameter (in meters)
+
+"""
+function colony_mean_area(colony_diam_means)
+    return pi .* ((colony_diam_means ./ 2.0).^2)
+end
```
ConnectedSystems, line 221: Function signature in the docstring doesn't match the implemented function? (L 211 compared to L 219)

src/ecosystem/Ecosystem.jl
```diff
@@ -207,6 +207,18 @@ function create_coral_struct(bounds::Tuple{Float64,Float64}=(0.9, 1.1))::Nothing
     return
 end
 
+"""
+    colony_areas(colony_diam_means)
+
+Generate colony areas for each size class.
```
ConnectedSystems, line 213: Is this statement always true?

src/ecosystem/Ecosystem.jl
ConnectedSystems: Oh I see, you copy pasted from below and forgot to update

src/ecosystem/corals/growth.jl
```diff
@@ -55,106 +55,24 @@ a 6 by \$n_sites\$ array.
...
     # rec : recruitment factors for each coral group (6 by n_sites)
 
-    @views @. du[p.acr_5_11, :] = p.sXr[p.acr_5_11-1, :] - p.sXr[p.acr_5_11, :] + (p.sm_comp * X[p.acr_5_11, :]) - p.X_mb[p.acr_5_11, :]
-    @views @. du[p.acr_6_12, :] = p.sXr[p.acr_6_12-1, :] + p.sXr[p.acr_6_12, :] + (p.sm_comp * X[p.acr_6_12, :]) - p.X_mb[p.acr_6_12, :]
-
-    @views @. du[p.small_massives, :] = p.sXr[p.small_massives-1, :] - p.sXr[p.small_massives, :] - p.M_sm
-
     @views @. du[p.small, :] = p.rec - p.sXr[p.small, :] - p.X_mb[p.small, :]
     @views @. du[p.mid, :] = p.sXr[p.mid-1, :] - p.sXr[p.mid, :] - p.X_mb[p.mid, :]
     @views @. du[p.large, :] = p.sXr[p.large-1, :] + p.sXr[p.large, :] - p.X_mb[p.large, :]
```
ConnectedSystems, line 71: If removing competition and other terms from the calculations make these identical, do we need to separate mid from large now? Previously it was done for conceptual clarity, but I'm not sure that's necessary now.

src/ecosystem/Ecosystem.jl
```diff
@@ -207,6 +207,18 @@ function create_coral_struct(bounds::Tuple{Float64,Float64}=(0.9, 1.1))::Nothing
...
+
+# Arguments
+- `colony_diam_means` : mean colony diameter (in meters)
+
+"""
+function colony_mean_area(colony_diam_means)
```
ConnectedSystems, line 219: Would `mean_colony_area()` make more sense for readability?

This looks like a parameter naming scheme I use: subject first (i.e., colony) so you don't get a million false matches when doing a quick search.

That said, it matches the naming scheme of `colony_areas()` below, so makes sense we leave it as is.

src/scenario.jl
```diff
@@ -495,7 +494,7 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
             env_horizon[1, :] .= dhw_t
             mcda_vars.heat_stress_prob .= vec((mean(env_horizon, dims=1) .+ std(env_horizon, dims=1)) .* 0.5)
 
-            mcda_vars.dam_prob .= vec(sum(dropdims((mean(Sw_t[horizon, :, :], dims=1) .+ std(Sw_t[horizon, :, :], dims=1)) .* 0.5, dims=1), dims=1))
+            #mcda_vars.dam_prob .= vec(sum(dropdims((mean(Sw_t[horizon, :, :], dims=1) .+ std(Sw_t[horizon, :, :], dims=1)) .* 0.5, dims=1), dims=1))
```
ConnectedSystems, line 497: space after # please

ConnectedSystems: Same for the commented code below

ConnectedSystems (changes requested):
Hi Rose, thanks for this.

A few more minor comments for your consideration.

Some questions:

1. Did you run an example scenario to see if outputs (coral cover and shelter volume) make sense?
2. Have you checked RCI yet? (it may be outside the scope of this PR - just wondering)
3. I assume you have updated the Domain datasets to use updated initial coral cover data. Could I please get a copy?

src/ecosystem/corals/growth.jl
Rosejoycrocker: There is still the difference in sign after the first term between mid and large. Could combine the last terms if you think that's useful?

src/ecosystem/Ecosystem.jl
Rosejoycrocker: Yeah I made it similar to colony_areas because they're related, but happy to change

ConnectedSystems: It's consistent, so happy to leave as is.

src/ecosystem/corals/growth.jl
ConnectedSystems: Ah, I see, sorry my eyes glazed over that. Leave as is then.

Rosejoycrocker:
> Hi Rose, thanks for this.
> 
> A few more minor comments for your consideration.
> 
> Some questions:
> 
> 1. Did you run an example scenario to see if outputs (coral cover and shelter volume) make sense?
> 2. Have you checked RCI yet? (it may be outside the scope of this PR - just wondering)
> 3. I assume you have updated the Domain datasets to use updated initial coral cover data. Could I please get a copy?

Sorry for the wait, I found an error in converting m to cm in SV. Here is a plot of some scenarios fr coral cover:
ex_samples_original_data

and shelter volume:

ex_samples_original_data_sv

Coral cover looks fine to me, I have less intuition about SV but the numbers seem reasonable?

Also I haven't updated the Domain dataset. Is this updated from the reefmod datasets?

ConnectedSystems:
@KRNA01 do the plots above look reasonable? Unit for Absolute Shelter Volume is in cubic meters per m^2 I believe.


@Rosejoycrocker 

>Also I haven't updated the Domain dataset. Is this updated from the reefmod datasets?

ReefMod will have very coarse data for this example location. From memory we used IPMF/C~Scape data to translate to the four coral types we represent.

@KRNA01 could you advise here as well?

KenAnthony1:
Assuming this is for SSP2, then the results above for coral cover, and especially absolute shelter volume, look aggressive to me. 
 
The results look like ReefMod results for SSP3 or SSP5.

Also, from memory, we had shelter volume as cubic metres per hectare. If we're using cubic meters per square meter, then the range on the Y-axis should be around 0 to 0.2 m3/m2 max.

ConnectedSystems:
@Rosejoycrocker you could go back and compare what SV was showing previously. The value range Ken mentions rings a bell.

Rosejoycrocker:
> @Rosejoycrocker you could go back and compare what SV was showing previously. The value range Ken mentions rings a bell.

Hi both, SV definitely looks different in Main, so I'm looking into why right now
ex_samples_original_data_sv_main

Coral cover looks pretty similar though (I  ran more scenarios in the fig below):
ex_samples_original_data_main

Rosejoycrocker:
Hi @KRNA01, could there be a difference in the new shelter volume metric we're using compared to the old one from Urbina-Barretto? The Aston et. al. paper you sent has two metrics, an 'Absolute spatial refuge' in dm^3 and 'Shelter size factor'. I used the former, but it seems to have a slightly different definition to shelter volume, being defined as "intertitial space... higher values of R indicate more shelter volume" (Table 2, pg. 6 of the paper). Is this the metric you meant me to use?

src/ecosystem/corals/growth.jl
ConnectedSystems, line 71: The block of commented code from Line 180 onwards either needs to be updated or removed

ConnectedSystems:
@Rosejoycrocker 

Are we also expanding the list of corals used for deployments?
The ones marked with asterixis below are those used in ReefMod

arborescent Acropora*
tabular Acropora*
corymbose Acropora*
Pocillopora + non Acropora corymbose
Mix encrusting and submassive*
Large Porites

Rosejoycrocker:
Good point, should we @KenAnthony1 ? I'm guessing it is best to be consistent with ReefMod

KRNA01:
Yes, I think it’s best to be consistent with ReefMod

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Rosejoycrocker:
Agree, I will add this.

Rosejoycrocker:
Hi @ConnectedSystems and @KenAnthony1 , I'm working through adding these additional species to planted corals and had some questions. @KenAnthony1, from the list above are we adding Aborrescent Acropora, Corymbose non-Acropora and small massives/encrusting to our list of planted corals? Let me know if I have missed any.

@ConnectedSystems, if we are adding species do we want to assume certain species are grouped together in terms of the number of corals to plant (so we don't have to add more parameters to the Intervention struct), or should I add a new parameter for each species?

ConnectedSystems:
>  if we are adding species do we want to assume certain species are grouped together in terms of the number of corals to plant (so we don't have to add more parameters to the Intervention struct), or should I add a new parameter for each species?

I say no, but it really depends on what Ken's table that maps the deployed species with their modelled representations. 

Also, from memory ReefMod just takes a single number and divides it evenly across the four, but I'd like to explore what happens with a bit more control. To do that we'd need the extra two parameters (for the additional two seeding corals being represented).

ConnectedSystems:
FYI, my changes for natural adaptation are essentially complete. Only thing left to do is reconcile the changes in #381 with the changes in this PR and update tests, so I'll wait until you're finished here to do that.

KenAnthony1:
@Rosejoycrocker and @ConnectedSystems 

See below. 

Note that these ECT / CAD species are likely to change, so it's a good idea for us to be flexible re how we model seeding efforts for each species.

I suspect Corymbose and Tabular Acropora will be the most prominent groups, but we (MDS) should recommend that a mixture of species will be better from the perspective of supporting reef function and resilience.  

image

Rosejoycrocker:
Thanks Ken, I'll try and adjust the code to be more flexible to adding species to seeding for future changes.

Rosejoycrocker:
Hey @ConnectedSystems, I've added small massives into seeding now so I think it's ready for review. I tried to use NamedDimsArray for seeded taxa parameters with the seeded taxa names as keys to be more generalisable but not sure I've done it in a way that's consistent with the rest of ADRIA.

src/metrics/metrics.jl
```diff
@@ -383,12 +383,10 @@ Helper function to convert coral colony values from Litres/cm² to m³/m²
...
-1. Urbina-Barreto, I., Chiroleu, F., Pinel, R., Fréchon, L., Mahamadaly, V., Elise, S., Kulbicki, M., Quod, J.-P.,
-     Dutrieux, E., Garnier, R., Henrich Bruggemann, J., Penin, L., & Adjeroud, M. (2021).
-   Quantifying the shelter capacity of coral reefs using photogrammetric 3D modeling:
-     From colonies to reefscapes.
-   Ecological Indicators, 121, 107151.
-   https://doi.org/10.1016/j.ecolind.2020.107151
+1. Aston Eoghan A., Duce Stephanie, Hoey Andrew S., Ferrari Renata (2022).
+    A Protocol for Extracting Structural Metrics From 3D Reconstructions of Corals.
+    Frontiers in Marine Science, 9.
+    https://doi.org/10.3389/fmars.2022.854395
```
ConnectedSystems, line 389: Noting this referencing style is not consistent. Don't worry about it for now, but we'll have to go through and standardize everything at some point.

src/scenario.jl
```diff
@@ -330,7 +328,7 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     site_ranks = SparseArray(zeros(tf, n_sites, 2)) # log seeding/fogging/shading ranks
     Yshade = SparseArray(spzeros(tf, n_sites))
     Yfog = SparseArray(spzeros(tf, n_sites))
-    Yseed = SparseArray(zeros(tf, 2, n_sites))  # 2 = the two enhanced coral types
+    Yseed = SparseArray(zeros(tf, 3, n_sites))  # 2 = the two enhanced coral types
```
ConnectedSystems, line 331: Update the comment too please

src/scenario.jl
```diff
@@ -368,29 +366,41 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
     p.mb .= corals.mb_rate  # background mortality
     @set! p.k = max_cover  # max coral cover
-    @set! p.comp = sim_params.comp  # competition rate
 
     # Proportionally adjust initial cover (handles inappropriate initial conditions)
     proportional_adjustment!(Y_cover[1, :, :], cover_tmp, max_cover)
+    
+    # Define constant table location for seed values and seeded coral types
+    taxa_to_seed = [2,3,5]
+    taxa_names = ["seed_CA","seed_TA","seed_SM"]
```
ConnectedSystems, line 375: Get the seeded taxa names from `param_set`.

You tried to make things more generic so no sense in hardcoding more than you have to.

src/scenario.jl
```diff
@@ -368,29 +366,41 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-    col_area_seed_TA = corals.colony_area_cm2[seed_sc_TA] / 10^4
-    col_area_seed_CA = corals.colony_area_cm2[seed_sc_CA] / 10^4
+    # Set up assisted adaptation values
+    a_adapt = zeros(n_species)
+    # Flag indicating whether to seed or not to seed
+    seed_corals = false
+    
+    # Extract colony areas for sites selected in m^2 and add adaptation values
+    for taxa in seeded_area.taxa
+        seeded_area[seeded_area.taxa.==taxa] .= colony_mean_area(corals.mean_colony_diameter_m[findall(seed_sc(taxa))])[1].*param_set(taxa)
```
ConnectedSystems, line 390: I'm not convinced `findall(seed_sc(taxa))])` is necessary here. Do you need it because `corals.mean_colony_diameter_m` is a NamedDimsArray or similar that has a quirk?

Actually, because you're looping over each element in `seeded_area`, this line doesn't seem necessary?

`seeded_area` is a three element vector, the order of which should align with the vectorized output of `colony_mean_area()` ?

src/scenario.jl
```diff
@@ -368,29 +366,41 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
+        a_adapt[taxa_to_seed_ids(taxa)] .= param_set("a_adapt")
```
ConnectedSystems, line 391: Actually, the for loop doesn't seem necessary. Is there something I'm not seeing that prevents each of the lines here from being vectorized?

src/scenario.jl
```diff
@@ -368,29 +366,41 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-    seed_sc_CA::Int64 = first(findall(corymbose_enhanced .& target_class_id))
+    seed_sc = taxa_to_seed_ids .& target_class_id # size class indices for TA, CA and SM
 
-    # Extract colony areas for sites selected and convert to m^2
-    col_area_seed_TA = corals.colony_area_cm2[seed_sc_TA] / 10^4
-    col_area_seed_CA = corals.colony_area_cm2[seed_sc_CA] / 10^4
+    # Set up assisted adaptation values
+    a_adapt = zeros(n_species)
+    # Flag indicating whether to seed or not to seed
+    seed_corals = false
```
ConnectedSystems, line 386: The check every loop below can be replaced with something like

```suggestion
    seed_corals = any(param_set(factors=seeded_area.taxa) .> 0.0)
```

src/scenario.jl
```diff
@@ -29,7 +29,7 @@ function setup_cache(domain::Domain)::NamedTuple
         depth_coeff=zeros(n_sites),  # store for depth coefficient
         site_area=Matrix{Float64}(site_area(domain)'),  # site areas
         TP_data=Matrix{Float64}(domain.TP_data),  # transition probabilities
-        waves=zeros(length(timesteps(domain)), n_species, n_sites)
+        # waves=zeros(length(timesteps(domain)), n_species, n_sites)
```
ConnectedSystems, line 32: Undo all changes to do with waves. I have temporary changes in PR #381 as we discussed.

src/scenario.jl
```diff
@@ -368,29 +366,41 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
+    seeded_area = NamedDimsArray(zeros(length(taxa_to_seed)),taxa=taxa_names)
+    taxa_to_seed_ids = NamedDimsArray(corals.taxa_id .== taxa_to_seed',species=1:n_species,taxa=taxa_names)
 
-    # Define constant table location for seed values
-    tabular_enhanced::BitArray = corals.taxa_id .== 1
-    corymbose_enhanced::BitArray = corals.taxa_id .== 3
     target_class_id::BitArray = corals.class_id .== 2  # seed second smallest size class
-    seed_sc_TA::Int64 = first(findall(tabular_enhanced .& target_class_id))  # size class indices for TA and CA
-    seed_sc_CA::Int64 = first(findall(corymbose_enhanced .& target_class_id))
+    seed_sc = taxa_to_seed_ids .& target_class_id # size class indices for TA, CA and SM
```
ConnectedSystems, line 381: Double space after code for inline comments please

src/io/result_io.jl
```diff
@@ -126,7 +126,7 @@ function setup_logs(z_store, unique_sites, n_scens, tf, n_sites)
     fog_dims::Tuple{Int64,Int64,Int64} = (tf, n_sites, n_scens)  # timeframe, sites, no. scenarios
 
     # tf, no. intervention sites, site id and rank, no. scenarios
-    seed_dims::Tuple{Int64,Int64,Int64,Int64} = (tf, 2, n_sites, n_scens)
+    seed_dims::Tuple{Int64,Int64,Int64,Int64} = (tf, 3, n_sites, n_scens)
```
ConnectedSystems, line 129: Could you update the comment in the line above as well please?

ConnectedSystems (changes requested):
Thanks, Rose, almost there.

src/scenario.jl
Rosejoycrocker: Sure, so using `occursin()`, find everything with "seed_"? Or is there another way to do this?

src/scenario.jl
Rosejoycrocker: I'll see if I can vectorize and remove findall, it's mostly copied from the previous code

Rosejoycrocker:
Hey @ConnectedSystems, I think I've addressed your comments.

src/scenario.jl
ConnectedSystems: Yes, `occursin()` is one way. Not sure how performant it is but given it should only happen once per run I'm less concerned about it.

Later we could move this to somewhere else and set as a constant.

src/scenario.jl
```diff
@@ -368,29 +366,37 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-    # Define constant table location for seed values
-    tabular_enhanced::BitArray = corals.taxa_id .== 1
-    corymbose_enhanced::BitArray = corals.taxa_id .== 3
     target_class_id::BitArray = corals.class_id .== 2  # seed second smallest size class
-    seed_sc_TA::Int64 = first(findall(tabular_enhanced .& target_class_id))  # size class indices for TA and CA
-    seed_sc_CA::Int64 = first(findall(corymbose_enhanced .& target_class_id))
+    seed_sc = taxa_to_seed_ids .& target_class_id  # size class indices for TA, CA and SM
+
+    # Extract colony areas for sites selected in m^2 and add adaptation values
+    seeded_area = colony_mean_area(repeat(corals.mean_colony_diameter_m,1,3)[seed_sc]).*param_set(taxa_names)
```
ConnectedSystems, line 384: `seeded_area = colony_mean_area(repeat(corals.mean_colony_diameter_m,1,3)[seed_sc]).*param_set(taxa_names)`

This seems very convoluted. I'll simplify.

src/scenario.jl
```diff
@@ -368,29 +366,36 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-    col_area_seed_CA = corals.colony_area_cm2[seed_sc_CA] / 10^4
+    # Extract colony areas for sites selected in m^2 and add adaptation values
+    seeded_area = colony_mean_area(corals.mean_colony_diameter_m[seed_sc]) .* param_set(taxa_names)
+
+    # Set up assisted adaptation values
+    a_adapt = zeros(n_species)
+    a_adapt[seed_sc] .= param_set("a_adapt")
+
+    # Flag indicating whether to seed or not to seed
+    seed_corals = any(param_set(taxa_names) .> 0.0)
```
ConnectedSystems, line 389: @Rosejoycrocker simplified approach avoiding repeated indexing and transposes

src/scenario.jl
```diff
@@ -537,16 +535,11 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-            # Seed each site with TA or CA
-            @views Y_pstep[seed_sc_TA, prefseedsites] .= Y_pstep[seed_sc_TA, prefseedsites] .+ scaled_seed.TA
-            @views Y_pstep[seed_sc_CA, prefseedsites] .= Y_pstep[seed_sc_CA, prefseedsites] .+ scaled_seed.CA
-
-            # Log seed values/sites (these values are relative to site area)
-            Yseed[tstep, 1, prefseedsites] .= scaled_seed.TA
-            Yseed[tstep, 2, prefseedsites] .= scaled_seed.CA
+            # Seed each site
+            @views Y_pstep[seed_sc, prefseedsites] .+= scaled_seed
+            Yseed[tstep, :, prefseedsites] .= scaled_seed
```
ConnectedSystems, line 542: This way, no need for that weird for loop.

FYI, when using loops in future, please conform to `idx` to indicate `index`.

ConnectedSystems (approved):
Looks all good now, will merge shortly.

PR #356, opened 2023-05-22
Add map visualization - displays k-area

ConnectedSystems:
**Currently throws a recoverable error.**

Seems like some issue within GLMakie or GeoMakie, so marking this as a draft PR until the issue is resolved.

```julia

dom = ADRIA.load_domain("... some domain ...")

ADRIA.viz.map(dom)

# or 

# rs = ADRIA.load_results("... some results ...")
# ADRIA.viz.map(rs)
```

image

PR #382, opened 2023-07-06
Temporal clustering for spatial data

Zapiano:
Add visualization for temporal clustering when working with spatial data.


src/analysis/clustering.jl
```diff
@@ -117,4 +117,11 @@ function time_series_clustering(data::AbstractMatrix{T}, n_clusters::Int64)::Vec
 
     # Hierarchical clustering with n_clusters clusters
     return cutree(dendogram, k=n_clusters)
+end
+function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T}, n_clusters::Int64)::Vector{Int64} where {T<:Real}
+    # find sites with k > 0.0
```
ConnectedSystems, line 122: Please use sentence case for consistency and neatness

src/viz/viz.jl
```diff
@@ -29,5 +29,6 @@ function outcome_map!() end
 # Clustering
 function ts_cluster() end
 function ts_cluster!() end
+function spatial_clustering() end
```
ConnectedSystems, line 32: Change `spatial_clustering` -> `spatial_cluster` for consistency with `ts_cluster` ?

ConnectedSystems:
This PR relies on #356 , so will get that in asap.

ConnectedSystems:
@Zapiano 

I've rebased onto `add-map-viz` so that the most recent map functionality is available.
We'll have to work out a good way of showing cluster membership along with outcomes - I see what your concern was now.

```julia
rs = ADRIA.load_results("some outputs")

# Get coral cover outcomes in absolute units (m²)
tac = ADRIA.metrics.total_absolute_cover(rs)

# Mean outcomes for each site (median is actually preferred but very slow!)
m_tac_locs = ADRIA.metrics.per_site(mean, tac)

# Mean outcomes for each site and time step
m_ts_tac = dropdims(mean(tac, dims=:scenarios), dims=:scenarios)  

# Get IDs of clusters according to site trajectories
cluster_ids = ADRIA.analysis.time_series_clustering(rs, m_ts_tac, 6)

# Show map using the mean outcomes for each site
ADRIA.viz.spatial_clustering(rs, m_tac_locs, cluster_ids; 
    opts=Dict{Symbol,Any}(:colorbar_label => "Total Absolute Cover"))
```

image

Zooming in:

image

Maybe if we flip the colors so that the stroke color indicates outcomes, and fill color indicates cluster membership?
Alternatively, what if we abandon the idea and just display two maps (one with membership, the other with outcomes)?

I'd also like to see a legend with indicating which color is what cluster, and the indicative values of cluster outcomes

src/viz/viz.jl
Zapiano: I have changed it to ts_spatial_clustering because it is in the same context, what do you think?

Zapiano:
@ConnectedSystems I have moved the function `spatial_cluster`  that was inside `viz/clustering_map.jl` to `viz/clustering.jl` and renamed it to `ts_spatial_cluster`. We can change that later, but I did that for two main reasons:

1. This function only makes sens in the context of this time series clustering (is not a general spatial plot, it's a specific type of spatial plot related to time series clustering)
2. There's is one internal function that is very specific to this clustering context and that is shared between `ts_spatial_cluster` and `ts_cluster`, which is `_clusters_colors`

What do you think?

ConnectedSystems:
@Zapiano 

Consolidating the code into a single file (`clustering.jl`) makes sense to me.

On `ts_spatial_cluster`, we can keep things simple, leveraging Julia's multiple dispatch: I've just changed it to `map()`

`ADRIA.viz.ts_spatial_cluster(rs, outcomes, cluster_ids)` -> `ADRIA.viz.map(rs, outcomes, cluster_ids)`

It's not really specific to time series as we're squashing that down to some summary statistic (mean, median etc) or some other arbitrary function.

So, we have a method that allows one to plot a map for the domain with a given value for each represented location (with a vector of values), and a similar method that accepts a matrix that represents a series of values for each location. Series data for each location is then summarized so we end back to a vector of values for each location.

I've adjusted the stroke and color defaults - looks a bit better although we should use a different color scheme for the stroke.
We're also still missing the cluster legend.

image

Zapiano:
@ConnectedSystems 

I have changed the visualization a little bit and I think it's a little bit easier to understand the map now, what do you think? My idea is: because we are using colors for the clusters, it would be better to use a gray scale as the metric continuous color scale. Also, I mapped the darkest color (in this case, black) to higher values and the lighter (white) to 0, and that also contributed to "clean" the visualization.  

I have also added legends and documentation, as you can see.

tsc_map

ConnectedSystems:
> I have also added legends and documentation, as you can see.

Wonderful, thank you!

> I have changed the visualization a little bit and I think it's a little bit easier to understand the map now, what do you think?

This is great, thank you! 

Could we perhaps make the lowest value a bit gray? It may look a little nicer if it's not white (given the background is also white).
I think also make the legend not to use scientific notation for the cluster means - easier to compare that way in my opinion.

I'll add some other comments as a code review.

docs/src/usage/analysis.md
```diff
@@ -165,12 +165,46 @@ tsc_fig = ADRIA.viz.ts_cluster(
...
+
+# Run sampled scenarios for multiple RCP (Representative Concentration Pathway)
+rcp = "45"
+rs = ADRIA.run_scenarios(scens, dom, rcp)
+
+# Extract metric from scenarios
+tac = ADRIA.metrics.total_absolute_cover(rs)
+
+# Summarize scenarios for each site
+m_tac = dropdims(median(tac, dims=:scenarios), dims=:scenarios)
```
ConnectedSystems, line 194: There's now a `per_site` method

https://github.com/open-AIMS/ADRIA.jl/blob/a550e03c6032938c8f9f9f8a57871100413782af/src/metrics/site_level.jl#L4-L17

(part of the other PR that this is rebased on)

I'll have to finalize the documentation for that though.

ext/AvizExt/viz/clustering.jl
```diff
@@ -29,24 +32,162 @@ function ADRIA.viz.ts_cluster!(g::Union{GridLayout,GridPosition}, data::Abstract
...
+function _cluster_legend_params(clusters::Vector{Int64}, 
+    clusters_colors::Vector{RGBA{Float32}}, data_statistics::Vector{Float32})::Tuple
+    # Filter non-zero clusters from clusters, colors and data
+    non_zero_clusters = clusters .!= 0
+    clusters_filtered = clusters[non_zero_clusters]
+    colors_filtered = clusters_colors[non_zero_clusters]
+    statistics_filtered = data_statistics[non_zero_clusters]
+
+    # Fill legend entries colors
+    legend_entries = [PolyElement(color = color, strokecolor = :transparent) for
```
ConnectedSystems, line 179: I'm surprised the formatter did not remove the whitespace in the argument keyword assignment (`color=color`)! Could you please re-format?

ext/AvizExt/viz/clustering.jl
```diff
@@ -29,24 +32,162 @@ function ADRIA.viz.ts_cluster!(g::Union{GridLayout,GridPosition}, data::Abstract
...
+    clusters_numbers = unique(clusters_filtered)
+    for cluster in clusters_numbers
+        stat_mean = mean(statistics_filtered[cluster .== clusters_filtered])
+        stat_mean_formatted = @sprintf "%.1e" stat_mean
+        push!(legend_labels, "Cluster $(cluster): $stat_mean_formatted")
+    end
 
+    legend_title = "Clusters mean value"
+    return (legend_entries, legend_labels, legend_title)
+end
```
ConnectedSystems, line 193: Please add newline at end of file

ext/AvizExt/viz/clustering.jl
```diff
@@ -29,24 +32,162 @@ function ADRIA.viz.ts_cluster!(g::Union{GridLayout,GridPosition}, data::Abstract
...
+    # Fill legend entries colors
+    legend_entries = [PolyElement(color = color, strokecolor = :transparent) for 
+        color in unique(colors_filtered)]
+
+    # Fill legend labels
+    legend_labels = String[]
+    clusters_numbers = unique(clusters_filtered)
+    for cluster in clusters_numbers
+        stat_mean = mean(statistics_filtered[cluster .== clusters_filtered])
+        stat_mean_formatted = @sprintf "%.1e" stat_mean
```
ConnectedSystems, line 187: As mentioned, I think using formatted values (10, 100, 1K, 10K, 100K, 1M, etc) is easier for comparison than scientific notation

src/analysis/clustering.jl
```diff
@@ -117,4 +121,15 @@ function time_series_clustering(data::AbstractMatrix{T}, n_clusters::Int64)::Vec
...
+    # Find sites with k > 0.0
+    non_null_sites = result_set.site_data.k .> 0.0
+    filtered_data = data[:, non_null_sites]
+
+    # Apply time series cluster for filtered data
+    clusters = time_series_clustering(filtered_data, n_clusters)
+
+    # Assign cluster 0 for filtered sites to use same indexing as the input data
+    return [site != 0 ? popfirst!(clusters) : 0 for site in non_null_sites]
 end
```
ConnectedSystems, line 135: Missing newline at end of file.

ConnectedSystems (commented):
Looking in good shape! I'll get #356 in while you address the minor comments here.

docs/src/usage/analysis.md
Zapiano: @ConnectedSystems I tried to use this method but apparently it returns a VECTOR, not a MATRIX. If I understood correctly, this is computing the metric over scenarios AND timesteps? If this is true, we may need to change this method..?

ConnectedSystems: Yes, it returns a vector providing a value for each site (as in the documentation), which is what you want when showing outcomes spatially (e.g., the mean outcomes over time, or the outcome at a point in time).

If you want the "raw" data for each site AND timestep, you're looking for the relevant metric functions:

https://github.com/open-AIMS/ADRIA.jl/blob/5e97710e8fd5f40ab35c26cc482be47fde7e8694/src/metrics/metrics.jl#L161-L173

ext/AvizExt/viz/clustering.jl
Zapiano: @ConnectedSystems  My vscode was configured to use a formatter that was doing that. I have changed to the other Julia formatter I can choose here and, apparently, it is right now. Please let me know if you see something else that is wrong.

Zapiano: UPDATE: apparently when I use this other format option it adds a lot of blank spaces at the end of some lines (I think we had the same problem in the past, with some spaces that we couldn't see at vscode but were appearing at GitKraken...). I am not sure what's going on, do you remember what we did when that happened a few weeks ago?

ext/AvizExt/viz/clustering.jl
ConnectedSystems: This is a small comment though - if it seems too time consuming to get working, please ignore for now.

ext/AvizExt/viz/clustering.jl
ConnectedSystems: I think just `trim` (ctrl+shift+p -> "trim" -> enter), I'll have to see what's going on next time we chat.

Zapiano: I have already tried `trim`, it didn't work... I'll have to investigate further...

docs/src/usage/analysis.md
Zapiano: But the thing is, what this line of code is doing:

`m_tac = dropdims(median(tac, dims=:scenarios), dims=:scenarios)`

I can't do with `per_site` - right? 
When you wrote this comment, how were you thinking I could use this method `per_site` here?

ConnectedSystems: You can use `per_site` here? I'm not sure I understand what the issue is.

`m_tac` here is a vector, of $N$ elements, where $N$ is the number of locations, no?

ConnectedSystems: `m_tac = ADRIA.metrics.per_site(median, tac)`

ConnectedSystems: Oh, I see what you're getting at

Sorry, I was thinking of this line:

```
# Plot figure
tsc_map_fig = ADRIA.viz.map(rs, m_tac, loc_clusters)
```

But internally this already does the `dropdims()`

https://github.com/open-AIMS/ADRIA.jl/blob/ed18747893f03769d24f5da7c81741068b9b2dc6/ext/AvizExt/viz/clustering.jl#L97

So we'd want to replace that with a call to `per_site()`

Sorry for the confusion

Zapiano: When I do `m_tac = dropdims(median(tac, dims=:scenarios), dims=:scenarios)` what I get is a: 
```
2-dimensional NamedDimsArray(KeyedArray(...)) with keys:
↓   timesteps ∈ 75-element Vector{Any}
→   sites ∈ 216-element Vector{Any}
```

ConnectedSystems: Sorry again for the confusion.

I've added a `loc_trajectory()` method using the same approach. This gets you what you need, and it's much faster than the previous approach.

Please also note I've updated `per_site` to `per_loc` (`loc` being short for "location").
I'm slowly standardizing away from "site" as this terminology is confusion.

src/metrics/site_level.jl
```diff
@@ -65,7 +65,7 @@ ADRIA.metrics.loc_trajectory(x -> quantile(x, 0.975), tac)
 - data : Data set to apply metric to
 
 # Returns
-Vector of \$N\$ elements, where \$N\$ is the number of sites.
+2D array of of \$T ⋅ S\$, where \$T\$ is total number of time steps and \$S\$ is number of sites
```
ConnectedSystems, line 68: Ah, forgot to update after copy/paste. Thanks for fixing.

src/metrics/site_level.jl
```diff
@@ -14,7 +14,7 @@ Get metric results applied to the location-level at indicated time (or across ti
 # Returns
 Named Vector of \$N\$ elements, where \$N\$ is the number of sites.
 """
-function per_loc(metric, data::NamedDimsArray{D,T,3,A})::NamedDimsArray where {D,T,A}
+function per_loc(metric, data::NamedDimsArray{D,T,N,A})::NamedDimsArray where {D,T,A}
```
ConnectedSystems, line 17: Ah, could you remove `where {D,T,A}` here please? It doesn't do anything here as I'm not specifying any types.

ConnectedSystems: Ah, no turns out you do need it, so add `N` to the list:

`where {D,T,N,A}`

src/metrics/site_level.jl
```diff
@@ -30,6 +28,53 @@ function per_site(metric, data::NamedDimsArray)::NamedDimsArray
 
     return NamedDimsArray(s, sites=axiskeys(data, :sites))
 end
-function per_site(metric, data::NamedDimsArray, timesteps::Union{UnitRange,Int64,Vector{Int64}})::NamedDimsArray
-    return per_site(metric, data[timesteps=timesteps])
+function per_loc(metric, data::NamedDimsArray{D,T,3,A}, timesteps::Union{UnitRange,Int64})::NamedDimsArray where {D,T,A}
+    return per_loc(metric, data[timesteps=timesteps])
+end
```
ConnectedSystems, line 33: I think same here: 

`data::NamedDimsArray{D,T,N,A}`
`where {D,T,N,A}`

src/metrics/site_level.jl
```diff
@@ -30,6 +28,53 @@ function per_site(metric, data::NamedDimsArray)::NamedDimsArray
...
+```
+
+# Arguments
+- metric : Any function (nominally from the Statistics package) to be applied to `data`
+- data : Data set to apply metric to
+
+# Returns
+2D array of of \$T ⋅ S\$, where \$T\$ is total number of time steps and \$S\$ is number of sites
+"""
+function loc_trajectory(metric, data::NamedDimsArray{D,T,3,A})::NamedDimsArray where {D,T,A}
```
ConnectedSystems, line 70: and again here

src/metrics/site_level.jl
ConnectedSystems: Don't forget to update docstrings as well

Zapiano:
@ConnectedSystems I think it's all good now. Just tested here and it is working.

ConnectedSystems (approved):
Confirm all tests pass so nothing unrelated seems to be affected. Thanks @Zapiano

PR #387, opened 2023-07-17
Consider near-term conditions with greater weight than far-future conditions

ConnectedSystems:
Puts greater weight on environmental conditions closer to the decision point than far future conditions when determining deployment locations


Rosejoycrocker:
Looks good to me and all tests passing. Any other changes before I merge?

ConnectedSystems:
Just adjusted the weighting factor from 0.99 to 0.95. 

All good to go!

Rosejoycrocker (approved):

PR #390, opened 2023-07-20
Changes to ode solver and time step for efficiency improvements

Rosejoycrocker:
The current ode solver is a BS method, which is an adaptive timestep 3rd order RK method. Adaptive timesteps can reduce efficiency in non-stiff ode solutions. Aditionally, higher order solvers are less efficient if there is neglible trade off in accuracy. Due to this fixed timesteps of varying precision with the lower order Euler scheme were investigated.

Timesteps of size 1/10,1/5, 1/3 and 1/2 were investigated with the Euler solver and the timings and RMSE in comparison to the original solver were compared.

Plot of average time for just the ode solver for 100 timesteps with 200 samples:
time_vs_solver

Plot of total time for 256 ADRIA scenarios for each solver:

time_vs_solver_total

Plot of RMSE between the original and Euler solvers:

RMSE_vs_time

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -283,7 +283,7 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     wave_scen::Matrix{Float64} = Matrix{Float64}(domain.wave_scens[:, :, wave_idx]) ./ maximum(domain.wave_scens[:, :, wave_idx])
 
     tspan::Tuple = (0.0, 1.0)
-    solver::BS3 = BS3()
+    solver::Euler = Euler()
 
     MCDA_approach::Int64 = param_set("guided")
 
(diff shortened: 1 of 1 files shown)


src/scenario.jl
```diff
@@ -565,7 +565,7 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
         p.X_mb .= tmp .* p.mb    # current cover * background mortality
 
         sol::ODESolution = solve(growth, solver, save_everystep=false, save_start=false,
-            alg_hints=[:nonstiff], abstol=1e-6, reltol=1e-6)  # , adaptive=false, dt=1.0
+            alg_hints=[:nonstiff], abstol=1e-6, reltol=1e-6,dt=0.5)  # , adaptive=false, dt=1.0
```
ConnectedSystems, line 568: Remove the commented code to avoid future confusion

```suggestion
            alg_hints=[:nonstiff], abstol=1e-6, reltol=1e-6,dt=0.5)
```

ConnectedSystems (commented):
Thanks Rose,

Suggest a minor change for clarity

src/scenario.jl
Rosejoycrocker: Thanks, just changed. Also sorry, I just realised I forgot to add you as reviewer

ConnectedSystems (approved):

PR #391, opened 2023-07-20
Add `adaptive=false` specification and remove tolerances

Rosejoycrocker:
Sorry to do a final PR for this - I forgot to add the 'adaptive=false` to specify a non-adaptive scheme. Also, the tolerances aren't used for non-adaptive schemes so they can be removed.

To confirm, the previous timings and RSMEs were compared for adaptive and non-adaptive schemes, I just forgot to add the flag in the final PR.


ConnectedSystems:
I expect the options that are not relevant for the given solver are ignored/turned off so leaving them in probably didn't have any effect.

That said, best to be clear about these things so I appreciate you removing them.

ConnectedSystems (approved):

PR #393, opened 2023-07-21
Scenario viz change

ConnectedSystems:
Place density plot with histogram.

image

Things left to do: 

- [x] Update documentation
- [x] Allow user to hide legend (looks nicer when composing multiple plots together)

### DIFF
--- a/docs/src/usage/analysis.md
+++ b/docs/src/usage/analysis.md
@@ -73,8 +73,8 @@ s_juves = ADRIA.metrics.scenario_relative_juveniles(rs)
 tf = Figure(resolution=(1600, 600))  # resolution in pixels
 
 # Implicitly create a single figure with 2 columns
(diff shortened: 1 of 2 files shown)


ConnectedSystems:
@Rosejoycrocker ready for review. Note I've updated example usage in the documentation, but not the corresponding figure. I intend to update all figures over the next few weeks.

Zapiano:
@ConnectedSystems have you considered using a Kernel Density plot instead of histogram? It can be easier to visualize when you only need the shape and there's intersection.

ConnectedSystems:
@Zapiano yes, that's essentially what we're moving away from.

Reason is the density plot can be interpreted in a misleading way (in my view).
Perhaps this will change after the MCDA assessments @Rosejoycrocker is doing currently.

image

Rosejoycrocker (approved):
Looks good to me

Issue #96, opened 2022-06-28
Coral evenness: Metric returns NaNs


ConnectedSystems:
Part of the calculation for evenness divides each species group by relative cover (RC).

This leads to NaNs when RC goes to 0.
Simple fix would be to replace NaNs with 0.

<img width="802" alt="image" src="https://user-images.githubusercontent.com/4075136/176090808-aa5eb864-a69e-4c17-a0d4-df6efd3ef4be.png">

@KRNA01 : Evenness was removed from the RCI metric as well. I remember there was a discussion between you and Ryan about it but cannot recall the reason for it. Do you remember?

PR #402, opened 2023-07-29
Updated Reef Indices

ConnectedSystems:
This PR adds updated/additional Reef Indices (Reef Condition Index, Reef Fish Index, Reef Tourism Index) based on work by Dr. R. Heneghan.

These indices are adapted to work with ADRIAmod outputs and require additional checks and confirmation they work as intended.

Principal differences include:
- For RCI, ADRIA does not have "rubble complementary" (or at least I'm not sure what that relates to)
- For RCI, expert elicited indicator values need to be reviewed and updated (I've used the older set of values for now, and have not checked if these differ from the most recent set)
- Regressions for RTI and RFI need to be double checked


ConnectedSystems:
Closes #96 (f4da073)

PR #408, opened 2023-08-02
Update ReefMod interface to match new species list

### LINKED ISSUE #399: Remove coral cover conversion for ReefModEngine interface
These lines need to be adjusted or removed as we're now representing the same species as ReefMod Engine:

https://github.com/open-AIMS/ADRIA.jl/blob/f906595d83de3e54df0f278d164413abe49b807e/src/ExtInterface/ReefMod/scenarios.jl#L9-L26

Rosejoycrocker:
Previously, the ReefMod extension required a change to initial covers as the species did not match those in ADRIA. Now that ADRIA has 6 species, the change is no longer required.

Resolves #399 .


src/ExtInterface/ReefMod/scenarios.jl
```diff
@@ -4,35 +4,8 @@
 Run ADRIA coral ecosystem model with ReefMod data.
 """
 function run_scenarios(::Type{ReefModDomain}, scens::DataFrame, dom::ReefModDomain, rcp::String)
```
ConnectedSystems, line 6: Hi Rose,

On reviewing what's left of the code, I think we don't need this entire function anymore?

It looks like we can call `run_scenarios()` as normal and it'll just work

Rosejoycrocker: Sure, will do

Rosejoycrocker: Removed the scenarios.jl file

Rosejoycrocker: So I tried loading the Domain and running scenarios.  Domain runs fine but there's a dimension mismatch when the scenarios are run so I'll look into fixing that before this is merged.

Rosejoycrocker:
Hi @ConnectedSystems. The dimension mismatch issue seems to be becuase the initial cover cover file for ReefMod has dimensions sites by number of species (nsites*6), whereas ADRIA's initial coral structure is sites by species and size classes (nsites*36). Do you know if the ReefMod initial data is aggregated across size classes (so we could split evenly across size classes for simplicity)? Otherwise I can ask Arne for more details.

ConnectedSystems:
This is weird, because this was working fine before. Could you please point me to where (file and line number) this error is raised?

Rosejoycrocker:
> This is weird, because this was working fine before. Could you please point me to where (file and line number) this error is raised?

It occurs for me at line 312 in `src/scenario.jl`, using the rme_ml_2023_03_30b domain.

ConnectedSystems:
Line numbers don't seem to line up with anything interesting.

Which branch are you on? Have you pulled the most recent changes?

Rosejoycrocker:
Sorry it's line 311 without an added breakpoint : https://github.com/open-AIMS/ADRIA.jl/blob/22f6a83449c481c28a9aeaa239222a63aaed1979/src/scenario.jl#L311

ConnectedSystems:
Hi Rose,

Please update `load_initial_cover(ReefModDomain, ...)` to do the necessary transpose. This was handled in `run_scenarios()` previously (which was poor design now that I think about it).

Rosejoycrocker:
Thanks Takuya, I've updated to transpose the data but there's still a dimension mismatch becasue the data is 6*nsites instead of 36*nsites. I think the previous code dealt with this by repeating the data but I'm not sure if this is appropriate for the ReefMod data- do you know if it is an aggregate of all size classes and should be split across size classes, or is it an average and so should be repeated across size classes? I can ask Arne if we're unsure.

ConnectedSystems:
Yes you're right - the values were interpolated previously.

I'm not sure how ReefMod handles this - they have 10 size classes I believe. 

For now, I would split it evenly across the six size classes we have, but yes, could you check with Arne or Ken?

Rosejoycrocker:
Thanks Takuya, I'll split evenly across size classes for now and I'll email Ken or Arne to check.

Rosejoycrocker:
Hi @ConnectedSystems , I asked Yves-Marie and Arne about how the files coral_spX_2022.csv are used to initialise size classes in ReefMod . A lognormal distribution is used to generate the area for different size classes, so that the total area adds up to the coral cover in coral_spx_2022.csv. Maybe our equivalent to this could be to weight each size class with weights which sum to one when we do the splitting of cover across size classes. Weights could be generated from a log-normal distribution where we choose the shape parameters to best match those of ReefMod. What do you think?

ConnectedSystems:
Is it not possible/sensible to use the same shape parameters?

Rosejoycrocker:
We woulduse the same shape parameters but they have more size classes so we'd have to split it up differently.

ConnectedSystems:
Makes sense to me, thank you

Rosejoycrocker:
I just asked Yves-Marie for where the shape parameters are for ReefMod and then I can implement this :)

Rosejoycrocker:
Hi @ConnectedSystems , Yves-Marie got back to me saying that the default distribution they use to initialise the size distribution is a log-normal distribution with mean log(700cm^2) and std log(4). What I've implemented takes the colony area at each of the bin edges in ADRIA's size classes and finds the density between each bin edge. These densities are then normalised and used to split the coral cover for each of the species into the size classes for ADRIA. An alternative would be to treat the largest size class as 60cm diameter and above and just integrate the density to infinity for the largest size class so you don't have to normalise. Let me know what you think is best.

The matrix reshaping I've used is a little convoluted so I'll try and think of another way to do it, but besides from that it's ready to review.

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,27 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
         icc_data[:, :, i] = Matrix(CSV.read(fn, DataFrame; drop=[1], header=false, comment="#"))
     end
 
+    # use ReefMod distribution for coral size class population (shape parameters have units log(cm^2))
+    # as suggested by YM (pers comm. 2023-08-08 12:55pm AEDT)
+    reef_mod_area_dist = LogNormal(log(700), log(4))
+    # get bin edges for coral diameters and transform to coral areas in cm^2
+    bin_edges_area = colony_mean_area(Float64[0, 2, 5, 10, 20, 40, 80])
```
ConnectedSystems, line 316: Please use sentence case for comments that are not inline.

Best practice is to comment not what is being done but _why_.
The occasions you include the what is when the code is necessarily complicated.

The comment for the code here for example is unnecessary. You should write more readable code (use informative variable names) instead.

```julia
# get bin edges for coral diameters and transform to coral areas in cm^2
bin_edges_area = colony_mean_area(Float64[0, 2, 5, 10, 20, 40, 80])
```

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,27 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
+
+    # find integral density between bounds of each size class areas
+    diff_cdf_integral = cdf.(reef_mod_area_dist, bin_edges_area[2:end]) .- cdf.(reef_mod_area_dist, bin_edges_area[1:end-1])
```
ConnectedSystems, line 319: Why not call `cdf()` for all of `bin_edges_area` once?

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,27 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
+    # normalise densities for each size class to create weightings
+    size_class_weights = diff_cdf_integral ./ sum(diff_cdf_integral)
```
ConnectedSystems, line 321: As with my previous comment - we can see from your variable name what this is - no need write a comment that repeats the same thing as the code.

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,27 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
+
     # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT)
     # Convert from percent to relative values
-    icc_data = dropdims(mean(icc_data, dims=2), dims=2) ./ 100.0
+    icc_data = ((dropdims(mean(icc_data, dims=2), dims=2)) ./ 100.0)
+
+    # Repeat species over each size class and reshape to give ADRIA compatible size (36 * nsites).
+    # Multiply by size class weights to give initial cover distribution over size class.
+    icc_data = reshape(repeat(icc_data, 6, 1), length(loc_ids), length(icc_files) * 6)' .* repeat(size_class_weights, 6)
```
ConnectedSystems, line 329: Agree this isn't the best approach, but I'll leave it to you to tidy up.

Is it not possible to apply the weights with L 325?

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,27 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
+    diff_cdf_integral = cdf.(reef_mod_area_dist, bin_edges_area[2:end]) .- cdf.(reef_mod_area_dist, bin_edges_area[1:end-1])
+    # normalise densities for each size class to create weightings
+    size_class_weights = diff_cdf_integral ./ sum(diff_cdf_integral)
+
     # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT)
     # Convert from percent to relative values
-    icc_data = dropdims(mean(icc_data, dims=2), dims=2) ./ 100.0
+    icc_data = ((dropdims(mean(icc_data, dims=2), dims=2)) ./ 100.0)
+
+    # Repeat species over each size class and reshape to give ADRIA compatible size (36 * nsites).
```
ConnectedSystems, line 327: As discussed previously, use the more general "locations" instead of "sites" please.

ReefMod does not work with sites.

```suggestion
    # Repeat species over each size class and reshape to give ADRIA compatible size (36 * n_locs).
```

ConnectedSystems (changes requested):

src/ExtInterface/ReefMod/Domain.jl
Rosejoycrocker: I tried doing this previously but you need the repeats of the coral species so that you get each of the size classes for each coral species, otherwise you just get one size class for each species. I could move the repeat up and then the final line is just reshaping and multiplying by weights...

ConnectedSystems: So `icc_data` and `size_class_weights` are not of compatible shapes?

Rosejoycrocker: icc data is n_locsX6 and size_class_weights is 6X1, but if you just multiply them together you get (species x* size class x) but not (species x * size class y). You need all permutations of  (species x * size class y), for each location, to get the final 36 species/size class dimension.

ConnectedSystems: How is `size_class_weights` of shape 6x1? Is it not a vector?

In any case:

```julia
n_locs = 50
n_species = 6
n_sc = 6
icc_data = rand(n_locs , n_species)  # 50 example locations
size_class_weights = rand(n_sc)

# Multiply each row of icc_data with weights with each element in size_class_weights
# resulting in a matrix (shape: 50 x 6) where each entry is a vector
# You could replace icc_data instead of assigning to x
x = icc_data .* [size_class_weights]

# You can verify correctness with:
all(icc_data[1, 1] .* size_class_weights .== x[1,1])

# Join vectors for each row and create matrix
Matrix(transpose(hcat(reduce.(vcat, eachrow(x))...)))

# ... or with pipes:
# hcat(reduce.(vcat, eachrow(x))...) |> transpose |> Matrix
```

Is that what you want?

Rosejoycrocker: Thanks Takuya, that's perfect. I didn't realise elementwise multiplication with square brakets gives a deconstructed dot product. I'll implement this and then it's ready for review.

Rosejoycrocker: Ready for review :)

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,26 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
 
-    # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT)
-    # Convert from percent to relative values
-    icc_data = dropdims(mean(icc_data, dims=2), dims=2) ./ 100.0
+    # Use ReefMod distribution for coral size class population (shape parameters have units log(cm^2))
+    # as suggested by YM (pers comm. 2023-08-08 12:55pm AEDT)
+    reef_mod_area_dist = LogNormal(log(700), log(4))
+    bin_edges_area = colony_mean_area(Float64[0, 2, 5, 10, 20, 40, 80])
+
+    # Find integral density between bounds of each size class areas
```
ConnectedSystems, line 317:
```suggestion
    # Use integral density between bounds of each size class areas to create weights for each size class
```

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,26 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
         icc_data[:, :, i] = Matrix(CSV.read(fn, DataFrame; drop=[1], header=false, comment="#"))
     end
 
-    # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT)
-    # Convert from percent to relative values
-    icc_data = dropdims(mean(icc_data, dims=2), dims=2) ./ 100.0
+    # Use ReefMod distribution for coral size class population (shape parameters have units log(cm^2))
+    # as suggested by YM (pers comm. 2023-08-08 12:55pm AEDT)
+    reef_mod_area_dist = LogNormal(log(700), log(4))
+    bin_edges_area = colony_mean_area(Float64[0, 2, 5, 10, 20, 40, 80])
```
ConnectedSystems, line 315: Please detail what the distribution is used for (the "why")

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,26 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
...
+    # Convert from percent to relative values.
+    icc_data = ((dropdims(mean(icc_data, dims=2), dims=2)) ./ 100.0)
+
+    # Repeat species over each size class and reshape to give ADRIA compatible size (36 * n_locs).
+    # Multiply by size class weights to give initial cover distribution over each size class.
+    icc_data = Matrix(hcat(reduce.(vcat, eachrow(icc_data .* [size_class_weights]))...))
 
     # Reorder dims to: locations, species
-    return NamedDimsArray(icc_data, locs=loc_ids, species=1:length(icc_files))
+    return NamedDimsArray(icc_data, species=1:(length(icc_files)*6), locs=loc_ids)
```
ConnectedSystems, line 331: This isn't reordering dimensions is it?

ConnectedSystems:
One last question: Have you tried running ADRIAmod run with RME data now?

Rosejoycrocker:
> One last question: Have you tried running ADRIAmod run with RME data now?

Yes, I ran a small number of scenarios and they run fine.

> One last question: Have you tried running ADRIAmod run with RME data now?

This is what it looks like for a small number of scenarios:
ReefMod_ADRIA_runs

ConnectedSystems (approved):

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -309,12 +309,26 @@ function load_initial_cover(::Type{ReefModDomain}, data_path::String, loc_ids::V
         icc_data[:, :, i] = Matrix(CSV.read(fn, DataFrame; drop=[1], header=false, comment="#"))
     end
 
-    # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT)
-    # Convert from percent to relative values
-    icc_data = dropdims(mean(icc_data, dims=2), dims=2) ./ 100.0
+    # Use ReefMod distribution for coral size class population (shape parameters have units log(cm^2))
+    # as suggested by YM (pers comm. 2023-08-08 12:55pm AEDT). Distribution is used to split ReefMod initial
```
ConnectedSystems, line 313: That time stamp looks suspicious. If it was this month, it can't have been AEDT?

```suggestion
    # as suggested by YM (pers comm. 2023-08-08 12:55pm AEST). Distribution is used to split ReefMod initial 
```

PR #411, opened 2023-08-04
Improvements to run_site_selection to make the function more user friendly

Rosejoycrocker:
Opening this pull request as a draft to discuss improvements to make run_site_selection() more user friendly. The current PR fixes bugs due to upstream changes in ADRIA and adds two aggregation functions which can be specified as output to run_site_selection.

The 2 aggregation functions allow the output to put given as frequency of site selection over the scenarios considered (for seeding and shading), or as the complete ordering of sites for each time step and scenario (using ordered site ids instead of ranks matched to site ids).

Suggestions for any other desirable aggregation functions or structural changes which will make the function more user friendly are appreciated.


src/decision/location_selection.jl
```diff
@@ -0,0 +1,151 @@
+
+"""
+    site_selection(domain::Domain, scenario::DataFrameRow{DataFrame,DataFrames.Index}, w_scens::NamedDimsArray, dhw_scens::NamedDimsArray, site_ids::Vector{Int64}, sum_cover::AbstractArray, area_to_seed::Float64)
+
+Perform site selection using a chosen mcda aggregation method, domain, initial cover, criteria weightings and thresholds.
+
+# Arguments
```
Zapiano, line 7: The argument `domain` is missing in this list and the argument `mcda_vars` does not exist

src/decision/location_selection.jl
```diff
@@ -0,0 +1,151 @@
...
+- `sum_cover` : array of size (number of scenarios * number of sites) containing the summed coral cover for each site selection scenario.
+- `area_to_seed` : area of coral to be seeded at each time step in km^2
+- `target_seed_sites` : list of candidate locations for seeding (indices)
+- `target_shade_sites` : list of candidate location to shade (indices)
+
+# Returns
+- `ranks_store` : number of scenarios * sites * 3 (last dimension indicates: site_id, seed rank, shade rank)
+    containing ranks for each scenario run.
+"""
+function run_site_selection(dom::Domain, scenarios::DataFrame, sum_cover::AbstractArray, area_to_seed::Float64; aggregate_function=nothing,
```
Zapiano, line 61: The arguments don't match the Docstring (both at the name of the function and at the list of arguments)

src/decision/location_selection.jl
```diff
@@ -0,0 +1,151 @@
...
+    ranks_to_location_order(ranks::NamedDimsArray, iv_type::String)
+
+
+Post-processing function for location ranks output of `run_location_selection()`. Gives the order 
+of location preference for each scenario as location ids.
+
+# Arguments
+- `ranks` : Contains location ranks for each scenario of location selection, as created by 
+    `run_location_selection()`.
+- `iv_type` : String indicating the intervention type to perform aggregation on.
```
Zapiano, line 120: Just a personal opinion here, but I think the name of this variable can be confusing. Maybe `interv_type` or `int_type` (or just `intervention_type`) would make the code easier to read (in my personal opinion). :)

Rosejoycrocker: Hi @Zapiano, thanks for your comments, I'll incorporate them as I move this PR from draft. Apologies for some missing variables and documentation, I decided to split up a file and haven't cleaned up the structure yet.

 The PR is still a work in progress but I was hoping to get any requests/opinions on aggregate functions for the function `run_site_selection()` which are helpful for analysis or visualisation purposes. The current function outputs ranks, which gives where each site was ranked in terms of suitability for each intervention and scenario. This is not the most intuitive output to analyse, so I've added a rank frequency aggregation function (which gives the frequency with which each site was selected for a particular rank over all scenarios) and a rank order aggregation function (which lists the complete order of sites for each scenario). Let me know if you have other ideas for aggregation functions which could lead to interesting analysis of mcda outcomes :)

Zapiano: @Rosejoycrocker oh, sorry, I haven't noticed it was a draft, my bad!

About the aggregation functions: thanks for the explanation! I was trying to understand the output of this function and what you just wrote helped a lot! I will think about it and let you know if I have any suggestions - or questions! :)

Rosejoycrocker: That's all good, the comments are good anyway because the PR aims to make the function more intuitive to use, so it's all helpful :) Any other ideas for improving the usability of the function are welcome, particularly any aggregation methods you think might be useful for other anlyses you're working on.

ConnectedSystems: On the variable name: I'd prefer `interv_type` or to a smaller degree `intervention_type` (too verbose in my opinion.

Definitely not `int_type` because in my mind that relates to integer types (uint8, int16, int64, etc)

ConnectedSystems: Oh just to say, I was okay with `iv_type` - `iv` is the shorthand the ReefMod team uses so it's an established convention.

Rosejoycrocker: Oh right, sorry misread the above. I'll change back to iv_type because it could be preferrable when we move to a more generic structure where intervention type will probably be a more widely used variable.

ConnectedSystems:
Could you add an example of how you use these functions in this discussion?

Rosejoycrocker:
Currently, the aggregate function is given as an optional argument to `run_site_selection()` 

```julia
ranks, rank_frequencies_seed, rank_frequencies_shade = run_site_selection(dom, scens, sum_cover, area_to_seed, aggregate_function=ranks_to_frequencies, target_seed_sites=target_seed_sites)
```
Or the aggregate function can be used as a standalone:
```jullia 
rank_frequencies_seed = ranks_to_frequencies(rs.ranks, "seed) 
```

The function gioves each timestep if you have a timeseries for ranks, but I think I'll add an aggreagte dimension option so you specify if you want to aggregate over some dimensions.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,169 @@
...
+        sites=domain.site_ids,
+        intervention=1:2,
+        scenarios=1:nrow(scenarios),
+    )
+
+    dhw_scens = domain.dhw_scens
+    wave_scens = domain.wave_scens
+
+    # Pre-calculate maximum depth to consider
+    scenarios[:, "max_depth"] .= scenarios.depth_min .+ scenarios.depth_offset
```
ConnectedSystems, line 76: Avoid modifying the user provided dataframe. You don't need `max_depth` as a column anyway

ConnectedSystems:
Rose, could you please rebase onto main please? Commits from a different PR seem to be mixed in here.

A few general comments while you're at it:

1. Returning a different number of results depending on the inputs is not a great way of doing this. Either always return the three result types, split the function up, or use multiple dispatch to handle your case. The code will be much cleaner if you do this.

2. There's copies of code, or very similar code, in `run_site_selection()` from other parts of ADRIA (use of the same calculation to consider mean + stdev of projected environmental conditions, for example). It might be okay in this specific circumstance because I'm not sure that will change, but in general duplication of code makes more work for us - instead of making changes in one place, now we have to remember that there are two places in the codebase we have to update if any modifications are necessary. 

3. Could you add an example of what the user would do with the data returned from this function?

Rosejoycrocker:
> Rose, could you please rebase onto main please? Commits from a different PR seem to be mixed in here.
> 
> A few general comments while you're at it:
> 
> 1. Returning a different number of results depending on the inputs is not a great way of doing this. Either always return the three result types, split the function up, or use multiple dispatch to handle your case. The code will be much cleaner if you do this.
> 2. There's copies of code, or very similar code, in `run_site_selection()` from other parts of ADRIA (use of the same calculation to consider mean + stdev of projected environmental conditions, for example). It might be okay in this specific circumstance because I'm not sure that will change, but in general duplication of code makes more work for us - instead of making changes in one place, now we have to remember that there are two places in the codebase we have to update if any modifications are necessary.
> 3. Could you add an example of what the user would do with the data returned from this function?

Yeah I'm not sure why that happened, I just rebased but it seems to have made it worse.... I'll address your other points shortly

PR #412, opened 2023-08-06
Small miscellaneous changes for scenario visualizations

ConnectedSystems:
Minor changes to allow scenario subsetting and correcting typos in docstrings

### DIFF
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -11,7 +11,7 @@ const COLORS::Dict{Symbol,Symbol} = Dict(
 )
 
 
-function scenario_type(rs)
+function scenario_type(rs; scenarios=(:))
     inputs = rs.inputs
 
     no_seed = (inputs.N_seed_TA .== 0) .& (inputs.N_seed_CA .== 0) .& (inputs.N_seed_SM .== 0)
@@ -26,7 +26,7 @@ function scenario_type(rs)
     # Guided scenarios must be the inverse of unguided/counterfactual scenarios
     guided = Bool.(ones(Int64, size(inputs, 1)) .⊻ (counterfactual .| unguided))
 
-    return (counterfactual=counterfactual, unguided=unguided, guided=guided)
+    return (counterfactual=counterfactual[scenarios], unguided=unguided[scenarios], guided=guided[scenarios])
 end
 
 function scenario_colors(rs, weight::Float64, hide::BitVector)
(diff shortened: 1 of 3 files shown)


ext/AvizExt/theme.jl
```diff
@@ -11,7 +11,7 @@ const COLORS::Dict{Symbol,Symbol} = Dict(
 )
 
 
-function scenario_type(rs)
+function scenario_type(rs; scenarios=(:))
```
Zapiano, line 14: @ConnectedSystems I don't get what does that `(:)` mean. Can you explain to me, please?

ConnectedSystems: colon operator to select all, e.g.,

```
some_dataframe[:, 1]
```

ext/AvizExt/viz/scenario.jl
```diff
@@ -4,6 +4,21 @@
 
 Plot scenario outcomes over time.
```
Zapiano, line 5: @ConnectedSystems 

Since you are editing this docstring: if I understand correctly, this function plots one specific metric for all the scenarios over time making a distinction between guided, unguided and no intervention. If that's correct, I think we could (at least) improve the description here a little bit..? (But I would argue that the name of this function could be changed to something like `scenarios_metric` or just `metric` maybe)

Zapiano:
@ConnectedSystems  I run the tests and they passed, I run this function and it's working and I agree with the changes. The only comment I have is the above, but is not a blocker. I'm going to wait for your answer.

ext/AvizExt/viz/scenario.jl
ConnectedSystems: I'm open to a rename but I disagree with the name change (at least the one suggested here).

`ADRIA.viz.scenario()` indicates you'd like to visualize the scenario (specifically, the `viz.scenario`).

How about `ADRIA.viz.scenario_outcomes()` ?

As for the description, I'm not sure what other text should be included. What were you thinking of adding?

Zapiano: My comment was more directed to the docstring than to the name of the function actually. I am more asking a question and only suggesting a change depending on the answer. What is the difference between a metric and an outcome? My confusion was that, because the variable `y` represent the "results of scenario metric", and we are plotting `y`, then to me we are plotting the scenario metric. 

But I don't think this is a blocker so if you prefer, I can approve this PR and we can have this discussion separately.

ConnectedSystems:
Hi Pedro, please go ahead and merge.

Zapiano (approved):

PR #414, opened 2023-08-08
Remove dependency on ImageMagick in favour of ImageIO

ConnectedSystems:
The ImageMagick library seems to have issues on Apple M1/M2 chipsets.

Thankfully, there's an alternative available, so this PR removes ImageMagick in favour of ImageIO.


ConnectedSystems:
Merging as this will not affect modelling processes as such, and willing to take the blame for anything that breaks.

Issue #405, opened 2023-07-31
ADRIAmod uses excessive amounts of memory during runs


ConnectedSystems:
ADRIAmod uses up to 99% memory when doing larger numbers of runs.

The method used to run scenarios is to create an anonymous function (i.e., a closure) to which we pass in each row in the scenario spec. `pmap` is used regardless of whether parallelism is active or not (where a single worker is used for single core use cases).

A named tuple of matrices used as intermediate caches throughout a simulation is used to reduce memory allocation. The size of these matrices are dependent on the number of species/locations represented and can be quite large.

Additionally, it seems that `pmap` will naively serialize data to send to each worker for every job. Using `CachingPool()` avoids this (as indicated in: https://github.com/JuliaLang/julia/pull/33892) and is set to be the default behavior (in Julia v1.10 I believe). This means currently these large cache matrices get reserialized for every scenario run. The Julia's garbage collector doesn't seem to clean these up very often, so they hang around, which can then lead to out-of-memory issues.

I have successfully run 8192 scenarios with just `CachingPool()` active (comparably high memory use, but Julia does not crash), and am trying creating the cache data structures within each scenario run to avoid their serialization.

The reserialization might explain why @Rosejoycrocker sees Julia crash even for single core runs (as `pmap()` is used regardless of whether runs are done with single or multiple cores).

A smarter strategy may be needed, where the regular `map()` is used for single core runs. The advantage is that in such cases the cache is created once and reused for every run, reducing memory use in single-core contexts.

PR #416, opened 2023-08-10
Fix workers crashing and leaking memory when running scenarios in parallel

### LINKED ISSUE #413: Errors when running scenarios in parallel when nworkers() > 1
https://github.com/open-AIMS/ADRIA.jl/blob/6bdbe38171100ec2cc7b1870467aba1adabf6adf/src/scenario.jl#L126C1-L130C16

When running more than 4095 scenarios the function `run_scenarios` will run in the scenarios in parallel. If nworkers() using more than one worker, `fun` won't be defined, but then at the lines 127 and 129 it will be called, raising the following error at line 127 of this file: `ERROR: UndefVarError: 'func' not defined`

ConnectedSystems:
Closes #413 and re-addresses #405.

Previous approach to parallel runs first initialized the provided domain object with environmental data (via `switch_RCPs!()`).
The `domain` was then sent to workers for each scenario run.

This had two effects:
1. For each run the domain object would be serialized and sent to workers, which unserialized the data on receipt to begin its workload. Due to the size of the domain data, this caused significant overhead
2. Because an explicit copy of the domain is not sent to each worker, I suspect each worker received a reference to the original data structure (although the data associated with the struct would be a copy).

Because of the above, Julia's garbage collector (GC) would recognise the Domain as still in use after a worker had finished a run, so it never cleans up the "copy" of the domain. This leads to ever-increasing amounts of memory to be used. Forcing the GC to run to free up memory (as in #405) only masked the issue.

To resolve both issues, we now send to each worker an uninitialized domain. As this data structure is not loaded with the environmental data, it is much smaller and thus faster to serialize/unserialize, reducing overheads. Each worker then configures the domain with the appropriate environmental data. For clarity, this PR still forces the GC to run (albeit much less than before), but one source of memory leaks has been addressed.

This PR also includes a fix for a zero division error bug, which stems from cases where MCDA returns ranks for `n_sites` but seeding deployments are distributed to < `n_sites`.

It also resolves an issue where an off-by-one error would very rarely arise when determining planning horizons.

### DIFF
--- a/src/ADRIA.jl
+++ b/src/ADRIA.jl
@@ -105,6 +105,10 @@ end
 
     @compile_workload begin
 
+        # Force precompiling of code that handles distributed infrastructure
+        addprocs(1)
+        @everywhere 1 + 1
+
         f() = begin
             @showprogress 1 for _ in 1:10
             end
--- a/src/ADRIA.jl
+++ b/src/ADRIA.jl
@@ -119,8 +119,12 @@ end
         ADRIA.sample(dom, 16)
         ADRIA.model_spec(dom)
 
+        ENV["ADRIA_DEBUG"] = "false"
+
         p_df = ADRIA.param_table(dom)
         rs1 = ADRIA.run_scenario(p_df[1, :], dom)
+
+        delete!(ENV, "ADRIA_DEBUG")
 
         # ENV["ADRIA_THRESHOLD"] = 1e-6
         # run_scenario(p_df[1, :], dom)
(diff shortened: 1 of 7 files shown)


ConnectedSystems:
Found another memory leak issue due to recent changes, so marking this as a draft until I can resolve it.

ConnectedSystems:
Small update: Memory leak seems to be stemming from the process used to represent natural adaptation / heritability. Looking into reworking the code to resolve the issue.

ConnectedSystems:
Last thing to do is to make the logging of species DHW tolerance optional as it makes the disk space requirements for very large runs extremely high (10's of gigabytes)

Zapiano (approved):

ConnectedSystems:
@Rosejoycrocker FYI you should be fine to go ahead and do runs now after pulling most recent dev version of ADRIA

PR #417, opened 2023-08-11
Adjustments to heritability based on feedback

ConnectedSystems:
Adjusting representation of natural adaptation and heritability based on feedback and discussion (thanks to @ecolology), principally:

Minimum bound of truncated distribution for each size class is the minimum of the Mixture distribution, such that over time corals are less likely to die from low DHW events. This makes the (truncated) normal distribution behave somewhat more like a log-normal distribution.

image

**Note: the CDF above is prior to any adjustments based on depth. Actual mortality may be significantly less**

Standard deviation is now fixed to the original value used for a given size class. Standard deviations does still change over time (we see an increase) due to the Mixture and truncation approach, indicating a greater variability in DHW sensitivity over time.

The +20 from mean rule used as a hypothetical upper bound of the DHW tolerance distribution is adjusted to be +10 from mean.

Corrected size classes that are deemed to be reproductive (fecund): size classes 4 - 6, not 3 - 6

Renamed variables for conceptual alignment: the prototype was based on $t$ to $t+1$, where as ADRIAmod works on $t-1$ to $t$

### DIFF
--- a/src/ecosystem/corals/Corals.jl
+++ b/src/ecosystem/corals/Corals.jl
@@ -214,7 +214,7 @@ function coral_spec()::NamedTuple
         ], inner=n_classes)
 
     # Data from Hughes et al., (2018) were based on the 2016 bleaching event.
-    # We some further adaptation (+3.0 DHW) has occurred since then.
+    # We assume some further adaptation (+3.0 DHW) has occurred since then.
     params.dist_mean .+= 3.0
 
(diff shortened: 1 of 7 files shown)


ConnectedSystems:
@Rosejoycrocker @Zapiano 

Going to bypass the usual review process and merge this in. Let's meet so I can walk you through how this all works.

PR #420, opened 2023-08-14
Implement relevant struct to handle SIRUS rule extraction

Zapiano:
The struct Rule together with some functions to get information once a rule is extracted from SIRUS.


Project.toml
```diff
@@ -48,11 +49,13 @@ ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 RelocatableFolders = "05181044-ff0b-4ac5-8273-598c1e38db00"
+SIRUS = "cdeec39e-fb35-4959-aadb-a1dd5dede958"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 SimpleWeightedGraphs = "47aef6b3-ad0c-573a-a1e2-d07658019622"
 SparseArrayKit = "a9a3c162-d163-4c15-8926-b8794fbefed2"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
ConnectedSystems, line 58: You'll also need to add `compat` entries to indicate what package versions you (currently) expect

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
+using SIRUS, MLJ
+using StableRNGs: StableRNG
+import ..performance: temporal_variability
+
+"""
+    Rule{V<:Vector{Float64},W<:Vector{Vector{Any}}}
+
+# A Rule contains a condition (vector of conditional clauses) and consequent (vector
```
ConnectedSystems, line 8: Stray header marker (`#`)

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
...
+end
+
+"""
+    _condition(rules::SIRUS.StableRules{Float64}, index::Int64)
+
+Vector containing condition clauses. Each condition clause is a vector with three 
+components: a feature_name::String, a direction::String (< or ≤) and a value:<Float64 
+
+# Arguments
+- rules - SIRUS.StableRules object containing all rules
```
ConnectedSystems, line 38: Use colon instead of dash. This goes for all docstrings.

```suggestion
- rules : SIRUS.StableRules object containing all rules
```

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
...
+Vector of Rule objects. These are **not** the same as SIRUS.Rule object.
+
+# Arguments
+- `rules` - SIRUS.StableRules object
+
+# Return
+Vector of Rule objects
+"""
+function rules(rules::SIRUS.StableRules{Float64})
+    [Rule(_condition(rules, i), _consequent(rules, i)) for i in eachindex(rules.rules)]
```
ConnectedSystems, line 28: I prefer explicit return statements, though here I'm okay with it as it's a single line...

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
...
+- rules - SIRUS.StableRules object containing all rules
+- `index` - Index of the rule
+
+# Return
+Vector of vectors of rule conditions.
+"""
+function _condition(rules::SIRUS.StableRules{Float64}, index::Int64)
+    feature_name(split) = split.splitpoint.feature_name
+    direction(split) = split.direction
+    value(split) = split.splitpoint.value
```
ConnectedSystems, line 47: Move these outside the function.
They will be redefined every time you call `_condition()`

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
...
+- `index` - Index of the rule
+
+# Return
+Vector of rule consequents
+"""
+function _consequent(rules::SIRUS.StableRules{Float64}, index::Int64)
+    weight = rules.weights[index]
+    rule = rules.rules[index]
+    then_probability = SIRUS._simplify_binary_probabilities(weight, rule.then)
+    otherwise_probability = SIRUS._simplify_binary_probabilities(weight, rule.otherwise)
```
ConnectedSystems, line 69: Relying on another package's private functions 😬 

Okay if there's no other way, but it does make me feel uncomfortable

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,88 @@
...
+
+# Arguments
+- `rules` - Vector of Rule objects
+"""
+function print_rules(rules::Vector{Rule{Vector{Float64},Vector{Vector{Any}}}})
+    condition(rule) = [c == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
+    consequent(rule) = " then $(rule.consequent[1]) else $(rule.consequent[2])\n"
+    rule_string(rule) = string("if ", join(condition(rule), " "), consequent(rule))
+    print(join([rule_string(rule) for rule in rules]))
+end
```
ConnectedSystems, line 88: This is great. Consider opening a PR to the SIRUS package too!

src/metrics/performance.jl
```diff
@@ -97,8 +97,8 @@ julia> temporal_variability(x, temporal_variabilty, P(x), D(x), E(x))
 function temporal_variability(x::AbstractVector{<:Real}; w=[0.5, 0.5])
     return mean([median(x), 1.0 - gmd(x)], weights(w))
 end
-function temporal_variability(x::AbstractArray{<:Real,2})
-    return temporal_variability.(eachcol(x))
+function temporal_variability(x::AbstractArray{<:Real,2}; w=[0.5, 0.5])
```
ConnectedSystems, line 100: Let's default to `[0.9, 0.1]` instead

src/analysis/rule_extraction.jl
Zapiano: https://github.com/rikhuijzer/SIRUS.jl/blob/78d74aa1cebfaf564fa6e547d03e0f8a3f96a7d4/src/ruleshow.jl#L31

This is where this function is being defined. The only other way around this I see is to define the same function inside ADRIA... which is not great too. What do you think?

ConnectedSystems: Agree that neither option is great.

Redefining the function inside ADRIA works so long as other aspects of SIRUS don't change.
Relying on SIRUS' private functions will only work so long as nothing in these functions changes (especially returned types).

I think leave it as is for now. Future Pedro can fix if anything breaks 😛

src/analysis/rule_extraction.jl
ConnectedSystems: Just in case you weren't aware, string concatenation operator in Julia is `*`

e.g., "this string plus" * " another string"

https://groups.google.com/g/julia-dev/c/4K6S7tWnuEs/m/RF6x-f59IaoJ

Wrapping everything inside a call to `string()` also works of course, not an issue.

ConnectedSystems: I somehow missed that there are inner functions defined here too.
Maybe move them outside and make them private so they don't conflict with variable names?

Not sure if this is a big issue though - I don't think we'd be printing rules too often

Zapiano: Cool, I didn't know that!

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,92 @@
...
+"""
+struct Rule{V<:Vector{Float64},W<:Vector{Vector{Any}}}
+    condition::W
+    consequent::V
+end
+
+"""
+    rules(rules::SIRUS.StableRules)
+
+Vector of Rule objects. These are **not** the same as SIRUS.Rule object.
```
ConnectedSystems, line 19:
```suggestion
Returns vector of Rule objects. These are **not** the same as SIRUS.Rule object.
```

src/analysis/rule_extraction.jl
Zapiano: Do you think we need docstrings for these too? I thought you had seen then but because we won't call them too often you'd considered that it wouldn't be a problem.

src/analysis/rule_extraction.jl
```diff
@@ -0,0 +1,92 @@
...
+"""
+    print_rules(rules::Vector{Rule})
+
+Print all rules in a human readable way. SIRUS also has this functionality but 1) it only 
+prints once, when the model is fitted and 2) this function prints the rules in a more 
+compact form.
+
+# Arguments
+- `rules` : Vector of Rule objects
+"""
```
ConnectedSystems, line 86: Maybe in a future clean up we can add an example to show what this "human readable" way looks like. But this is fine for now.

ConnectedSystems (approved):
Looks good, will merge later this afternoon.

src/analysis/rule_extraction.jl
ConnectedSystems: If they're moved out, no, as they'll be private and anyone should be able to figure out what they do in the context of their use.

Also no if they are kept in. And yes, I don't see it being a computational bottleneck, so okay to leave as is.

PR #421, opened 2023-08-15
Implement rule extracting function

### LINKED ISSUE #404: Add Rule Induction (with SIRUS)

Zapiano:
Implement necessary functions to use Julia implementation of SIRUS rule extraction algorithm (https://github.com/rikhuijzer/SIRUS.jl).

Closes #404.

### DIFF
--- a/src/analysis/rule_extraction.jl
+++ b/src/analysis/rule_extraction.jl
@@ -6,7 +6,9 @@ import ..performance: temporal_variability
     Rule{V<:Vector{Vector},W<:Vector{Float64}}
 
 A Rule contains a condition (vector of conditional clauses) and consequent (vector
-of values for then and else results of the condition)
+of values for then and else results of the condition). Conditional clauses represent
+'a < b' statements. Consequents contain the probabilities of a observation begin
+past of the 'positive class' given that the condition is true or false.
 """
 struct Rule{V<:Vector{Vector},W<:Vector{Float64}}
     condition::V
@@ -17,14 +19,15 @@ end
     rules(rules::SIRUS.StableRules{Float64})
 
 Collates vector of Rule objects. These are **not** the same as SIRUS.Rule object.
+See also [`Rule`](@ref).
 
 # Arguments
-- `rules` : SIRUS.StableRules object
+- `rules` : SIRUS.StableRules object containing all rules
 
 # Returns
 Vector{ADRIA.analysis.Rule{Vector{Float64}, Vector{Vector}}}
 """
-function rules(rules::SIRUS.StableRules{Float64})
+function rules(rules::SIRUS.StableRules{Float64})::Vector{Rule{Vector{Vector},Vector{Float64}}}
     [Rule(_condition(rules, i), _consequent(rules, i)) for i in eachindex(rules.rules)]
 end
 
@@ -35,14 +38,13 @@ Vector containing condition clauses. Each condition clause is a vector with thre
 components: a feature_name::String, a direction::String (< or ≤) and a value:<Float64
 
 # Arguments
-- rules : SIRUS.StableRules object containing all rules
+- `rules` : SIRUS.StableRules object containing all rules
 - `index` : Index of the rule
 
 # Returns
-Vector{Vector}
-Vector of vectors of rule conditions.
+Vector of Rule condition clauses (each one being a vector itself).
 """
-function _condition(rules::SIRUS.StableRules{Float64}, index::Int64)
+function _condition(rules::SIRUS.StableRules{Float64}, index::Int64)::Vector{Vector}
     condition::Vector{Vector} = []
     for split in rules.rules[index].path.splits
         feature_name = split.splitpoint.feature_name
@@ -64,9 +66,9 @@ probability of the 'else' clause (here called otherwise)
 - `index` : Index of the rule
 
 # Returns
-Vector{Float64}
+Probabilities vector, one for Rule condition == true, one for Rule condition == false.
 """
-function _consequent(rules::SIRUS.StableRules{Float64}, index::Int64)
+function _consequent(rules::SIRUS.StableRules{Float64}, index::Int64)::Vector{Float64}
     weight = rules.weights[index]
     rule = rules.rules[index]
     then_probability = SIRUS._simplify_binary_probabilities(weight, rule.then)
@@ -98,11 +100,12 @@ end
 Use SIRUS package to extract rules from time series clusters based on some summary metric (default is median)
 
 # Arguments
-- `clusters` - Vector of cluster indexes for each scenario outcome
-- `X::DataFrame` - DataFrame with factors to be used as input by SIRUS
-- `outcomes` - Matrix of some metric over time for all scenarios
-- `max_rules` - Maximum number of rules, to be used as input by SIRUS
-- `n_trees` - Number of trees to be created by SIRUS algorithm
+- `clusters` : Vector of cluster indexes for each scenario outcome
+- `X` : Features to be used as input by SIRUS
+- `outcomes` : Matrix of some metric over time for all scenarios
+- `max_rules` : Maximum number of rules, to be used as input by SIRUS
+- `n_trees` : Number of trees to be created by SIRUS algorithm
+- `seed` : Seed to be used by RGN
 
 # Returns
 A StableRules object (implemented by SIRUS).
@@ -135,13 +138,13 @@ function cluster_rules(clusters::Vector{T}, X::DataFrame, outcomes::AbstractMatr
 end
 
 """
-    robust_cluster(clusters::Vector{T}, result_set::ResultSet) where {T<: Integer}
+    target_cluster(clusters::Vector{T}, result_set::ResultSet) where {T<: Integer}
 
 Find most robust cluster.
 
 # Arguments
-- `clusters` - Vector with outcomes cluster indexes
-- `outcomes` - AbstractMatrix of outcomes factors
+- `clusters` : Vector with outcomes cluster indexes
+- `outcomes` : AbstractMatrix of outcomes features
 
 # Returns
 Index of the most robust cluster
--- a/src/analysis/rule_extraction.jl
+++ b/src/analysis/rule_extraction.jl
@@ -3,9 +3,9 @@ using StableRNGs: StableRNG
 import ..performance: temporal_variability
 
 """
-    Rule{V<:Vector{Vector{Any}},W<:Vector{Float64}}
+    Rule{V<:Vector{Vector},W<:Vector{Float64}}
 
-A Rule contains a condition (vector of conditional clauses) and consequent (vector 
+A Rule contains a condition (vector of conditional clauses) and consequent (vector
 of values for then and else results of the condition)
 """
 struct Rule{V<:Vector{Vector},W<:Vector{Float64}}
@@ -21,8 +21,8 @@ Collates vector of Rule objects. These are **not** the same as SIRUS.Rule object
 # Arguments
 - `rules` : SIRUS.StableRules object
 
-# Return
-Vector{ADRIA.analysis.Rule{Vector{Float64}, Vector{Vector{Any}}}}
+# Returns
+Vector{ADRIA.analysis.Rule{Vector{Float64}, Vector{Vector}}}
 """
 function rules(rules::SIRUS.StableRules{Float64})
     [Rule(_condition(rules, i), _consequent(rules, i)) for i in eachindex(rules.rules)]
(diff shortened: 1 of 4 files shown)


src/analysis/rule_extraction.jl
```diff
@@ -3,26 +3,26 @@ using StableRNGs: StableRNG
...
 
-A Rule contains a condition (vector of conditional clauses) and consequent (vector 
+A Rule contains a condition (vector of conditional clauses) and consequent (vector
 of values for then and else results of the condition)
 """
-struct Rule{V<:Vector{Float64},W<:Vector{Vector{Any}}}
-    condition::W
-    consequent::V
+struct Rule{V<:Vector{Vector},W<:Vector{Float64}}
+    condition::V
```
ConnectedSystems, line 12: Does this have to be flexible to `Any`? Or could we tighten the expect type to:

`struct Rule{V<:Vector{Vector{StableRules}},W<:Vector{Float64}}`

src/analysis/rule_extraction.jl
```diff
@@ -3,26 +3,26 @@ using StableRNGs: StableRNG
...
 
 # Arguments
 - `rules` : SIRUS.StableRules object
 
-# Return
-Vector{ADRIA.analysis.Rule{Vector{Float64}, Vector{Vector{Any}}}}
+# Returns
+Vector{ADRIA.analysis.Rule{Vector{Float64}, Vector{Vector}}}
 """
 function rules(rules::SIRUS.StableRules{Float64})
```
ConnectedSystems, line 27: Docstring should detail purpose more so than the technical "what".

Suggest adding a return type to function signature:

```julia
function rules(rules::SIRUS.StableRules{Float64})::Vector{Rule{Vector{Float64}, Vector{Vector}}}
```

Or at least:

```julia
function rules(rules::SIRUS.StableRules{Float64})::Vector
```

src/analysis/rule_extraction.jl
```diff
@@ -31,19 +31,19 @@ end
...
 
-# Return
-Vector{Vector{Any}}
+# Returns
+Vector{Vector}
 Vector of vectors of rule conditions.
 """
 function _condition(rules::SIRUS.StableRules{Float64}, index::Int64)
-    condition = []
+    condition::Vector{Vector} = []
```
ConnectedSystems, line 46: Delete type info in docstring, move to function signature.

```julia
function _condition(rules::SIRUS.StableRules{Float64}, index::Int64)::Vector{Vector}
```

src/analysis/rule_extraction.jl
```diff
@@ -77,16 +77,97 @@ end
...
+   https://doi.org//10.1214/20-EJS1792
+"""
+function cluster_rules(clusters::Vector{T}, X::DataFrame, outcomes::AbstractMatrix{F},
+    max_rules::T; n_trees::T=1000) where {T<:Int64,F<:Real}
+    # Find robust cluster index
+    rc = robust_cluster(clusters, outcomes)
+    y = clusters .== rc
+
+    # Set seed and Random Number Generator
+    seed = 123
```
ConnectedSystems, line 127: I think best to make this a keyword argument with default `123`.

One approach to convergence analysis is to change the seed value to see how big of a difference in conclusions you get so it would be useful to be able to change the seed

src/analysis/rule_extraction.jl
```diff
@@ -77,16 +77,97 @@ end
...
+    max_outcome = maximum(outcomes)
+    for cluster in unique(clusters)
+        normalized_outcomed = outcomes[:, clusters.==cluster] ./ max_outcome
+        statistic = median(temporal_variability(normalized_outcomed; w=[0.9, 0.1]))
+        push!(clusters_statistics, statistic)
+    end
+
+    # Find index of the cluster with maximum value of summary statistic
+    return argmax(clusters_statistics)
+end
```
ConnectedSystems, line 161: Add kwarg to allow any function to be passed in to be used in place of `temporal_variability`. 

There are many (meta-)metrics that could be adopted here, so it would good to have the flexibility.

src/analysis/rule_extraction.jl
```diff
@@ -77,16 +77,97 @@ end
...
+
+"""
+    robust_cluster(clusters::Vector{T}, result_set::ResultSet) where {T<: Integer}
+
+Find most robust cluster.
+
+# Arguments
+- `clusters` - Vector with outcomes cluster indexes
+- `outcomes` - AbstractMatrix of outcomes factors
```
ConnectedSystems, line 144:
```suggestion
- `clusters` : Vector with outcomes cluster indexes
- `outcomes` : AbstractMatrix of outcomes factors
```

src/analysis/rule_extraction.jl
```diff
@@ -77,16 +77,97 @@ end
...
+    cluster_rules(clusters::Vector{T}, X::DataFrame, outcomes::AbstractMatrix{F}, max_rules::Int64; n_trees::Int64=1000) where {T<:Integer,F<:Real}
+
+Use SIRUS package to extract rules from time series clusters based on some summary metric (default is median)
+
+# Arguments
+- `clusters` - Vector of cluster indexes for each scenario outcome
+- `X::DataFrame` - DataFrame with factors to be used as input by SIRUS
+- `outcomes` - Matrix of some metric over time for all scenarios
+- `max_rules` - Maximum number of rules, to be used as input by SIRUS
+- `n_trees` - Number of trees to be created by SIRUS algorithm
```
ConnectedSystems, line 105: Julia best practice suggests not to repeat information that is provided in the function signature (e.g., the types)

Note: I've opted to replace `factors` with `features` because that's the usual terminology for Random Forest/ML - feel free to change this back to `factors` to be consistent within this documentation.

```suggestion
- `clusters` : Cluster indexes for each scenario outcome
- `X` : Features to be used as input for SIRUS
- `outcomes` : Matrix of some metric over time for all scenarios
- `max_rules` : Maximum number of rules, to be passed to SIRUS
- `n_trees` : Number of trees to be created by SIRUS
```

ConnectedSystems (changes requested):
I may have more comments when I look at this again, but this is what I've picked up today.

src/analysis/rule_extraction.jl
Zapiano: Condition is not a `Vector{Vector{StableRules}}`, it is a Vector of Vectors and each internal Vector has 3 entries: [String, Symbol, Float64]. One possibility is to create the struct:

```julia
struct Clause
feature::String
direction::Symbol  # :L means < and :R means >=
breakpoint::Float64
end
```

And then Condition would be a `Vector{Clause}`. What I like about this approach is that the code I will write accessing this information becomes more readable (e.g., instead of writing `condition[1]` I can write `condition.feature`).

ConnectedSystems: Does `direction` need to be a Symbol? What will you do with `:L` / `:R`?

ConnectedSystems: I assume also that `feature` needs to be a string to allow for easy print out and concatenation with other strings/rule descriptions?

Just checking as I would prefer `Symbol` if the above is not the case.

Zapiano: SIRUS returns a Symbol, and I was just using it. For now I need to use it to

- print either `<` or `>=` symbols 
- when plotting, I do things like: `direction == :L ? [start_point, breakpoint] : [breakpoint, end_point]`
- later, when we actually use this to filter some dataframe, I think we will do something like the above point `direction == :L ? df[:, start:breakpoint] : df[:, breakpoint:end]`

src/analysis/rule_extraction.jl
Zapiano: I am adding more details about the purpose/meaning of the Rules object, but I have one question: this is a Vector of Rules, so my idea is to give more details at the Rules object Docstring and leave this one as it is. Do you think this make sense or do you think there is something specific I should add to this one?

src/analysis/rule_extraction.jl
Zapiano: I agree, that's better.

src/analysis/rule_extraction.jl
ConnectedSystems: You can add a  `# See also` section to the docstring and point the reader to the documentation for `Rules`

src/analysis/rule_extraction.jl
Zapiano: About `feature`, I think it would be just as easy if it was a Symbol. I can change that.

src/analysis/rule_extraction.jl
Zapiano: Why `kwargs` and not a variable called `metric`?

ConnectedSystems: I was assuming you want a default metric...

Zapiano: What do you think changing this function's name to `target_cluster` instead?

Zapiano: About the `kwargs` I was thinking about having a `metric` argument setting `temporal_variability` as the default value. But I think is "a better approach" to set this default value inside the function, in this case?

ConnectedSystems: > What do you think changing this function's name to `target_cluster` instead?

Could be. Another option is to use a term borrowed from statistics: `behavioral`

The cluster with the desirable properties are `behavioral` and those that are not are `non-behavioral`.
So `behavioral_cluster` or `identify_behavioral` ?

ConnectedSystems: > About the kwargs I was thinking about having a metric argument setting temporal_variability as the default value. But I think is "a better approach" to set this default value inside the function, in this case?

Could you elaborate on this? Is this for flexibility in some way?

Zapiano: Maybe if my OO side speaking here, but I prefer `behavioral_cluster` - to me what, in this case what is important is not what this function does, but the information it returns. So I prefer a name the describes what is returns than what is does. But I'm happy either way (:

ConnectedSystems: Haha, it's always a contentious debate. Some people say functions/methods should relate to what they do ("functions are actions"), with documentation on what/why, for example: `Dog.speak()`, `Printer.print()`

I'm happy with `behavioral_cluster` or `target_cluster`, I'll leave it up to you.

Zapiano: > Could you elaborate on this? Is this for flexibility in some way?

Actually I think I prefer you suggestion. But initially I thought of doing something like:

```
function behavioral_cluster(clusters::Vector{T}, outcomes::AbstractMatrix{F}; metric=temporal_variability)
```

And my question was just why one option is better than the other?

ConnectedSystems: > > Could you elaborate on this? Is this for flexibility in some way?
> 
> Actually I think I prefer you suggestion. But initially I thought of doing something like:
> 
> ```
> function behavioral_cluster(clusters::Vector{T}, outcomes::AbstractMatrix{F}; metric=temporal_variability)
> ```

This is what I was suggesting actually. When you said:

`But I think is "a better approach" to set this default value inside the function, in this case?`

By "inside the function" I was imagining you meant something like:

```julia
function behavioral_cluster(...; metric=nothing)
    if isnothing(metric)
        metric = temporal_variability
    end
end
```

But (I guess) you meant inside the function signature rather than inside its definition.

Zapiano: What I meant is what I just pushed:

```
function target_cluster(clusters::Vector{T}, outcomes::AbstractMatrix{F}; kwargs...)
    metric = haskey(kwargs, :metric) ? kwargs[:metric] : temporal_variability

    ...
```

ConnectedSystems: Ah, well in that case I would say if you always expect a `metric` keyword arg, why not just specify it?

The only time I would adopt that pattern is if I expect to need to pass multiple (but separate) kwargs to other functions or if this function is a barrier function that does some preprocessing for some other functions (for example, how `Metrics` are implemented in ADRIA or the visualizations).

Here, `target_cluster` is a single function, with no other kwargs expected.

Zapiano:
> I may have more comments when I look at this again, but this is what I've picked up today.

I think I solved everything here.

ConnectedSystems:
All looks good to me, will merge shortly.

Note @Zapiano that there seems to be some extra commits from a previous PR I did mixed in with this PR now. Please rebase rather than merging if you are incorporating recent changes to a branch.

ConnectedSystems (approved):
LGTM

PR #422, opened 2023-08-16
Implement rule scatter plot function

Zapiano:
This plot may be useful for bump hunting in the future. It is a first  way of visualizing the rules extracted using SIRUS algorithm.

### DIFF
--- a/ext/AvizExt/viz/rule_extraction.jl
+++ b/ext/AvizExt/viz/rule_extraction.jl
@@ -85,7 +85,8 @@ function ADRIA.viz.rules_scatter!(
             condition = rules[index].condition
 
             # Human readable feature names
-            feature_names = _feature_names(first.(condition), rs)
+            fieldnames::Vector{String} = first.(condition)
+            feature_names = _feature_names(fieldnames, rs)
 
             ax::Axis = Axis(
                 sub_g[r, c],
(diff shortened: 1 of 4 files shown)


ext/AvizExt/viz/rule_extraction.jl
```diff
@@ -0,0 +1,114 @@
...
+            index = c + (r - 1) * n_cols
+            if length(rules) < index
+                break
+            end
+            condition = rules[index].condition
+
+            # Feature names
+            condition_features = first.(condition)
+            foi = ms.fieldname .∈ [condition_features]
+            h_names = ms[foi, :name]
```
ConnectedSystems, line 57: I think this extracts the factors/features of interest but does not check for order so it is possible for the figure axis labels to be in the wrong order

Zapiano:
@ConnectedSystems
I've added a legend, did some small refactoring and changed the color to something less aggressive. You can review it again now. 

image

ConnectedSystems:
I still argue the vertical/horizontal guide lines should be given more transparency.
Same with the hypercube/box/rectangle really.

ext/AvizExt/viz/viz.jl
```diff
@@ -43,8 +43,34 @@ function timesteps(outcomes::NamedDimsArray)::Array{Int64}
...
+    if n_factors <= 4
+        n_cols::Int64 = 2
+    else
+        n_cols = max_cols
+    end
+
+    n_rows::Int64 = ceil(Int64, n_factors / n_cols)
+
+    return n_rows, n_cols
+end
```
ConnectedSystems, line 70: This needs to include the changes I raised earlier:

https://github.com/open-AIMS/ADRIA.jl/pull/425/files

Zapiano:
> Looking good Pedro
> 
> I still argue the vertical/horizontal guide lines should be given more transparency. Same with the hypercube/box/rectangle really.

About that:

- I have added a transparency to the guide lines and made the target_area even more transparent.
- I have also made it optional to plot this _highlight_target_area. Below are two examples with this option equals true and false:

rules_all_4096_6_true

rules_all_4096_6_false

ext/AvizExt/viz/viz.jl
Zapiano: @ConnectedSystems Sorry about that, I fixed it now.

ConnectedSystems (approved):
Currently, this approach uses `target_cluster()` to determine which cluster to target. 

In future we will need something that accepts a pre-determined target cluster (or set of scenarios).

PR #424, opened 2023-08-17
Move and modify behaviour of target_cluster function

### LINKED ISSUE #423: Change the way `target_cluster` function is works
Instead of having a function inside `rule_extraction.jl` that finds the index of the `target_cluster`, it would be better to have a function inside `time_series_clustering.jl` that returns a version of the clusters vector but only with 1s and 0s (for target and non target clusters). The advantages of this approach are:

- We would be able to easily plot those two clusters and say something like "these 6 clusters that we had before are now this two clusters"
- `target_cluster` is now being called inside both `cluster_rules` (function that uses SIRUS to extractr rules) and `viz.rules_scatter`. With this approach, we can just pass this vector of clusters to both and be more certain that they both are using the same information
- This way is also easier to compute things like the variance of the target cluster in the future to assess how good this `target_cluster` is

Zapiano:
- The function target_cluster was moved from rule_extraction.jl to time_series_clustering.jl, which makes more sense
- It now returns vector identifying scenarios that belong to target cluster instead of the target cluster index
- It now merges the best cluster with the second one if the number of scenarios inside it is below some threshold

Closes #423 .

### DIFF
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -163,11 +163,11 @@ function target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F};
     target_index = argmax(clusters_statistics)
     target_indexes = [target_index]
 
-    # Merge target cluster if is above 1% of size
+    # Merge target cluster if it is below 1% of size
     sizes = [size(outcomes[:, clusters.==c], 2) for c in unique(clusters)]
     target_size = sizes[target_index] / sum(sizes)
     while target_size < size_limit
-        # Nullify targe_index to find the next argmax
+        # Nullify target_index to find the next argmax
         clusters_statistics[target_index] = 0
 
         # Find next best cluster and add to target_indexes
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -140,17 +140,18 @@ end
 Categorize all clusters into target or non-target according to some metric.
 
 # Arguments
-- `clusters` : Vector with outcomes cluster indexes
-- `outcomes` : AbstractMatrix of outcomes features
+- `clusters` : Vector with outcome cluster indexes
+- `outcomes` : AbstractMatrix of scenario outcomes
 - `metric` : Metric used to aggregate outcomes for each cluster
-- `size_limit` : This function will repeatedely merge the best cluster with the second best
+- `size_limit` : This function will iteratively merge the best cluster with the second best
     if the fraction of scenarios inside it is below `size_limit`
 
 # Returns
 Vector containing 1's for target and 0's for non-target clusters
 """
 function target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F};
-    metric=temporal_variability, size_limit=0.01) where {T<:Int64,F<:Real}
+    metric=temporal_variability, size_limit=0.01)::Vector{T} where {T<:Int64,F<:Real}
+
     # Compute statistic for each cluster
     clusters_statistics::Vector{Float64} = []
     for cluster in unique(clusters)
@@ -159,10 +160,7 @@ function target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F};
         push!(clusters_statistics, statistic)
     end
 
-    # Index of target cluster
     target_index = argmax(clusters_statistics)
-
-    # Index of all target clusters (it may be more than one depending on the below)
     target_indexes = [target_index]
 
     # Merge target cluster if is above 1% of size
@@ -181,5 +179,5 @@ function target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F};
     end
 
     # Return new clusters vector with only 1 and 0 for target and non-target clusters
-    [c ∈ target_indexes ? 1 : 0 for c in clusters]
+    return [c ∈ target_indexes ? 1 : 0 for c in clusters]
 end
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -11,7 +11,7 @@ time steps and \$S\$ is number of scenarios.
 # Arguments
 - `x` : series matrix of shape \$T ⋅ S\$
 
-# Return
+# Returns
 Vector of \$N\$ elements
 
 # Examples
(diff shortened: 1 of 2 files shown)


src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+        # Find next best cluster and add to target_indexes
+        target_index = argmax(clusters_statistics)
+        push!(target_indexes, target_index)
+
+        # Update target_size with next best cluster size
+        target_size += sizes[target_index] / sum(sizes)
+    end
+
+    # Return new clusters vector with only 1 and 0 for target and non-target clusters
+    [c ∈ target_indexes ? 1 : 0 for c in clusters]
```
ConnectedSystems, line 184: I prefer explicit return statements

src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+- `outcomes` : AbstractMatrix of outcomes features
+- `metric` : Metric used to aggregate outcomes for each cluster
+- `size_limit` : This function will repeatedely merge the best cluster with the second best
+    if the fraction of scenarios inside it is below `size_limit`
+
+# Returns
+Vector containing 1's for target and 0's for non-target clusters
+"""
+function target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F};
+    metric=temporal_variability, size_limit=0.01) where {T<:Int64,F<:Real}
```
ConnectedSystems, line 153:
```suggestion
    metric=temporal_variability, size_limit=0.01)::Vector{T} where {T<:Int64,F<:Real}

```

src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+"""
+    target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F}; metric=temporal_variability, size_limit=0.01) where {T<:Int64,F<:Real}
+
+Categorize all clusters into target or non-target according to some metric.
+
+# Arguments
+- `clusters` : Vector with outcomes cluster indexes
+- `outcomes` : AbstractMatrix of outcomes features
+- `metric` : Metric used to aggregate outcomes for each cluster
+- `size_limit` : This function will repeatedely merge the best cluster with the second best
```
ConnectedSystems, line 146:
```suggestion
- `size_limit` : This function will iteratively merge the best cluster with the second best

```

src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
 end
+
+"""
+    target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F}; metric=temporal_variability, size_limit=0.01) where {T<:Int64,F<:Real}
+
+Categorize all clusters into target or non-target according to some metric.
+
+# Arguments
+- `clusters` : Vector with outcomes cluster indexes
+- `outcomes` : AbstractMatrix of outcomes features
```
ConnectedSystems, line 144:
```suggestion
- `outcomes` : AbstractMatrix of scenario outcomes

```

src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
     return [site != 0 ? popfirst!(clusters) : 0 for site in non_null_sites]
 end
+
+"""
+    target_clusters(clusters::Vector{T}, outcomes::AbstractMatrix{F}; metric=temporal_variability, size_limit=0.01) where {T<:Int64,F<:Real}
+
+Categorize all clusters into target or non-target according to some metric.
+
+# Arguments
+- `clusters` : Vector with outcomes cluster indexes
```
ConnectedSystems, line 143:
```suggestion
- `clusters` : Vector with outcome cluster indexes

```

src/analysis/clustering.jl
```diff
@@ -133,3 +133,53 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+    # Compute statistic for each cluster
+    clusters_statistics::Vector{Float64} = []
+    for cluster in unique(clusters)
+        normalized_outcomed = outcomes[:, clusters.==cluster] ./ maximum(outcomes)
+        statistic = median(metric(normalized_outcomed))
+        push!(clusters_statistics, statistic)
+    end
+
+    # Index of target cluster
+    target_index = argmax(clusters_statistics)
```
ConnectedSystems, line 163: I think the code is readable enough without having to add a comment.

ConnectedSystems (changes requested):
Suggesting minor changes for your review

Zapiano:
> Suggesting minor changes for your review

I have implemented all your suggestions.

src/analysis/clustering.jl
```diff
@@ -133,3 +133,51 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+    end
+
+    target_index = argmax(clusters_statistics)
+    target_indexes = [target_index]
+
+    # Merge target cluster if is above 1% of size
+    sizes = [size(outcomes[:, clusters.==c], 2) for c in unique(clusters)]
+    target_size = sizes[target_index] / sum(sizes)
+    while target_size < size_limit
+        # Nullify targe_index to find the next argmax
```
ConnectedSystems, line 170:
```suggestion
        # Nullify target_index to find the next argmax

```

src/analysis/clustering.jl
```diff
@@ -133,3 +133,51 @@ function time_series_clustering(result_set::ResultSet, data::AbstractMatrix{T},
...
+    for cluster in unique(clusters)
+        normalized_outcomed = outcomes[:, clusters.==cluster] ./ maximum(outcomes)
+        statistic = median(metric(normalized_outcomed))
+        push!(clusters_statistics, statistic)
+    end
+
+    target_index = argmax(clusters_statistics)
+    target_indexes = [target_index]
+
+    # Merge target cluster if is above 1% of size
```
ConnectedSystems, line 166:
```suggestion
    # Merge target cluster if it is above 1% of size

```

Is this meant to be "below 1%"?

ConnectedSystems (commented):
Two minor comments. Will merge once addressed.

Zapiano:
> Two minor comments. Will merge once addressed.

@ConnectedSystems  done!

src/analysis/clustering.jl
Zapiano: Yes.

ConnectedSystems (approved):

PR #430, opened 2023-08-24
Fix scenario metrics with field :scenario instead of :scenarios

Zapiano:
Some scenario metrics were returning a Matric object with a field `:scenario` instead of `:scenarios`, which was raising some errors.


ConnectedSystems (approved):
Thanks for picking up on this - surprising this was around for so long!

Issue #324, opened 2023-03-21
Make `run_site_selection()` more user-friendly


ConnectedSystems:
Raising new issue continuing on discussions re [this commit](https://github.com/open-AIMS/ADRIA.jl/commit/65197a3e66e7af4f544411eefe8706523e18b7e1#commitcomment-105303167)

This is also relevant to #252.

Currently `run_site_selection()` requires the user to do a lot of gymnastics to get a usable list of deployment locations.
A small aside: (and as raised previously) the naming should also move away from the "sites" paradigm so it is agnostic to the reef/site/cluster definitions.

The approach I've adopted for this iteration is to fix a few deployment options:

```julia
# TODO: Function that accepts a splatted list of criteria => value pairs to do the below
#       in a single function call. 
# Only use VIKOR
ADRIA.fix_factor!(dom, :guided, 3)

# Always seed (half million each enhanced species)
ADRIA.fix_factor!(dom, :seed_TA, Int64(5e5))
ADRIA.fix_factor!(dom, :seed_CA, Int64(5e5))

# Never fog/shade
ADRIA.fix_factor!(dom, :SRM, 0.0)
ADRIA.fix_factor!(dom, :fogging, 0.0)

# Spread locations around
ADRIA.fix_factor!(dom, :use_dist, 1)
ADRIA.fix_factor!(dom, :dist_thresh, 0.1)
```

Then run the selection process with 128 - 1024 criteria combinations and picking out locations that were ranked the most times for a given rank position.

For example:

Rank 1: The location that was ranked No. 1 the most times
Rank 2: The location that was ranked No. 2 the most times
etc. etc.

I am seeing the same sites being selected regardless of what `dist_thresh` is set to.

![image](https://user-images.githubusercontent.com/4075136/226514384-187c6216-171f-4465-b2e5-46335b0762e9.png)

@Rosejoycrocker wondered what the effect of the `top_n` criteria is. 
As described above, I'm doing a "first past the post" style voting/ranking.

`top_n` ranges from 5 to 50, and Sobol' sampling guarantees an equi-distributed sample.
Is it that low values of `top_n` are biasing results? i.e., different locations become highly ranked with increasing `top_n` such that there is greater consensus with lower values of `top_n`?

Ideally the process below should be hidden away from the user.

```julia
# Target priority sites (this functionality is not yet in `Main`)
# Currently using site indexes, but should use their named IDs instead.
[...]

PR #431, opened 2023-08-24
Adds a series of aggregation functions for location selection ranks

Rosejoycrocker:
Addresses issue #324
- adds 3 aggregation functions for ranks outputs
- ranks can be taken in as result set for ADRIA runs outputs or as ranks NamedDimsArray from run_site_selection()
- 3 functions for calculating location selection frequencies 
- location_selection_frequencies() counts frequency with which sites are actually selected (ranks 1-5)
- ranks_to_frequencies() counts frequencies with which each site is selected for each rank.
- ranks_to_frequencies_ts() calculates the same but for each time step.
- ranks_to_location_order() gives the ranking order as location names for each scenario (no time dependent version).
- Other changes fix bugs in run_site_selection() due to upstream changes.

### DIFF
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -146,9 +146,8 @@ using ADRIA:
 using DataFrames
 using Statistics, StatsBase
 
-@info "Loading data package"
-here = @__DIR__
-dom = ADRIA.load_domain(joinpath(here, "Moore_2023-08-17"), "45")
+# Load data package
+dom = ADRIA.load_domain("path to Domain files", "RCP")
 scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.
 
 area_to_seed = 962.11  # Area of seeded corals in m^2.
@@ -156,14 +155,13 @@ area_to_seed = 962.11  # Area of seeded corals in m^2.
 # Initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites)).
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
-@info "Run site selection"
 # Use run_site_selection to get ranks
 ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
 
 # Get frequencies with which each site is selected for each rank for set of stand alone location selections.
 rank_freq = ranks_to_frequencies(ranks[intervention=1])
 
-@info "Calculate rank aggregations"
+# Calculate rank aggregations
 # Get location selection freqencies for set of standalone location selections.
 location_selection_frequency = location_selection_frequencies(ranks[intervention=1])
 # Get summed inverse rank for set of standalone location selections.
@@ -202,6 +200,5 @@ unguided_freq = location_selection_frequencies(
 inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1])
 # Get summed inverse rank over time.
 inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1]; dims=[:scenarios])
-
 
 ```
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -117,7 +117,7 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
 
 ```julia
 using ADRIA
-using ADRIA: rank_locations
+using ADRIA: run_site_selection
 
 @info "Loading data package"
 here = @__DIR__
@@ -130,8 +130,8 @@ area_to_seed = 962.11  # Area of seeded corals in m^2.
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
 @info "Run site selection"
-# Use rank_locations to get ranks
-ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
+# Use run_site_selection to get ranks
+ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
 ```
 
 ## Intervention location selection - summary functions
@@ -139,57 +139,49 @@ ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
 ```julia
 using ADRIA
 using ADRIA:
-    rank_locations,
+    run_site_selection,
     ranks_to_frequencies,
+    ranks_to_location_order,
     location_selection_frequencies,
     summed_inverse_rank
 using DataFrames
 using Statistics, StatsBase
 
-# Load data package
-dom = ADRIA.load_domain("path to Domain files", "RCP")
-scens = ADRIA.sample_site_selection(dom, 8) # Get site selection scenario dataframe.
+@info "Loading data package"
+here = @__DIR__
+dom = ADRIA.load_domain(joinpath(here, "Moore_2023-08-17"), "45")
+scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.
 
 area_to_seed = 962.11  # Area of seeded corals in m^2.
 
 # Initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites)).
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
-# Use rank_locations to get ranks
-ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
-
+@info "Run site selection"
+# Use run_site_selection to get ranks
+ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
 # Get frequencies with which each site is selected for each rank for set of stand alone location selections.
 rank_freq = ranks_to_frequencies(ranks[intervention=1])
 
-# Calculate rank aggregations
+@info "Calculate rank aggregations"
 # Get location selection freqencies for set of standalone location selections.
 location_selection_frequency = location_selection_frequencies(ranks[intervention=1])
 # Get summed inverse rank for set of standalone location selections.
 # Measure of magnitude and frequency of high rank.
 inv_summ_rank = summed_inverse_rank(ranks[intervention=1])
 
-# Use aggregation function within rank_locations to get direct output.
+# Use aggregation function within run_site_selection to get direct output.
 # To get rank frequencies:
-rank_frequencies_seed = rank_locations(
+rank_frequencies_seed = run_site_selection(
     dom, scens, sum_cover, area_to_seed, ranks_to_frequencies, 1
-)
-rank_frequencies_seed = rank_locations(
-    dom, scens, sum_cover, area_to_seed, location_selection_frequencies, 1
-)
-rank_frequencies_seed = rank_locations(
-    dom, scens, sum_cover, area_to_seed, summed_inverse_rank, 1
 )
 
 # Example using ADRIA runs
-scens = ADRIA.sample(dom, 2^5) # Get scenario dataframe.
+scens = ADRIA.sample(dom, 8) # Get scenario dataframe.
 rs = ADRIA.run_scenarios(dom, scens, "45") # Run scenarios.
 
 # Get frequencies with which each site is selected for each rank for set of runs.
-rank_freq = ranks_to_frequencies(rs.ranks[intervention=1]) # with timesteps not aggregated
-rank_freq = ranks_to_frequencies(
-    rs.ranks[intervention=1];
-    agg_func=x -> dropdims(sum(x; dims=:timesteps); dims=:timesteps),
-) # with timesteps aggregated
+rank_freq = ranks_to_frequencies(rs.ranks[intervention=1])
 
 # Get selection frequencies for set of runs.
 selection_freq = location_selection_frequencies(rs.ranks[intervention=1])
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -141,7 +141,6 @@ using ADRIA
 using ADRIA:
     run_site_selection,
     ranks_to_frequencies,
-    ranks_to_location_order,
     location_selection_frequencies,
     summed_inverse_rank
 using DataFrames
@@ -160,6 +159,7 @@ sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 @info "Run site selection"
 # Use run_site_selection to get ranks
 ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
+
 # Get frequencies with which each site is selected for each rank for set of stand alone location selections.
 rank_freq = ranks_to_frequencies(ranks[intervention=1])
 
@@ -175,9 +175,15 @@ inv_summ_rank = summed_inverse_rank(ranks[intervention=1])
 rank_frequencies_seed = run_site_selection(
     dom, scens, sum_cover, area_to_seed, ranks_to_frequencies, 1
 )
+rank_frequencies_seed = run_site_selection(
+    dom, scens, sum_cover, area_to_seed, location_selection_frequencies, 1
+)
+rank_frequencies_seed = run_site_selection(
+    dom, scens, sum_cover, area_to_seed, summed_inverse_rank, 1
+)
 
 # Example using ADRIA runs
-scens = ADRIA.sample(dom, 8) # Get scenario dataframe.
+scens = ADRIA.sample(dom, 2^5) # Get scenario dataframe.
 rs = ADRIA.run_scenarios(dom, scens, "45") # Run scenarios.
 
 # Get frequencies with which each site is selected for each rank for set of runs.
@@ -196,5 +202,6 @@ unguided_freq = location_selection_frequencies(
 inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1])
 # Get summed inverse rank over time.
 inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1]; dims=[:scenarios])
+
 
 ```
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -119,17 +119,16 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
 using ADRIA
 using ADRIA: rank_locations
 
-@info "Loading data package"
-here = @__DIR__
-dom = ADRIA.load_domain(joinpath(here, "Moore_2023-08-17"), "45")
-scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.
 
-area_to_seed = 962.11  # Area of seeded corals in m^2.
+dom = ADRIA.load_domain("path to domain", "45")
+scens = ADRIA.sample_site_selection(dom, 8)
+
+# Area of seeded corals in m^2
+area_to_seed = 962.11
 
 # Initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites)).
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
-@info "Run site selection"
 # Use rank_locations to get ranks
 ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
 ```
@@ -142,33 +141,38 @@ using ADRIA:
     rank_locations,
     ranks_to_frequencies,
     location_selection_frequencies,
-    summed_inverse_rank
+    selection_score
 using DataFrames
 using Statistics, StatsBase
 
 # Load data package
 dom = ADRIA.load_domain("path to Domain files", "RCP")
-scens = ADRIA.sample_site_selection(dom, 8) # Get site selection scenario dataframe.
 
-area_to_seed = 962.11  # Area of seeded corals in m^2.
+# Select locations for interventions without any model runs
+scens = ADRIA.sample_site_selection(dom, 8)  # Get site selection scenario dataframe.
 
-# Initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites)).
+# Area of seeded corals in m^2
+area_to_seed = 962.11
+
+# Initial coral cover matching number of criteria samples
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
 # Use rank_locations to get ranks
 ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
 
-# Get frequencies with which each site is selected for each rank for set of stand alone location selections.
+# Get frequencies with which each site is selected for each rank for set of stand alone
+# location selections
 rank_freq = ranks_to_frequencies(ranks[intervention=1])
 
 # Calculate rank aggregations
-# Get location selection freqencies for set of standalone location selections.
+# Get location selection freqencies for set of standalone location selections
 location_selection_frequency = location_selection_frequencies(ranks[intervention=1])
-# Get summed inverse rank for set of standalone location selections.
-# Measure of magnitude and frequency of high rank.
+
+# Get summed inverse rank for set of standalone location selections
+# Measure of magnitude and frequency of high rank
 sel_score = selection_score(ranks[intervention=1])
 
-# Use aggregation function within rank_locations to get direct output.
+# Use aggregation function within rank_locations to get direct output
 # To get rank frequencies:
 rank_frequencies_seed = rank_locations(
     dom, scens, sum_cover, area_to_seed, ranks_to_frequencies, 1
@@ -177,32 +181,34 @@ rank_frequencies_seed = rank_locations(
     dom, scens, sum_cover, area_to_seed, location_selection_frequencies, 1
 )
 rank_frequencies_seed = rank_locations(
-    dom, scens, sum_cover, area_to_seed, summed_inverse_rank, 1
+    dom, scens, sum_cover, area_to_seed, selection_score, 1
 )
 
 # Example using ADRIA runs
-scens = ADRIA.sample(dom, 2^5) # Get scenario dataframe.
-rs = ADRIA.run_scenarios(dom, scens, "45") # Run scenarios.
+scens = ADRIA.sample(dom, 8)
+rs = ADRIA.run_scenarios(dom, scens, "45")
 
-# Get frequencies with which each site is selected for each rank for set of runs.
-rank_freq = ranks_to_frequencies(rs.ranks[intervention=1]) # with timesteps not aggregated
+# Get frequencies with which each site was selected for each rank for a set of runs
+rank_freq = ranks_to_frequencies(rs.ranks[intervention=1])  # with timesteps not aggregated
+
+# With timesteps aggregated
 rank_freq = ranks_to_frequencies(
     rs.ranks[intervention=1];
     agg_func=x -> dropdims(sum(x; dims=:timesteps); dims=:timesteps),
-) # with timesteps aggregated
-
-# Get selection frequencies for set of runs.
-selection_freq = location_selection_frequencies(rs.ranks[intervention=1])
-
-# Get selection frequencies over time for unguided runs only.
-unguided_freq = location_selection_frequencies(
-    rs.seed_log[scenarios=findall(scens.guided .== 1)]
 )
 
-# Get summed inverse rank for set of runs.
-# Measure of magnitude and frequency of high rank.
-sel_score = selection_score(rs.ranks[intervention=1])
-# Get summed inverse rank over time.
-sel_score = selection_score(rs.ranks[intervention=1]; dims=[:scenarios])
+# Get selection frequencies for set of runs
+selection_freq = location_selection_frequencies(rs.ranks[intervention=1])
 
+# Get selection frequencies over time for unguided runs only
+unguided_freq = location_selection_frequencies(
+    rs.seed_log[scenarios=findall(scens.guided .>= 1)]
+)
+
+# Get selection score for set of runs
+# Measure of magnitude and frequency of high rank
+sel_score = selection_score(rs.ranks[intervention=1])
+
+# Get selection score for locations over time
+sel_score = selection_score(rs.ranks[intervention=1]; dims=[:scenarios])
 ```
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -166,7 +166,7 @@ rank_freq = ranks_to_frequencies(ranks[intervention=1])
 location_selection_frequency = location_selection_frequencies(ranks[intervention=1])
 # Get summed inverse rank for set of standalone location selections.
 # Measure of magnitude and frequency of high rank.
-inv_summ_rank = selection_score(ranks[intervention=1])
+sel_score = selection_score(ranks[intervention=1])
 
 # Use aggregation function within rank_locations to get direct output.
 # To get rank frequencies:
@@ -201,8 +201,8 @@ unguided_freq = location_selection_frequencies(
 
 # Get summed inverse rank for set of runs.
 # Measure of magnitude and frequency of high rank.
-inv_summ_rank = selection_score(rs.ranks[intervention=1])
+sel_score = selection_score(rs.ranks[intervention=1])
 # Get summed inverse rank over time.
-inv_summ_rank = selection_score(rs.ranks[intervention=1]; dims=[:scenarios])
+sel_score = selection_score(rs.ranks[intervention=1]; dims=[:scenarios])
 
 ```
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -117,7 +117,7 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
 
 ```julia
 using ADRIA
-using ADRIA: run_site_selection
+using ADRIA: rank_locations
 
 @info "Loading data package"
 here = @__DIR__
@@ -130,8 +130,8 @@ area_to_seed = 962.11  # Area of seeded corals in m^2.
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
 @info "Run site selection"
-# Use run_site_selection to get ranks
-ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
+# Use rank_locations to get ranks
+ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
 ```
 
 ## Intervention location selection - summary functions
@@ -139,7 +139,7 @@ ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
 ```julia
 using ADRIA
 using ADRIA:
-    run_site_selection,
+    rank_locations,
     ranks_to_frequencies,
     location_selection_frequencies,
     summed_inverse_rank
@@ -155,8 +155,8 @@ area_to_seed = 962.11  # Area of seeded corals in m^2.
 # Initial coral cover matching number of criteria samples (size = (no. criteria scens, no. of sites)).
 sum_cover = repeat(sum(dom.init_coral_cover; dims=1), size(scens, 1))
 
-# Use run_site_selection to get ranks
-ranks = run_site_selection(dom, scens, sum_cover, area_to_seed)
+# Use rank_locations to get ranks
+ranks = rank_locations(dom, scens, sum_cover, area_to_seed)
 
 # Get frequencies with which each site is selected for each rank for set of stand alone location selections.
 rank_freq = ranks_to_frequencies(ranks[intervention=1])
@@ -168,15 +168,15 @@ location_selection_frequency = location_selection_frequencies(ranks[intervention
 # Measure of magnitude and frequency of high rank.
 inv_summ_rank = summed_inverse_rank(ranks[intervention=1])
 
-# Use aggregation function within run_site_selection to get direct output.
+# Use aggregation function within rank_locations to get direct output.
 # To get rank frequencies:
-rank_frequencies_seed = run_site_selection(
+rank_frequencies_seed = rank_locations(
     dom, scens, sum_cover, area_to_seed, ranks_to_frequencies, 1
 )
-rank_frequencies_seed = run_site_selection(
+rank_frequencies_seed = rank_locations(
     dom, scens, sum_cover, area_to_seed, location_selection_frequencies, 1
 )
-rank_frequencies_seed = run_site_selection(
+rank_frequencies_seed = rank_locations(
     dom, scens, sum_cover, area_to_seed, summed_inverse_rank, 1
 )
 
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -148,7 +148,7 @@ using Statistics, StatsBase
 
 # Load data package
 dom = ADRIA.load_domain("path to Domain files", "RCP")
-scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.
+scens = ADRIA.sample_site_selection(dom, 8) # Get site selection scenario dataframe.
 
 area_to_seed = 962.11  # Area of seeded corals in m^2.
 
--- a/docs/src/usage/cookbook.md
+++ b/docs/src/usage/cookbook.md
@@ -166,7 +166,7 @@ rank_freq = ranks_to_frequencies(ranks[intervention=1])
 location_selection_frequency = location_selection_frequencies(ranks[intervention=1])
 # Get summed inverse rank for set of standalone location selections.
 # Measure of magnitude and frequency of high rank.
-inv_summ_rank = summed_inverse_rank(ranks[intervention=1])
+inv_summ_rank = selection_score(ranks[intervention=1])
 
 # Use aggregation function within rank_locations to get direct output.
 # To get rank frequencies:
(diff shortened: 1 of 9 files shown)


ConnectedSystems:
@Zapiano could you review please?

ConnectedSystems:
@Zapiano reminding you of this PR in case it dropped off your todo list.

Rosejoycrocker:
Hey @ConnectedSystems, Pedro and I spoke about it yesterday and we thought that since it's not urgent he could wait until I get back and we can talk about how location selection works in ADRIA. Let us know what you think.

Zapiano:
> @Zapiano reminding you of this PR in case it dropped off your todo list.

Yes, I started review tomorrow but haven't finished. It's on my todo list.

ConnectedSystems:
> Yes, I started reviewing tomorrow but haven't finished.

How was the time travelling?

Zapiano:
> > Yes, I started reviewing tomorrow but haven't finished.
> 
> How was the time travelling?

I mean, yesterday... 🤦🏻‍♂️

src/decision/location_selection.jl
```diff
@@ -0,0 +1,265 @@
...
+    end
+
+    if isnothing(target_seed_sites) && isnothing(target_shade_sites)
+        target_site_ids = collect(1:length(domain.site_ids))
+    end
+
+    n_sites = length(domain.site_ids)
+    for (scen_idx, scen) in enumerate(eachrow(scenarios))
+
+        depth_criteria = (domain.site_data.depth_med .<= (scen.depth_min .+ scen.depth_offset)) .& (domain.site_data.depth_med .>= scen.depth_min)
```
Zapiano, line 97: You could assign `domain.site_data.depth_med .<= (scen.depth_min .+ scen.depth_offset)` and `domain.site_data.depth_med .>= scen.depth_min` to separate variables. The number of allocations would still be the same as well as the memory usage. I did the following test here:

```julia
function func1(x, y, z)
    iguais = x .== y
    return z[iguais]
end

function func2(x, y, z)
    return z[x .== y]
end

x, y, z = bitrand(100), bitrand(100), bitrand(100)

@btime func1(x,y,z)
#   135.755 ns (4 allocations: 208 bytes)
@btime func2(x,y,z)
#  136.114 ns (4 allocations: 208 bytes)
```

Zapiano (changes requested):
@Rosejoycrocker I did some comments but, because this PR was done before we stablished the style guide, there are a lot more that need to be changed in to follow it - in particular the 92 characters limit per line. I added this as suggestion in a few places but I think it would be better if you just check the entire file again.

Zapiano:
@Rosejoycrocker Can you give me the context where this new functions will be used? Because just by looking at the examples I understand how to use them, but not for what.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,265 @@
...
+        depth_priority = findall(depth_criteria)
+
+        considered_sites = target_site_ids[findall(in(depth_priority), target_site_ids)]
+
+        ranks_store(scenarios=scen_idx, sites=domain.site_ids[considered_sites]) .= site_selection(
+            domain,
+            scen,
+            vec((mean(wave_scens[:, :, target_wave_scens], dims=(:timesteps, :scenarios)) .+ std(wave_scens[:, :, target_wave_scens], dims=(:timesteps, :scenarios))) .* 0.5),
```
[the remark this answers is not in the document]
Rosejoycrocker: This is addressed in #434

src/decision/location_selection.jl
Rosejoycrocker: This will all be done in an external function as in #434

src/decision/location_selection.jl
```diff
@@ -153,7 +153,8 @@ function ranks_to_location_order(ranks::NamedDimsArray)
     end
     return location_orders
 end
-function ranks_to_location_order(ranks::NamedDimsArray, iv_type::String)
+function ranks_to_location_order(ranks::NamedDimsArray, 
+    iv_type::String)
```
ConnectedSystems, line 159: Hi Rose, is the JuliaFormatter doing this?

Rosejoycrocker: No it's not, I did that trying to get it to work, is there anything else I have to do other than have the JuliaFormatter extension and the Julia formatter file?

ConnectedSystems: @Zapiano any advice here?

Rosejoycrocker:
Hey @ConnectedSystems, currently ranks as output from run_site_selection() has interventions labeled ["seed","shade"], while in the result set they are labelled [1,2]. I think it would be good to have consistent labels and I'm lesning towards labelling with strings, but I'm not sure how to change the ResultSet labelling here: https://github.com/open-AIMS/ADRIA.jl/blob/b11152dea75f88e70c6aa63ff64cfafad2fcb161/src/io/ResultSet.jl#L66

Would I have to add labels for each dimension?

src/decision/location_selection.jl
Zapiano: @Rosejoycrocker If you press Ctrl+Shift+P and write "Settings" and in the Settings search bar write "format on save", can you confirm that the "Format on save" option is selected?

Rosejoycrocker: Hi @Zapiano, just checked and format on save is on

Zapiano: And you have a file named `.JuliaFormatter.toml` (starting with a dot) and the content of this file is

```
style = "blue"
```

?

Zapiano: If all the above is true, one more thing: you said you have another formatter extension installed, right? One called "black formatter" or something? Have you tried disabling it?

Rosejoycrocker: I have the .JuliaFormatter.toml file with style="blue". I have it in the ADRIA.jl folder, is this the correct place?

I don't have any other Julia formatters installed. Black formatter I use to format my Python code.

Zapiano: Yes, I have my `.JuliaFormatter.toml` file inside Adria.jl folder...

Are you sure Black formatter is not interfering with Julia formatter (even thought it is only for python)? 

The one you have is this, right?

image

v.1.2.1? Enabled globally? I'm running out of ideas...

Rosejoycrocker: I workled it out, I had format on save on but I seeded to change the mode to "modifications" :'D
Thanks @Zapiano !

Zapiano: Awesome!

ConnectedSystems:
> I'm leaning towards labelling with strings but I'm not sure how to change the ResultSet labelling here

I'm trying to standardize to symbols where possible. 

As for the current labelling, this was implemented before we adopted AxisKeys so it only specifies the dimension labels.

We'll hopefully be moving to something else to make this whole thing easier soon

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+)
+    selected_ranks = _get_iv_type(rs.ranks, iv_type)
+    return summed_inverse_rank(selected_ranks; dims=dims,agg_func=agg_func)
+
+end
+function summed_inverse_rank(
+    ranks::NamedDimsArray; dims::Symbol=:scenarios, agg_func::Function=x -> x
+)
+    n_ranks = maximum(ranks)
+    inv_ranks = agg_func(dropdims(sum(n_ranks .- ranks,dims=dims),dims=dims))
```
ConnectedSystems, line 319: Fix up spacing between arguments please.

(I'm surprised the formatter isn't fixing this for you)

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+    return loc_count
+end
+
+"""
+    summed_inverse_rank(ranks::NamedDimsArray, iv_type::String; dims=:scenarios,agg_func=x->x)
+    summed_inverse_rank(rs::ResultSet, iv_type::String; dims=:scenarios,agg_func=x->x)
+    summed_inverse_rank(ranks::NamedDimsArray; dims=:scenarios,agg_func=x->x)
+ 
+Calculates (number of sites) .- ranks summed over the dimension dims and transformed using agg_func 
+    (default no transformation).
```
ConnectedSystems, line 285: I've read this a few times now and I feel there's a simpler flow here.

You could have a `summed_inverse_rank` that does not have the `agg_func` argument, so it passes on `x -> x`.

I get the sense that a few responsibilities are being mixed together here. The flow might work better if:

1. You get the ranks of interest first
2. Pass it into this function for collation

```julia
iv_ranks = selection_ranks(rs, iv_type)  # this function is just renamed `_get_iv_type`

# AND/OR

iv_ranks = selection_ranks(ranks, iv_type)  # this function is just renamed `_get_iv_type`

ranks_of_interest = summed_inverse_rank(iv_ranks)
```

If you were to create a pipeline of processing steps:

```julia
ranks_of_interest = selection_ranks(rs, iv_type) |> 
    summed_inverse_ranks
```

This way you only have to support one use case.

I suppose the larger issue is I'm not sure how this is used or when it's useful - the function name and/or descriptor could be more informative - yes it's the summed inverse rank but for what purpose?
What alternate transformations/aggregations would be preferable?

And if you're supporting alternate aggregations, could the inversion not be one of them?

```julia
selection_ranks(rs, iv_type; agg_func=selection_inversion_function)
```

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+# Arguments
+- `ranks` : Contains location ranks for each scenario of location selection, as created by 
+    `run_location_selection()`.
+
+# Returns
+- Ids for full scenario set in ranks.
+"""
+function _get_scen_ids(ranks::NamedDimsArray)
+    return collect(ranks.scenarios)
+end
```
ConnectedSystems, line 356: The function is so specific, is it really worthwhile having it?

Rather than the pattern:

```julia
function some_function(array; ids=_get_scen_ids(...))
end
```

the pattern adopted elsewhere in ADRIA (see scenario metrics and visualizations) is:

```julia
function some_function(array)
    # use scenario ids by default
end
function some_function(array, filter_func)
    # apply provided filter function to extract relevant scenarios
end
```

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+
+    # site_id, seeding rank, shading rank
+    rankingsin = [mcda_vars.site_ids zeros(Int64, n_sites) zeros(Int64, n_sites)]
+
+    prefseedsites::Vector{Int64} = zeros(Int64, mcda_vars.n_site_int)
+    prefshadesites::Vector{Int64} = zeros(Int64, mcda_vars.n_site_int)
+
+    # Determine connectivity strength
+    # Account for cases where no coral cover
+    in_conn, out_conn, strong_pred = connectivity_strength(domain.TP_data .* site_k_area(domain), Array(mcda_vars.sum_cover))
```
ConnectedSystems, line 31: I see you changed the expected type in `connectivity_strength()` for this.

I'm just wondering why an array is expected here. Isn't vector more appropriate?

ConnectedSystems (changes requested):

ConnectedSystems:
By the way, there's a conflict in `dMCDA.jl` - seems `distance_sorting()` has some new arguments that aren't used?

src/decision/location_selection.jl
Rosejoycrocker: This was a metric Pedro suggested because it meant the higher ranked the larger the number of summed inverse ranked. But you're right it could probably be an option in selection_ranks. Do you think it is useful? I think it could be a could measure but may be hard to explain to a wider audience.

src/decision/location_selection.jl
Rosejoycrocker: Good point, I can change this

src/decision/location_selection.jl
Rosejoycrocker: I can't remember exactly why I did this but I believe it was necessary to fit both the ecological model and standalone site selection cases... but I'll get back to you on this

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+    `run_location_selection()`.
+- `rs` : ADRIA result set. 
+- `iv_type` : String indicating the intervention type to perform aggregation on.
+- `n_ranks` : Consider first n_ranks, default is all ranks (n_locs).
+
+# Returns 
+- Frequency with which each location was selected for each rank over time.
+"""
+function ranks_to_frequencies_ts(ranks::NamedDimsArray; n_ranks::Int64=length(ranks.sites))
+    rank_frequencies = NamedDimsArray(zeros(length(ranks.timesteps), length(ranks.sites), length(ranks.sites)), timesteps=ranks.timesteps, sites=ranks.sites, ranks=1:length(ranks.sites))
```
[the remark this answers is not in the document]
Rosejoycrocker: I'll fix all the formatting, when I created this PR we hadn't decided to migrate to Blue style yet

Rosejoycrocker:
> By the way, there's a conflict in `dMCDA.jl` - seems `distance_sorting()` has some new arguments that aren't used?

This was due to removing `top_n`, I just rebased so should be fine now

src/decision/location_selection.jl
ConnectedSystems: Oh I see, so it's not the ranks themselves, it's the rankings and the frequency they're at those ranks?

Rosejoycrocker: Sort of but not entriely. For each rank it is (number of sites - rank), so the highest rank will have the highest values. Then it is summed over scenarios, so sites selected most frequently with the highest rank will have higher values than those selected less frequently at higher rank, if that makes sense.

ConnectedSystems: Alright - could you weight by the maximum rank to ease assessment? Values closer to 1.0 would then be locations that are most selected at higher ranks.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,356 @@
...
+    ranks_frequencies = ranks_to_frequencies(ranks[scenarios=ind_metrics], iv_type; n_ranks=n_loc_int)
+    loc_count = sum(ranks_frequencies[ranks=1:n_loc_int], dims=2)[ranks=1]
+
+    return loc_count
+end
+function location_selection_frequencies(
+    rs::ResultSet,
+    iv_type::String;
+    n_loc_int::Int64=5,
+    ind_metrics::Vector{Int64}=_get_scen_ids(rs.ranks),
```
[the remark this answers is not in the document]
Rosejoycrocker: n_location_int is the number of locations to intervene at

docs/src/usage/cookbook.md
```diff
@@ -118,18 +118,83 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
 ```julia
 using ADRIA
 using ADRIA: run_site_selection
-using DataFrames
 
+@info "Loading data package"
+here = @__DIR__
+dom = ADRIA.load_domain(joinpath(here, "Moore_2023-08-17"), "45")
+scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.
```
ConnectedSystems, line 126: Replace `@info` with comments

Make the example more general, assume users do not have the specific data package you are using.

`scens = ADRIA.sample_site_selection(dom, 8) # Get scenario dataframe.`

As mentioned before, **two** spaces before an inline comment please (we're adopting PEP8-style recommendations here).

But is this comment really necessary?
Is the returned dataframe really scenarios?

docs/src/usage/cookbook.md
```diff
@@ -118,18 +118,83 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
...
+unguided_freq = location_selection_frequencies(
+    rs.seed_log[scenarios=findall(scens.guided .== 1)]
+)
+
+# Get summed inverse rank for set of runs.
+# Measure of magnitude and frequency of high rank.
+inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1])
+# Get summed inverse rank over time.
+inv_summ_rank = summed_inverse_rank(rs.ranks[intervention=1]; dims=[:scenarios])
+
```
ConnectedSystems, line 199: Remove extra unnecessary space

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+
+    prefseedsites::Vector{Int64} = zeros(Int64, mcda_vars.n_site_int)
+    prefshadesites::Vector{Int64} = zeros(Int64, mcda_vars.n_site_int)
+
+    # Determine connectivity strength
+    # Account for cases where no coral cover
+    in_conn, out_conn, strong_pred = connectivity_strength(domain.TP_data .* site_k_area(domain), Array(mcda_vars.sum_cover))
+
+    # Perform location selection for seeding and shading.
+    seed_true, shade_true = [true, true]
```
ConnectedSystems, line 36:
```suggestion
    seed_true, shade_true = true, true
```

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+    ranks::NamedDimsArray{D,T,2,A};
+    n_ranks=length(ranks.sites),
+    agg_func=nothing,
+) where {D,T,A}
+    if !isnothing(agg_func)
+        return agg_func(ranks_to_frequencies(ranks, n_ranks))
+    end
+
+    return ranks_to_frequencies(ranks, n_ranks)
+end
```
ConnectedSystems, line 192: We don't need the first one with a default `agg_func`, do we?

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+
+"""
+    location_selection_frequencies(ranks::NamedDimsArray;n_loc_int::Int64=5)
+    location_selection_frequencies(inv_log::NamedDimsArray{D,T,4,A};dims::Union{Symbol,Vector{Symbol}}=:coral_id) where {D,T,A}
+
+Post-processing function for intervention logs. Calculates the frequencies with which locations were selected for a particular intervention,
+for a selection of scenarios (e.g. selected robust scenarios).
```
ConnectedSystems, line 199: Good docs state clearly what the function does.
There's no need to say "such and such function", "This function", etc.

We know it's a function - it's why it has documentation.

```
Post-process intervention logs to calculate the frequency with which locations were selected for a particular intervention, for a set of scenarios.
```

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+
+# Arguments
+- `ranks` : Contains location ranks for each scenario of location selection, as created by
```
ConnectedSystems, line 202:
```
Contains location ranks for each scenario of location selection
```

This is a bit cryptic. How about:

```
Rankings of locations `run_location_selection()`
```

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+            domain, mcda_vars_temp, scen.guided
+        )
+    end
+    # Set filtered locations as n_locs+1 for consistency with time dependent ranks
+    ranks_store[ranks_store.==0.0] .= length(domain.site_ids)+1
+
+    return ranks_store
+
+end
+function run_site_selection(domain::Domain,
```
ConnectedSystems, line 118: This returns location rankings.

Could this just be `rank_locations()` ?

Incidentally, here the function is `run_site_selection` but elsewhere it is referred to as `run_location_selection`.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+
+# Arguments
+- `ranks` : Contains location ranks for each scenario of location selection, as created by 
+    `run_location_selection()`.
+- `dims` : Dimensions to sum over.
+
+# Returns 
+Inverse rankings (i.e. the greater the number the higher ranked the site).
+"""
+function summed_inverse_rank(
```
ConnectedSystems, line 247: @Zapiano are you happy with this function name?

ConnectedSystems (changes requested):
Some changes requested for clarity

src/decision/location_selection.jl
Rosejoycrocker: If I left out the default aggfunc the default output would include timesteps. Happy to do this, probablky more consistent, but I'd have to add the check for the aggfunc as in the second case

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+    ranks::NamedDimsArray{D,T,2,A};
+    dims::Vector{Symbol}=[:scenarios],
+) where {D,T,A}
+    return summed_inverse_rank(ranks, dims)
+end
+function summed_inverse_rank(
+    ranks::NamedDimsArray,
+    dims::Vector{Symbol},
+)
+    n_ranks = maximum(ranks)
```
Zapiano, line 260: Why is `maximum(ranks)` the number of ranks? Isn't that just the "maximum rank"?

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+    inv_ranks = dropdims(sum(n_ranks .- ranks; dims=dims); dims=dims[1])
```
Zapiano, line 265: Since this function receives a NamedDimsArray as input and doesn't give support for simples Arrays, why not pass the dimension you want to drop as a key (I assume is `:timestamp`, right?) instead of a number?

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+Calculates (number of sites) .- ranks summed over the dimension dims and transformed using agg_func 
+    (default no transformation).
+
+# Arguments
+- `ranks` : Contains location ranks for each scenario of location selection, as created by 
+    `run_location_selection()`.
+- `dims` : Dimensions to sum over.
+
+# Returns 
+Inverse rankings (i.e. the greater the number the higher ranked the site).
```
Zapiano, line 246: Isn't "the greater the number, the higher ranked the site" the definition of a "normal"/common ranking?

src/decision/location_selection.jl
Rosejoycrocker: The number of ranks is  also the maximum value of ranks, but happy to change this to make it clear

src/decision/location_selection.jl
Rosejoycrocker: The dimension here is a key, it's just when using dropdims you don't want to drop all the dimensions that are summed over, just the first one

src/decision/location_selection.jl
Rosejoycrocker: In our case the highest rank is 1, second highest is 2 etc, so when we take the inverse the greater the number the greater the rank

src/decision/location_selection.jl
Zapiano: Maybe doing `n_ranks = size(ranks, dim(rank, :location)) + 1` would make this more explicit?

Zapiano: OR maybe what you want IS the maximum rank, and not the number of ranks...? (Reading the rest of this code, that makes more sense to me..., what do you think?)

src/decision/location_selection.jl
```diff
@@ -0,0 +1,266 @@
...
+) where {D,T,A}
+    return summed_inverse_rank(ranks, dims)
+end
+function summed_inverse_rank(
+    ranks::NamedDimsArray,
+    dims::Vector{Symbol},
+)
+    n_ranks = maximum(ranks)
+    inv_ranks = dropdims(sum(n_ranks .- ranks; dims=dims); dims=dims[1])
+    return inv_ranks ./ (n_ranks * prod([size(ranks, d) for d in dims]))
```
Zapiano: You can replace `prod([size(ranks, d) for d in dims])` with `prod(dim(ranks, dims))`

src/decision/location_selection.jl
Rosejoycrocker: I use the maximum because in the case where sites are filtered out of the ranking process the maximum rank is the number of sites +1, whereas when no filtering occurs the maximum rank is just the number of sites. Using the maximum accounts for either case

src/decision/location_selection.jl
Zapiano: To answer your question I need to ask: @Rosejoycrocker what is the purpose of this function? If I understand correctly, a scenario with higher ranks will get a low value and vice-versa, right? So what is this going to be used for?

src/decision/location_selection.jl
Rosejoycrocker: @Zapiano the function `dim()` is not recognised - is it from base or a package?

Rosejoycrocker: Worked it out, it's from NamedDims

Rosejoycrocker: I don't think it does the same thing though, for example for a set of 8 scenarios
`[size(ranks, d) for d in [:scenarios]] = 8`
whereas
` prod(NamedDims.dim(ranks, [:scenarios])) = 2`

Am I using this correctly?

src/decision/location_selection.jl
Rosejoycrocker: Hey @Zapiano, this is the function you suggested in one of our meetings :)
For `ranks` the lower the rank the `better` the site for intervention. For example, 1 is highest ranked, 2 is second highest etc.
So the inverse ranks function instead gives a higher number for higher ranked sites.
As the values are also summed, it gives a measure of the number of times that site was highly ranked . It is also normalised.
This could be used to plot maps which summarise which sites were ranked highly most frequently with a number from 0-1 and hence can be easily compared.
I intend to submit another PR which includes functions to map these values as heatmaps.

src/decision/location_selection.jl
Zapiano: Oh, ok, my bad!

src/decision/location_selection.jl
ConnectedSystems: Oh I see, then we should document what the default behaviour is. Please clarify the docstring.

src/decision/location_selection.jl
Zapiano: Yes you are, I just mixed up `dim`s with `axes`. What I meant was  `axes(ranks)` but I guess that won't work well in this case because you can't pass an array of `axis` as argument ... So I think you should just stick to your first solution. :)

src/decision/location_selection.jl
Zapiano: It seems to me that this variable should be called something like `lowest_rank` (and maybe have a comment explaining that the lowest rank is the highest value?)

src/decision/location_selection.jl
Rosejoycrocker: Will do. Also should I add an `agg_func` instead of `dims` argument to `summed_inverse_rank()`? I'd have to change the normalisation but it would add flexibility (if desired).

src/decision/location_selection.jl
Zapiano: Yes, I know is that function, but that was a long time ago, so I'm trying to remember and get context again 😄 

So, this functions name, right now, is kind of a "summary" of the mathematical operations it does (it takes the inverse ranks and then sum it). But what it does "conceptually" sound more like an `agg_selection_score` (or `selection_score`) to me - trying to think if there's a better name for this. Because when I say I have the "aggregated selection score" for all locations, you know that is aggregation information about the selection (for all locations), and the greater the number the better.

What do you think?

src/decision/location_selection.jl
Rosejoycrocker: I've changed the name and added more explanation of the ranks :)

src/decision/location_selection.jl
Rosejoycrocker: All good haha, it was a while ago

I like the name `selection_score()`, what do you think @ConnectedSystems ?

src/decision/location_selection.jl
Zapiano: If we are doing this, we should call this variable `score` instead

src/decision/location_selection.jl
Zapiano: Also: what if instead of normalizing by `(n_ranks * prod([size(ranks, d) for d in dims]))` we normalize by `maximum(score)`? Then we would have score going from 0 to 1, with 0 being the worst and 1 the best. This might be easier to interpret in the future, what do you think?

Rosejoycrocker: Because we sum over dims, normalising by dividing by maximum(rank)*length(dim1)*length(dim2)... makes sure the values are between 0 and 1. This is because, for example, MAX(dropdims(sum(n_ranks .- ranks; dims=:scenarios); dims=:scenarios))) = n_ranks* (number of scenarios). Alternatively `inv_ranks = dropdims(sum((n_ranks .- ranks)/n_ranks; dims=dims); dims=dims[1])` should give the same result. Is that a clearer way of doing it?

Zapiano: Hum, two things:

1 - I might be doing some wrong math here, but shouldn't it be `(maximum(rank)-1)*length(dim1)*etc` ? Suppose you have 4 locations and 3 timesteps. The maximum score would be 4. The best case would be for a location to have ranks [1, 1, 1] (is that right? one for each `timestep`?), which would give a non-normalized score of 9 (and not 4*3 = 12).

2 - In any case, there's one question to consider: if in a particular run there was NO LOCATION with maximum possible score, than there will be no location with normalized score 1. I think that's the difference between doing what you are proposing and normalizig by "the greatest score". In the second option we ALWAYS will have some location with score 1. In the first option, not always. (I'm not saying we should do one or the other, just that this is the difference). @ConnectedSystems what do you think?

ConnectedSystems: I'll have to look at this more closely later, sorry, as I've lost context. Need to understand why `prod`...

src/decision/location_selection.jl
ConnectedSystems: `selection_score()` makes sense to me

src/decision/location_selection.jl
Rosejoycrocker: I've changed the types tp be more specific now and work for both the ecological model and standalone case

src/decision/location_selection.jl
Rosejoycrocker: To answer your questions @Zapiano , you're correct it should be -1. We could use the number of sites +1 as the set maximum value to avoid the case where the maximum rank changes. In this case it would be (n_sites*length(dim1)*length(dim2)... etc. I would prefer this as we could compare easily between runs. Let me know what you think.

src/decision/location_selection.jl
ConnectedSystems: > I've changed the types tp be more specific now and work for both the ecological model and standalone case

I'd then have to ask why they have to use different types?

Would this be resolved simply by passing in a `vec` instead?

Rosejoycrocker: For the standalone site selection the second argument is Vector{Float32} whereas for the ecological model it is Vector{Float64}

ConnectedSystems: `collect()` is what you want.

ConnectedSystems (approved):

ConnectedSystems:
Apologies for possible repeat spamming @Zapiano , my previous comment seems to have disappeared.

@Zapiano could you please do a final review?

Note that tests fail but are to be fixed by #539 and #546 .

Also note that changes unrelated to this PR seem to have been merged in or some reason, and that #434 has some relevant changes too.

Rosejoycrocker:
> 
> Also note that changes unrelated to this PR seem to have been merged in or some reason, and that #434 has some relevant changes too.

Sorry about the mixed changes, I'm not sure what happended because I only rebased with Main

src/decision/location_selection.jl
```diff
@@ -0,0 +1,296 @@
...
+
+# Arguments
+- `ranks` : Location ranks from `rank_locations()`
+- `n_ranks` : Number of rankings (default is number of locations).
+- `agg_func` : Aggregation function to appy after frequencies are calculated.
+
+# Returns
+Frequency with which each location was selected for each rank.
+"""
+function ranks_to_frequencies(ranks::NamedDimsArray, n_ranks::Int64)
```
ConnectedSystems, line 183: @Rosejoycrocker is `n_ranks` ever set by the user?

I can't think of a case where we'd want to set `n_ranks` less or more than the number of locations.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,296 @@
...
+- `domain` : The geospatial domain locations were selected from
+- `scenarios` : Scenario specification
+- `sum_cover` : Matrix[n_scenarios ⋅ n_sites] containing the total coral cover for each
+    location, for each scenario
+- `area_to_seed` : Area of coral to be seeded at each time step in km²
+- `agg_func` : Aggregation function to apply, e.g `ranks_to_frequencies`,
+    `ranks_to_location_order`
+- `iv_type` : ID of intervention (1 = seeding, 2 = fogging)
+- `target_seed_sites` : list of candidate locations for seeding (indices)
+- `target_shade_sites` : list of candidate location to shade (indices)
```
ConnectedSystems, line 70: Noting for later cleanup - there's mixed use of "fog" and "shade" here.

src/decision/location_selection.jl
Rosejoycrocker: My original thought with this is if you have a large number of sites (such as in the whole of GBR domain), you might only be interested in the frequencies of ranks 1:k where k is less than the total number of sites. This way you could save  time by not considering all the ranks. Not sure how relevant it is. I could set it as an optional parameter maybe?

ConnectedSystems: It already is an optional argument for the higher level functions. I was wondering if we could simplify the function signatures. I'm not sure if we need this flexibility but okay with leaving it in - might turn out to be useful.

src/decision/location_selection.jl
```diff
@@ -0,0 +1,298 @@
...
+    ranks::NamedDimsArray{D,T,3,A};
+    dims::Vector{Symbol}=[:scenarios, :timesteps],
+) where {D,T,A}
+    return _drop_single(selection_score(ranks, dims))
+end
+function selection_score(
+    ranks::NamedDimsArray{D,T,2,A};
+    dims::Vector{Symbol}=[:scenarios],
+) where {D,T,A}
+    return _drop_single(selection_score(ranks, dims))
```
Zapiano, line 289: I tried to run the code from the cookbook and I got an error in this line when running `sel_score = selection_score(ranks[intervention=1])`:

```julia
ERROR: MethodError: no method matching _drop_single(::NamedDims.NamedDimsArray{(:sites,), Float64, 1, AxisKeys.KeyedArray{Float64, 1, Vector{Float64}, Base.RefValue{UnitRange{Int64}}}})

Closest candidates are:
  _drop_single(::AbstractMatrix)
   @ ADRIA C:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\decision\location_selection.jl:260

Stacktrace:
 [1] selection_score(ranks::NamedDims.NamedDimsArray{(:sites, :scenarios), Float64, 2, AxisKeys.KeyedArray{Float64, 2, Matrix{Float64}, Tuple{UnitRange{Int64}, UnitRange{Int64}}}}; dims::Vector{Symbol})
   @ ADRIA C:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\decision\location_selection.jl:289
 [2] selection_score(ranks::NamedDims.NamedDimsArray{(:sites, :scenarios), Float64, 2, AxisKeys.KeyedArray{Float64, 2, Matrix{Float64}, Tuple{UnitRange{Int64}, UnitRange{Int64}}}})
   @ ADRIA C:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\decision\location_selection.jl:285
 [3] top-level scope
   @ c:\Users\pribeiro\AIMS\Code\ADRIA.jl\sandbox\cookbook\loc_agg_funcs.jl:36
```

Which leads me to: it would be better if the returning types were explicit. (In this case it would make debugging easier).

Zapiano: I also got the same error when running `rank_frequencies_seed = rank_locations(
    dom, scens, sum_cover, area_to_seed, selection_score, 1
)` also from the cookbook.

docs/src/usage/cookbook.md
```diff
@@ -117,19 +117,98 @@ rs = ADRIA.run_scenarios(dom, p_df, "45")
...
+rank_frequencies_seed = rank_locations(
+    dom, scens, sum_cover, area_to_seed, location_selection_frequencies, 1
+)
+rank_frequencies_seed = rank_locations(
+    dom, scens, sum_cover, area_to_seed, selection_score, 1
+)
+
+# Example using ADRIA runs
+scens = ADRIA.sample(dom, 8)
+rs = ADRIA.run_scenarios(dom, scens, "45")
```
Zapiano, line 189: I tried to run this here and got the following error, not sure if it is related to this PR though 

```julia
ERROR: BoundsError: attempt to access 75×216×50 NCDatasets.Variable{Float64, 3, NCDatasets.NCDataset{Nothing}} at index [:, :]
Stacktrace:
  [1] interpret_indices_disk(A::NCDatasets.Variable{Float64, 3, NCDatasets.NCDataset{Nothing}}, r::Tuple{Colon, Colon})
    @ DiskArrays C:\Users\pribeiro\.julia\packages\DiskArrays\QlfRF\src\diskarray.jl:114
  [2] getindex_disk(::NCDatasets.Variable{Float64, 3, NCDatasets.NCDataset{Nothing}}, ::Function, ::Vararg{Function})
    @ DiskArrays C:\Users\pribeiro\.julia\packages\DiskArrays\QlfRF\src\diskarray.jl:38
  [3] getindex
    @ C:\Users\pribeiro\.julia\packages\DiskArrays\QlfRF\src\diskarray.jl:211 [inlined]
  [4] getindex(::CommonDataModel.CFVariable{Union{Missing, Float64}, 3, NCDatasets.Variable{Float64, 3, NCDatasets.NCDataset{Nothing}}, NCDatasets.Attributes{NCDatasets.NCDataset{Nothing}}, NamedTuple{(:fillvalue, :missing_values, :scale_factor, :add_offset, :calendar, :time_origin, :time_factor), Tuple{Nothing, Tuple{Float64}, Vararg{Nothing, 5}}}}, ::Colon, ::Colon)
    @ CommonDataModel C:\Users\pribeiro\.julia\packages\CommonDataModel\pO4st\src\cfvariable.jl:390
  [5] load_nc_data(data_fn::String, attr::String, site_data::DataFrame)
    @ ADRIA C:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\io\inputs.jl:127
```

src/decision/dMCDA.jl
```diff
@@ -218,20 +248,33 @@ Get location ranks using mcda technique specified in mcda_func, weights and a de
 # Returns
 - `s_order` : [site_ids, criteria values, ranks]
 """
-function retrieve_ranks(S::Matrix{Float64}, site_ids::Vector{Float64}, weights::Vector{Float64}, mcda_func::Function)
+function retrieve_ranks(
+    S::Matrix{Float64},
+    site_ids::Vector{Float64},
+    weights::Vector{Float64},
+    mcda_func::Function)
```
Zapiano, line 255: Explicit returning type for these functions

src/decision/dMCDA.jl
```diff
@@ -401,7 +467,14 @@ Tuple (SH, wsh)
...
 """
-function create_shade_matrix(A, max_area, wt_conn_shade, wt_waves, wt_heat, wt_predec_shade, wt_predec_zones_shade, wt_hi_cover)
+function create_shade_matrix(A::Matrix{Float64},
+    max_area::Vector{Float64}, 
+    wt_conn_shade::T, 
+    wt_waves::T, 
+    wt_heat::T, 
+    wt_predec_shade::T, 
+    wt_predec_zones_shade::T, 
+    wt_hi_cover)where {T<:Float64}
```
Zapiano, line 477: Add a space before `where`

src/decision/dMCDA.jl
```diff
@@ -771,7 +739,15 @@ Here, `max_cover` represents the max. carrying capacity for each site (the `k` v
...
-function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shade_years, n_site_int, available_space, depth)
+function unguided_site_selection(
+    prefseedsites,
+    prefshadesites,
+    seed_years,
+    shade_years,
+    n_site_int,
+    available_space, 
+    depth
+)
```
Zapiano, line 750: Explicit returning type

src/decision/location_selection.jl
```diff
@@ -0,0 +1,298 @@
...
+
+# Arguments
+- `ranks` : Location ranks from `rank_locations()`
+- `n_ranks` : Number of rankings (default is number of locations).
+- `agg_func` : Aggregation function to appy after frequencies are calculated.
+
+# Returns
+Frequency with which each location was selected for each rank.
+"""
+function ranks_to_frequencies(ranks::NamedDimsArray, n_ranks::Int64)
```
Zapiano, line 182: Explicit returning type for these functions

Zapiano (changes requested):
@Rosejoycrocker 

Some small comments (basically asking to add returning type for some functions) and also I got some errors when trying to run the cookbook, not sure is this is because of the datapackage I'm using, if it is please let me know and feel free to mark those as resolved!

Zapiano:
> Apologies for possible repeat spamming @Zapiano , my previous comment seems to have disappeared.
> 
> @Zapiano could you please do a final review?
> 
> Note that tests fail but are to be fixed by #539 and #546 .
> 
> Please ignore the `examples/site_selection.jl` file - all examples should be moved into the documentation (e.g., as a cookbook recipe or similar), so the files in the `examples` folder will eventually be removed.
> 
> Also note that changes unrelated to this PR seem to have been merged in or some reason, and that #434 has some relevant changes too.

Done.

src/decision/location_selection.jl
ConnectedSystems: Could you try pulling again Pedro? I pushed some more changes after I asked you to review

Zapiano: There was no change since the version I had when I tested this - which was about an hour ago.

src/decision/location_selection.jl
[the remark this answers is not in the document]
ConnectedSystems: We agreed that interventions would be signified with `iv` as `int` can be confusing.

Similarly, standardize `ind` to `idx`.

ConnectedSystems:
Thanks for the review Pedro, I've addressed all your comments so am merging now.

PR #433, opened 2023-08-24
Fix time_series_clustering for variable series with k > 0

### LINKED ISSUE #428: Error: `non_null_sites`
Inside `analysis/clustering.jl` `time_series_clustering` we filter off sites without corals:

```julia
non_null_sites = result_set.site_data.k .> 0.0
filtered_data = data[:, non_null_sites]

# Apply time series cluster for filtered data
clusters = time_series_clustering(filtered_data, n_clusters)
```

But now this seems not to be working anymore. When I run this I get the error:

```
ERROR: ArgumentError: Distance matrix should be symmetric.
Stacktrace:
 [1] assertdistancematrix
   @ C:\Users\pribeiro\.julia\packages\Clustering\yuxBr\src\hclust.jl:48 [inlined]
 [2] hclust(d::Matrix{Float64}; linkage::Symbol, uplo::Nothing, branchorder::Symbol)
   @ Clustering C:\Users\pribeiro\.julia\packages\Clustering\yuxBr\src\hclust.jl:757
 [3] hclust
   @ C:\Users\pribeiro\.julia\packages\Clustering\yuxBr\src\hclust.jl:752 [inlined]
 [4] time_series_clustering(data::NamedDims.NamedDimsArray{(:timesteps, :sites), Float32, 2, AxisKeys.KeyedArray{Float32, 2, Matrix{Float32}, 
Tuple{Vector{Any}, Vector{Any}}}}, n_clusters::Int64)
   @ ADRIA.analysis C:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\analysis\clustering.jl:120
 ...
```

That is caused by the existence of series with all values equals 0.

Zapiano:
We were filtering off series whose result_set.site_data.k .> 0, but that doesn't prevent it from being constant and throwing an error.

Now we are checking if some series has complexity == 0 to filter off.

We are also throwing a warning when this is the case.

Closes #428


ext/AvizExt/viz/clustering.jl
```diff
@@ -32,8 +40,8 @@ function ADRIA.viz.ts_cluster!(g::Union{GridLayout,GridPosition}, data::Abstract
         axis_opts...
     )
 
-    # Filter clusters and data for non-zero clusters
-    clusters_filtered = filter(c -> c != 0, clusters)
+    # Filter off clusters with index -1  (those represent constant series)
```
ConnectedSystems, line 43:
```suggestion
    # Filter clusters with index value of "-1"  (these represent constant series)

```

ConnectedSystems: Actually, the code does exactly what you describe so not sure if this comment is needed?

src/analysis/clustering.jl
```diff
@@ -112,26 +113,32 @@ Vector of cluster ids indicating which cluster each scenario belongs to.
...
+    complexity = _complexity(data)
+    variable_series = complexity .!= 0
+
+    # Compute CID Matrix
+    filtered_data = data[:, variable_series]
+    filtered_complexity = complexity[variable_series]
+    distances = complexity_invariance_distance(filtered_data, filtered_complexity)
+
+    if 0 in variable_series
+        @warn "Some series on you data are constant. They will be put in cluster -1"
```
ConnectedSystems, line 130:
```suggestion
        @warn "Constant series detected. These will be clustered with id `-1`"

```

ConnectedSystems (approved):

Zapiano:
@ConnectedSystems We can discuss it later but I changed this PR so the minimum complexity value would be 1. The consequence of this change is that:

- When both series are constant, CF (correction factor) is 1
- When one of them is constant and the other isn't, CF is the complexity of the later (which now is slightly bigger than before)
- When both are not constant, the correction factor is computed the same way but the complexities now are different. That shouldn't affect the final result very much, specially given the amplitude of the series we are dealing with here

This is a different approach from the paper we were following so you may want to discuss it carefully before merging this one.

ConnectedSystems:
Summary of discussion:

Agree to change complexity factor to 1 + sqrt(). 

This changes metric so that the indicated range is between 1 and infinity, as opposed to 0 to infinity. The benefit is that the division by 0 error is avoided in cases where the series is constant.

PR #434, opened 2023-08-24
Add functions for repeated tasks when performing location selection

Rosejoycrocker:
Adds functions to find depth criteria and environmental layer means which are used in location selection in `run_site_selection()` and scenario.jl.

### DIFF
--- a/src/decision/dMCDA.jl
+++ b/src/decision/dMCDA.jl
@@ -552,19 +552,6 @@ function guided_site_selection(
     max_cover = d_vars.max_cover[site_ids]
     area = d_vars.area[site_ids]
 
-    risk_tol = d_vars.risk_tol
-    w_in_conn = d_vars.wt_in_conn_seed
-    w_out_conn = d_vars.wt_out_conn_seed
-    w_shade_conn = d_vars.wt_conn_shade
-    w_waves = d_vars.wt_waves
-    w_heat = d_vars.wt_heat
-    w_high_cover = d_vars.wt_hi_cover
-    w_low_cover = d_vars.wt_lo_cover
-    w_predec_seed = d_vars.wt_predec_seed
-    w_predec_shade = d_vars.wt_predec_shade
-    w_predec_zones_seed = d_vars.wt_zones_seed
-    w_predec_zones_shade = d_vars.wt_zones_shade
-
     # site_id, seeding rank, shading rank
     rankings = Int64[site_ids zeros(Int64, n_sites) zeros(Int64, n_sites)]
 
@@ -597,10 +584,18 @@ function guided_site_selection(
     zones_criteria = zone_preds .+ zone_sites
     mcda_func = methods_mcda[alg_ind]
 
-    A, filtered_sites = create_decision_matrix(site_ids, in_conn, out_conn, sum_cover, max_cover, area, wave_stress, heat_stress, site_depth, predec, zones_criteria, risk_tol)
+    A, filtered_sites = create_decision_matrix(
+        site_ids, in_conn, out_conn, sum_cover, max_cover,
+        area, wave_stress, heat_stress, site_depth, predec,
+        zones_criteria, d_vars.risk_tol
+    )
     if isempty(A)
         # if all rows have nans and A is empty, abort mission
-        return zeros(Int64, length(prefseedsites)), zeros(Int64, length(prefshadesites)), rankingsin
+        return (
+            zeros(Int64, length(prefseedsites)),
+            zeros(Int64, length(prefshadesites)),
+            rankingsin
+        )
     end
 
     # cap to number of sites left after risk filtration
@@ -608,13 +603,20 @@ function guided_site_selection(
 
     # if seeding, create seeding specific decision matrix
     if log_seed
-        SE, wse = create_seed_matrix(A, d_vars.min_area, w_in_conn, w_out_conn, w_waves, w_heat, w_predec_seed, w_predec_zones_seed, w_low_cover)
+        SE, wse = create_seed_matrix(
+            A, d_vars.min_area, d_vars.wt_in_conn_seed, d_vars.wt_out_conn_seed,
+            d_vars.wt_waves, d_vars.wt_heat, d_vars.wt_predec_seed, d_vars.wt_zones_seed,
+            d_vars.wt_lo_cover
+        )
     end
 
     # if shading, create shading specific decision matrix
     if log_shade
         max_area = (area.*max_cover)[filtered_sites]
-        SH, wsh = create_shade_matrix(A, max_area, w_shade_conn, w_waves, w_heat, w_predec_shade, w_predec_zones_shade, w_high_cover)
+        SH, wsh = create_shade_matrix(
+            A, max_area, d_vars.wt_conn_shade, d_vars.wt_waves, d_vars.wt_heat,
+            d_vars.wt_predec_shade, d_vars.wt_zones_shade, d_vars.wt_hi_cover
+        )
     end
 
     if log_seed && isempty(SE)
--- a/src/decision/dMCDA.jl
+++ b/src/decision/dMCDA.jl
@@ -906,7 +906,7 @@ Calculates mean over specified dimensions plus half the standard deviation.
 - `w` : Weighting for std offset to mean.
 
 # Returns
-Mean + std of input env_layer, where std is weighted by (1-w), mean is weighted by
+Weighted mean + std of input env_layer, where std is weighted by (1-w), mean is weighted by
 w and aggregated over dims.
 """
 function summary_stat_env(
@@ -928,7 +928,7 @@ Calculates criteria for depth filtering in MCDA.
 - `depth_min` : Minimum depth for each considered location.
 
 # Returns
-Vector of indices indicating locations which satisfy the depth criteria.
+Vector of logical indices indicating locations which satisfy the depth criteria.
 """
 function depth_criteria_idx(
     depth_med::Vector{T}, depth_max::T, depth_min::T
--- a/src/decision/dMCDA.jl
+++ b/src/decision/dMCDA.jl
@@ -246,7 +246,7 @@ Get location ranks using mcda technique specified in mcda_func, weights and a de
 - `maximize` : Boolean indicating whether a mcda method is maximizing score (true), or minimizing (false).
 
 # Returns
-`s_order` : [site_ids, criteria values, ranks]
+Matrix, containing site_ids, criteria values, ranks
 """
 function retrieve_ranks(
     S::Matrix{Float64},
@@ -363,6 +363,7 @@ function create_decision_matrix(
 
     # Remove rows with NaNs
     A = A[filtered, :]
+
     return A, filtered
 end
 
@@ -509,9 +510,9 @@ end
 
 # Returns
 Tuple :
-    - `prefseedsites` : n_site_int highest ranked seeding sites
-    - `prefshadesites` : n_site_int highest ranked shading/fogging sites
-    - `rankings` : n_sites ⋅ 3 matrix holding [site_id, seeding_rank, shading_rank],
+    - `prefseedsites` : Vector, Indices of preferred seeding locations
+    - `prefshadesites` : Vector, Indices of preferred shading locations
+    - `rankings` : Matrix[n_sites ⋅ 3] where columns are site_id, seeding_rank, shading_rank
         Values of 0 indicate sites that were not considered
 """
 function guided_site_selection(
@@ -539,7 +540,7 @@ function guided_site_selection(
         return zeros(Int64, length(prefseedsites)), zeros(Int64, length(prefshadesites)), rankingsin
     end
 
-    n_site_int::Int64 = d_vars.n_site_int
+    n_iv_locs::Int64 = d_vars.n_site_int
     priority_sites::Array{Int64} = d_vars.priority_sites[in.(d_vars.priority_sites, [site_ids])]
     priority_zones::Array{String} = d_vars.priority_zones
 
@@ -603,7 +604,7 @@ function guided_site_selection(
     end
 
     # cap to number of sites left after risk filtration
-    n_site_int = min(n_site_int, length(A[:, 1]))
+    n_iv_locs = min(n_iv_locs, length(A[:, 1]))
 
     # if seeding, create seeding specific decision matrix
     if log_seed
@@ -617,9 +618,9 @@ function guided_site_selection(
     end
 
     if log_seed && isempty(SE)
-        prefseedsites = zeros(Int64, n_site_int)
+        prefseedsites = zeros(Int64, n_iv_locs)
     elseif log_seed
-        prefseedsites, s_order_seed = rank_sites!(SE, wse, rankings, n_site_int, mcda_func, 2)
+        prefseedsites, s_order_seed = rank_sites!(SE, wse, rankings, n_iv_locs, mcda_func, 2)
         if use_dist != 0
             prefseedsites, rankings = distance_sorting(
                 prefseedsites, s_order_seed, d_vars.dist, min_dist, rankings, 2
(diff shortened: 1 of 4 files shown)


src/decision/dMCDA.jl
```diff
@@ -786,3 +786,31 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+
+Calculates mean over specified dimensions plus half the standard deviation.
+
+# Arguments
+- `env_layer` : Environmental data layer to calculate the mean of.
+- `dims_agg` : Dimensions to aggregate over.
+"""
+function env_mean(env_layer::AbstractArray, dims_agg::Union{Symbol,Tuple{Symbol,Symbol},Int,Tuple{Int,Int}})
+    return vec((mean(env_layer, dims=dims_agg) .+ std(env_layer, dims=dims_agg)) .* 0.5)
+end
```
ConnectedSystems, line 802: `env_mean()` isn't a very clear name for this - it does more than just take the mean.

@Zapiano any suggestions here?

src/decision/dMCDA.jl
```diff
@@ -786,3 +786,31 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+    set_depth_criteria(depth_med::Vector{Float64}, depth_max::Float64, depth_min::Float64)
+
+Sets criteria for depth filtering in MCDA.
+
+# Arguments
+- `depth_med` : Median depth (length n_locs).
+- `depth_max` : Maximum depth.
+- `depth_min` : Minimum depth.
+"""
+function set_depth_criteria(depth_med::Vector{Float64}, depth_max::Float64, depth_min::Float64)
```
ConnectedSystems, line 814:
```suggestion
function set_depth_criteria(depth_med::Vector{T}, depth_max::T, depth_min::T)::Vector{T} where {T<:Float64}
```

src/decision/dMCDA.jl
ConnectedSystems: I'm also not comfortable with name and usage of `dims_agg`.

If `env_layer` has named dimensions, we should preference using those instead of numeric IDs for clarity.

src/decision/dMCDA.jl
```diff
@@ -786,3 +786,31 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+
+
+"""
+    env_mean(env_layer::AbstractArray, dims_agg::Union{Symbol,Tuple{Symbol,Symbol},Int,Tuple{Int,Int}})
+
+Calculates mean over specified dimensions plus half the standard deviation.
+
+# Arguments
+- `env_layer` : Environmental data layer to calculate the mean of.
+- `dims_agg` : Dimensions to aggregate over.
```
ConnectedSystems, line 798: Add `# Returns` section please

src/decision/dMCDA.jl
```diff
@@ -786,3 +786,31 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+"""
+    set_depth_criteria(depth_med::Vector{Float64}, depth_max::Float64, depth_min::Float64)
+
+Sets criteria for depth filtering in MCDA.
+
+# Arguments
+- `depth_med` : Median depth (length n_locs).
+- `depth_max` : Maximum depth.
+- `depth_min` : Minimum depth.
+"""
```
ConnectedSystems, line 813: Add `# Returns` section please

src/decision/dMCDA.jl
Zapiano: @Rosejoycrocker you are passing the result of this function to `site_selection` and the description of these parameters there are: "array of length nsites containing wave/dhw scenario.". It is not clear to me what do you mean by "length nsites". Maybe answering that could help us think of a better name for this function...?

src/decision/dMCDA.jl
```diff
@@ -689,16 +689,16 @@ function run_site_selection(dom::Domain, scenarios::DataFrame, sum_cover::Abstra
 
     n_sites = length(dom.site_ids)
     for (scen_idx, scen) in enumerate(eachrow(scenarios))
-        depth_criteria = (dom.site_data.depth_med .<= scen.max_depth) .& (dom.site_data.depth_med .>= scen.depth_min)
+        depth_criteria = set_depth_criteria(dom.site_data.depth_med, scen.max_depth, scen.depth_min)
         depth_priority = findall(depth_criteria)
 
         considered_sites = target_site_ids[findall(in(depth_priority), target_site_ids)]
 
         ranks_store(scenarios=scen_idx, sites=dom.site_ids[considered_sites]) .= site_selection(
```
Zapiano, line 697: I think this code is a little hard to read and in this case it could be useful to assign each function call to a variable and then do the comparison on a separate line..? If you do this with at least `size_selection` it will already be better. Something like the following:

```julia
selected_sites = site_selection(
    dom,
    scen,
    env_mean(wave_scens[:, :, target_wave_scens], (:timesteps, :scenarios)),
    env_mean(dhw_scens[:, :, target_dhw_scens], (:timesteps, :scenarios)),
    considered_sites,
    sum_cover[scen_idx, :],
    area_to_seed
)

ranks_store(scenarios=scen_idx, sites=dom.site_ids[considered_sites]) .= selected_sites
```

src/decision/dMCDA.jl
```diff
@@ -689,16 +689,16 @@ function run_site_selection(dom::Domain, scenarios::DataFrame, sum_cover::Abstra
 
     n_sites = length(dom.site_ids)
```
Zapiano, line 690: I think this wasn't caused by this commit BUT this variable seems no to be used.

src/decision/dMCDA.jl
Zapiano: You could call it `mean_var_offset`..? And also, does it make sense for this parameter `0.5` to be passed as an argument with default value equals `0.5`? (maybe this doesn't make sense in this context, I thought about it because it's an arbitrary number)

Rosejoycrocker: > I'm also not comfortable with name and usage of `dims_agg`, as:
> 
> * `dims` is the convention.
> * If `env_layer` has named dimensions, we should preference using those instead of numeric IDs for clarity.
> * If `env_layer` is always a named dimension, then constrain the type further to `NamedDimsArray`.

I'll see if env_layer in scenario.jl in actually a NamedDimsArray, I just copied the dimensions already used, but might be able to get rid of the input.

Rosejoycrocker: > @Rosejoycrocker you are passing the result of this function to `site_selection` and the description of these parameters there are: "array of length nsites containing wave/dhw scenario.". It is not clear to me what do you mean by "length nsites". Maybe answering that could help us think of a better name for this function...?

Good point, they're basically a representation of dhws/waves that don't require looping over timesteps and performing site selection for each replicate of the dhws/waves. `mean_var_offset` could work... We only use 0.5 currently, but could be good to have the option to use other values.

src/decision/dMCDA.jl
Rosejoycrocker: Can definitely do this

ConnectedSystems: Yes, it's more readable, but hurts performance (I expect this approach to use double the memory allocations). MCDA is a performance critical part so I would prefer a comment over readable code in this case.

That said, there are definitely other things here that hurt performance more than the readable code, so I'm fine to concede to @Zapiano 's point here.

src/decision/dMCDA.jl
ConnectedSystems: > `mean_var_offset` could work... We only use 0.5 currently, but could be good to have the option to use other values.

It's not an offset though. It summarizes the expected conditions $n$ number of years into the future.

ConnectedSystems: Oh and about the `0.5` it's just an equal weight between the mean and stdev, as I did this in a hurry.

If you want to consider it as something that can be adjusted, it would make more sense to flesh it out properly:

```julia
w = 0.5

vec((mean(env_layer, dims=dims_agg) .* w) .+ (std(env_layer, dims=dims_agg) .* (1.0 - w)))
```

src/decision/dMCDA.jl
Zapiano: @ConnectedSystems why does it use double the memory?

Rosejoycrocker:
I think I've addressed your comments, ley me know if there's anything else

src/decision/dMCDA.jl
```diff
@@ -784,3 +783,37 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+# Arguments
+- `env_layer` : Environmental data layer to calculate the mean of.
+- `dims` : Dimensions to aggregate over.
+- `w` : Weighting for std offset to mean.
+
+# Returns
+- Mean + std of input env_layer, where std is weighted by (1-w) and aggregated over dims.
+"""
+function mean_var_offset(env_layer::NamedDimsArray dims::Union{Symbol,Tuple{Symbol,Symbol}}; w=0.5)
+    return vec((mean(env_layer, dims=dims) .+ std(env_layer, dims=dims)) .* (1.0 - w))
```
ConnectedSystems, line 801: This does not do a weighted aggregation

https://github.com/open-AIMS/ADRIA.jl/pull/434#discussion_r1308511361

src/decision/dMCDA.jl
```diff
@@ -784,3 +783,37 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+
+# Arguments
+- `env_layer` : Environmental data layer to calculate the mean of.
+- `dims` : Dimensions to aggregate over.
+- `w` : Weighting for std offset to mean.
+
+# Returns
+- Mean + std of input env_layer, where std is weighted by (1-w) and aggregated over dims.
+"""
+function mean_var_offset(env_layer::NamedDimsArray dims::Union{Symbol,Tuple{Symbol,Symbol}}; w=0.5)
```
ConnectedSystems, line 800: The function is not intended to calculate an offset

https://github.com/open-AIMS/ADRIA.jl/pull/434#discussion_r1308503109

src/decision/dMCDA.jl
```diff
@@ -784,3 +783,37 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
 
     return prefseedsites[prefseedsites.>0], prefshadesites[prefshadesites.>0]
 end
+
+
+"""
+    mean_var_offset(env_layer::NamedDimsArray dims::Union{Symbol,Tuple{Symbol,Symbol}}; w=0.5)
+Calculates mean over specified dimensions plus half the standard deviation.
```
ConnectedSystems, line 790: Please follow style guide

src/decision/dMCDA.jl
```diff
@@ -685,18 +685,17 @@ function run_site_selection(dom::Domain, scenarios::DataFrame, sum_cover::Abstra
         append!(target_site_ids, target_shade_sites)
     end
 
-    n_sites = length(dom.site_ids)
     for (scen_idx, scen) in enumerate(eachrow(scenarios))
-        depth_criteria = (dom.site_data.depth_med .<= scen.max_depth) .& (dom.site_data.depth_med .>= scen.depth_min)
+        depth_criteria = set_depth_criteria(dom.site_data.depth_med, scen.max_depth, scen.depth_min)
```
ConnectedSystems, line 689: An issue I have with the function name here is that it does not `set` anything.

src/decision/dMCDA.jl
```diff
@@ -784,3 +783,37 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+
+# Arguments
+- `depth_med` : Median depth (length n_locs).
+- `depth_max` : Maximum depth.
+- `depth_min` : Minimum depth.
+
+# Returns
+- Vector of indices indicating locations which satisfy the depth criteria.
+"""
+function set_depth_criteria(depth_med::Vector{T}, depth_max::T, depth_min::T)::Vector{T} where {T<:Float64}
```
ConnectedSystems, line 817: See comment elsewhere - this function does not set anything.

src/decision/dMCDA.jl
Rosejoycrocker: typo- I'll fix this

Rosejoycrocker:
I think I've addressed requested changes

src/decision/dMCDA.jl
```diff
@@ -792,3 +799,43 @@ function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shad
...
+# Arguments
+- `env_layer` : Environmental data layer to calculate the mean of.
+- `dims` : Dimensions to aggregate over.
+- `w` : Weighting for std offset to mean.
+
+# Returns
+- Mean + std of input env_layer, where std is weighted by (1-w), mean is weighted by
+w and aggregated over dims.
+"""
+function sum_stat_env(
```
ConnectedSystems, line 818: This is named `sum_stat_env()`, which is different from its usage elsewhere.

Is there a test for this?

Rosejoycrocker: This is due to the function name being requested to be changed multiple times. I've fixed it now, it should be summary_stat_env

src/decision/dMCDA.jl
```diff
@@ -689,18 +689,17 @@ function run_site_selection(dom::Domain, scenarios::DataFrame, sum_cover::Abstra
         append!(target_site_ids, target_shade_sites)
     end
 
-    n_sites = length(dom.site_ids)
     for (scen_idx, scen) in enumerate(eachrow(scenarios))
-        depth_criteria = (dom.site_data.depth_med .<= scen.max_depth) .& (dom.site_data.depth_med .>= scen.depth_min)
+        depth_criteria = depth_criteria_indx(dom.site_data.depth_med, scen.max_depth, scen.depth_min)
```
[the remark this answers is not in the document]
Rosejoycrocker: Because they are logical indices for the sites

Rosejoycrocker:
Comments have been addressed. Running fine on my end with ecological model and standalone site selection. It seems the type of wave_scens in the ecological model has changed since I last looked at this, making the aggregation dim different from dhw_scens in scenario.jl. I tried to make them the same but it seems dhw_scens is a matrix becuase of the cache and I imagine the cache has to be a matrix? (whereas dhw_scens is a NamedDimsArray). Tests themselves aren't passing but this seems to be due to https://github.com/open-AIMS/ADRIA.jl/pull/539 .

ConnectedSystems:
> I tried to make them the same but it seems dhw_scens is a matrix becuase of the cache and I imagine the cache has to be a matrix? (whereas dhw_scens is a NamedDimsArray)

I'm not sure if I follow this - I think I need some links to the code for clarity.

Rosejoycrocker:
> > I tried to make them the same but it seems dhw_scens is a matrix becuase of the cache and I imagine the cache has to be a matrix? (whereas dhw_scens is a NamedDimsArray)
> 
> I'm not sure if I follow this - I think I need some links to the code for clarity.

I was just looking into recreating this to explain it and I'm running into another error after rebasing so I'll get back to you when I've sorted that

Rosejoycrocker:
> > I tried to make them the same but it seems dhw_scens is a matrix becuase of the cache and I imagine the cache has to be a matrix? (whereas dhw_scens is a NamedDimsArray)
> 
> I'm not sure if I follow this - I think I need some links to the code for clarity.

So because dhw_scens is defined without any normalisation here:
https://github.com/open-AIMS/ADRIA.jl/blob/1645f3f1eb8b3fe3f373e0478fca7aec3b1c805a/src/scenario.jl#L413
it remains a NamedDimsArray and uses a symbol for the aggregation dim when finding the vector to be used in mcda.

wave_scens is defined differently, with normalisation here:
https://github.com/open-AIMS/ADRIA.jl/blob/ffee4d9c7ee742a599095d70b109bc9a6d546608/src/scenario.jl#L586-L588
which converts it to a Matrix. This means the aggregation dim is an integer. I tried removing the `Matrix()` on the second line but it is still converted to a Matrix mean divided by the maximum. This is the reason for the aggregation dimensions being different types for the wave and dhw scenario aggregation lines.

ConnectedSystems (approved):
Done

PR #436, opened 2023-08-25
Update prototype exploration UI to work with recent changes

ConnectedSystems:
As in title.

Includes minor fix to make locations appear properly for spatial visualizations


ConnectedSystems:
@Zapiano this should get the UI working again so the figure in the documentation can be updated

Zapiano (approved):

Issue #440, opened 2023-08-29
`viz.outcome_map()` gives duplicate x-axis values when plotting `:guided` factor


Rosejoycrocker:
When plotting factor mapping outcomes for the guided factor, `viz.outcome_map()`  uses 
```julia
fv_s = quantile(f_vals, b_slices)
```
where `f_vals` are the scenario values for the guided factor and `b_slices` are the bins for factor mapping. `f_vs` gives duplicate values but is used as the x-axis when plotting.
For example if
```julia
infil> f_vals
8192-element Vector{Float64}:
  5.0
 14.0
  1.0
 10.0
 16.0
  ⋮
  7.0
  1.0
 10.0
  5.0
 14.0

infil> b_slices
18-element Vector{Float64}:
 0.06
 0.11
 0.17
 0.22
 0.28
 ⋮
 0.78
 0.83
 0.89
 0.94
 1.0

infil> quantile(sort(f_vals), b_slices)
18-element Vector{Float64}:
  2.0
  2.0
  4.0
  4.0
  6.0
  ⋮
 15.0
 15.0
 17.0
 17.0
 18.0
```

PR #442, opened 2023-08-30
Fix duplicate x-axis values bug in `viz.outcome_map()`

Rosejoycrocker:
Resolves issue #440.

Previously, plotting factors where the splitting of the factor domain had uneven steps (due to rounding used to define mapping labels possibly), duplicate values would arise when creating the x-axis for outcome mapping plots (see issue). 

Previously plots of factor with uneven splitting would look like this:

outcome_map_guided_moore_all_time

with duplicate x-values.
Current plots now look like this:
outcome_map_guided_moore_alltime

without duplicate x-values.

### DIFF
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -202,7 +202,7 @@ function ADRIA.viz.rsa!(g::Union{GridLayout,GridPosition}, rs::ResultSet, si::Na
     for r in 1:n_rows
         for c in 1:n_cols
             f_vals = rs.inputs[:, Symbol(factors[curr])]
-            fv_s = quantile(f_vals, b_slices)
+            fv_s = round.(quantile(f_vals, b_slices),digits=2)
             # fv_s = String[(i == 1) || iseven(i) ? @sprintf("%.1f", fv) : "" for (i, fv) in enumerate(quantile(f_vals, b_slices))]
             # xtick_labels = (1:length(bin_slices), fv_s)
 
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -202,7 +202,7 @@ function ADRIA.viz.rsa!(g::Union{GridLayout,GridPosition}, rs::ResultSet, si::Na
     for r in 1:n_rows
         for c in 1:n_cols
             f_vals = rs.inputs[:, Symbol(factors[curr])]
-            fv_s = round.(quantile(f_vals, b_slices),digits=2)
+            fv_s = quantile(f_vals, b_slices)
             # fv_s = String[(i == 1) || iseven(i) ? @sprintf("%.1f", fv) : "" for (i, fv) in enumerate(quantile(f_vals, b_slices))]
             # xtick_labels = (1:length(bin_slices), fv_s)
 
@@ -310,7 +310,7 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
         for c in 1:n_cols
             f_name = Symbol(factors[curr])
             f_vals = rs.inputs[:, f_name]
-            fv_s = quantile(f_vals, b_slices)
+            fv_s = round.(quantile(f_vals, b_slices), digits=2)
 
             ax::Axis = Axis(
                 g[r, c],
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -302,7 +302,8 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
     end
 
     bin_slices, factor_list, CIs = axiskeys(outcomes)
-    b_slices = parse.(Float64, bin_slices)
+    b_slices = collect(LinRange(0.0, 1.0, length(bin_slices)))
+
     curr::Int64 = 1
     axs = Axis[]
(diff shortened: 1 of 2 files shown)


ConnectedSystems:
The proposed fix seems to just hide the issue. It also looks like incorrect/misleading bin ranges will be shown for non-categorical factors, though I haven't checked this directly.

Instead of replacing the indicated bin ranges, we should address the cause of duplicate values.

ConnectedSystems:
One other thing - can you check your scenario spec to see if method 5 is included, and the associated outcomes?

It seems to be missing in your plot so worried there's something else wrong here.

Rosejoycrocker:
> One other thing - can you check your scenario spec to see if method 5 is included, and the associated outcomes?
> 
> It seems to be missing in your plot so worried there's something else wrong here.

Can confirm method 5 is represented in the scenarios, not sure why the output is assigned "missing" in the outcome mapping.

Rosejoycrocker:
> The proposed fix seems to just hide the issue. It also looks like incorrect/misleading bin ranges will be shown for non-categorical factors, though I haven't checked this directly.
> 
> Instead of replacing the indicated bin ranges, we should address the cause of duplicate values.

I looked into the cause before making these changes and it happens because of rounding up when assigning the labels for the mapping intervals when creating the NamedDimsArray in `outcome_map()`. The rounding up means the stepsize varies between intervals, so soem values are determined to fall in the wrong interval when `quantile()` is used. You're right thouh, the splitting is slightly off for continuous variables, so we might need to introduce a check for categorical vs. continuous variables, maybe using spec(). I'll look into this.

ConnectedSystems:
> The rounding up means the stepsize varies between intervals, so some values are determined to fall in the wrong interval when quantile() is used

"Easy" fix is to not round up and only display rounded values in the visualisation.

Rosejoycrocker:
> > The rounding up means the stepsize varies between intervals, so some values are determined to fall in the wrong interval when quantile() is used
> 
> "Easy" fix is to not round up and only display rounded values in the visualisation.

I ended up increasing the rounding number of decimals so the outcome map table labels aren't too long but rounding doesn't effect step size too much.

ConnectedSystems:
Hi Rose,

What I meant was:

1. Do not round anything in the step that calculates sensitivity values
2. When creating the figure, round the values for display.

The current proposed approach just delays the issue - hypothetically it's entirely possible to have non-unique values at the 5th decimal place.

Rosejoycrocker:
> Hi Rose,
> 
> What I meant was:
> 
> 1. Do not round anything in the step that calculates sensitivity values
> 2. When creating the figure, round the values for display.
> 
> The current proposed approach just delays the issue - hypothetically it's entirely possible to have non-unique values at the 5th decimal place.

But also hypothetically the number of decimal places could be very long and the labels for the bins are used to create the x-axis, so doesn't this risk having huge labels on the NamedDimsArray?

ConnectedSystems:
> the labels for the bins are used to create the x-axis

Yes. That's why I suggest to round the values when creating the figure, i.e., round the values for the x-axis tick labels.

Rosejoycrocker:
Ok, I'll do that. I just thought you wouldn't want huge labels

src/analysis/sensitivity.jl
```diff
@@ -314,7 +314,7 @@ function outcome_map(X::DataFrame, y::AbstractVecOrMat{T}, rule::V, target_facto
 
     p_table = NamedDimsArray(
         zeros(Union{Missing,Float64}, length(steps) - 1, length(target_factors), 3);
-        bins=["$(round(i, digits=2))" for i in steps[2:end]],
+        bins=["$i" for i in steps[2:end]],
```
ConnectedSystems, line 317:
```suggestion
        bins=string.(steps[2:end]),
```

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -234,7 +234,7 @@ function ADRIA.viz.rsa!(g::Union{GridLayout,GridPosition}, rs::ResultSet, si::Na
     for r in 1:n_rows
         for c in 1:n_cols
             f_vals = rs.inputs[:, Symbol(factors[curr])]
-            fv_s = quantile(f_vals, b_slices)
+            fv_s = round.(quantile(f_vals, b_slices),digits=2)
```
ConnectedSystems, line 237: Follow style guide please. Spaces between arguments and commas for readability.

Rosejoycrocker:
Requested changes made. Here's what a categorical plot looks like:
outcome_map_guided_moore_6599

and a continuous plot:
outcome_map_wavestress_moore_6599

ConnectedSystems:
Hi Rose, please raise an issue to correct the same bug for regional sensitivity analysis

Zapiano:
Hi @Rosejoycrocker, I tried to replicate this bug but I couldn't. Can you tell me how many scenarios did you run and what exactly did you run to get those duplicate values you show in [Issue#440](https://github.com/open-AIMS/ADRIA.jl/issues/440)? 

I suspect this might not be a bug but a consequence of the way quantiles are estimated AND the fact the the input array has a lot of duplicate values. For example:

```julia
julia> quantile([1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3, 3], 0:0.1:1)
11-element Vector{Float64}:
 1.0
 1.0
 1.0
 ⋮
 3.0
 3.0
 3.0
```

So.. I maybe we need to discuss what to do with that

cc: @ConnectedSystems

Zapiano (approved):

PR #446, opened 2023-09-02
More updates to documentation

ConnectedSystems:
This PR includes/adds a cookbook page and uses a workaround to include documentation for metrics which use the `Metric` type.
As a bonus, the updated approach to defining metrics makes the docstring available for the built-in Julia help system.

I've noted that documentation for metrics are often incomplete or inconsistently defined. This will be addressed in a later PR.


ConnectedSystems:
Merging directly as the changes should not affect anything other than documentation.

PR #455, opened 2023-09-05
Miscellaneous changes to improve type infererability and precompilation times

ConnectedSystems:
Has minimal effect overall in terms of reducing compilation times, but still good to tidy these things up.


ConnectedSystems:
Merging in directly as changes are focused on cleaning up defunct code/packages and (should) have no impact.

Tests all pass too.

PR #460, opened 2023-09-05
Fix print_rules condition

Zapiano:
Instead of checking if a clause second element was :L or :R, it was checking the whole clause. which was always false. That was causing the prints to always use the "greater than or equal" for all clauses.

### DIFF
--- a/src/analysis/rule_extraction.jl
+++ b/src/analysis/rule_extraction.jl
@@ -87,7 +87,7 @@ compact form.
 - `rules` : Vector of Rule objects
 """
 function print_rules(rules::Vector{Rule{Vector{Vector},Vector{Float64}}})
-    condition(rule) = [c == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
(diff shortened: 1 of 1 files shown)


src/analysis/rule_extraction.jl
```diff
@@ -87,7 +87,7 @@ compact form.
 - `rules` : Vector of Rule objects
 """
 function print_rules(rules::Vector{Rule{Vector{Vector},Vector{Float64}}})
-    condition(rule) = [c == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
+    condition(rule) = [c[2] == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
     consequent(rule) = " then $(rule.consequent[1]) else $(rule.consequent[2])\n"
     rule_string(rule) = "if " * join(condition(rule), " & ") * consequent(rule)
```
ConnectedSystems, line 92: Could you add a return type to the function signature (`::Nothing`) and a corresponding return statement?

Zapiano: Done

ConnectedSystems (approved):

PR #461, opened 2023-09-06
Update ADRIA logo

Zapiano:
This is how it looks on a white background
image

And on a dark background

image


ConnectedSystems (approved):
Thanks!

PR #466, opened 2023-09-08
Update Reef Tourism Index regression coefficient (temporary solution)

ConnectedSystems:
Stopgap update to coefficients used for the Reef Tourism Index metric.

Notes from Ryan:
Note the coral juveniles coefficient is not significant. The overall model fit isn’t great either – 13.4% R-squared. 
This is a very ad-hoc solution for the tourism indicator until something better is thought of.


ConnectedSystems:
@Zapiano merging in directly as merging this does not affect any other use case except for Ryan's analysis.

PR #469, opened 2023-09-09
Precompilation reduction

ConnectedSystems:
**Note: Julia environment will need to be updated after this PR is merged.**

This PR includes a number of changes in an attempt to reduce initial precompilation times.

The principal change is that the depedency on DifferentialEquations.jl is replaced with OrdinaryDiffEq.jl, following the advice outlined [here](https://docs.sciml.ai/DiffEqDocs/stable/features/low_dep/).
In short, DifferentialEquations is a very large meta-package which offers many features/capabilities.
However, we only use the ODE solvers which are provided by the sub-package OrdinaryDiffEq.

In addition to this dependency change, I have modified the ADRIA precompilation directives to avoid precompiling domain loading and scenario running methods.
Further changes to improve inferability of specific methods are also included (specifically, defining concrete types where possible).
For the most part, I have treated changes to update code style to Blue as **out of scope** for the purpose of this PR.

I have experienced:

- A reduction of initial package compilation times (in a completely fresh environment) from ~10 mins down to ~200 seconds (~ 3 mins 20 secs)
- An initial package load time of ~60 seconds (down from ~2 mins, which may happen if changes are made to the code)
- A package load time of ~10-15 seconds if no changes are made

The tradeoff of not precompiling domain loading and scenario running methods is that:
 
- Loading domains will take a little longer (+ 10-20 seconds)
- Set up and model runs will take a little longer (+ 20-30 seconds)

This is balanced by the fact that we're not always loading ADRIA for model runs. 
Sometimes we just want to do some analyses, or visualize existing results.


src/ecosystem/Ecosystem.jl
```diff
@@ -33,20 +33,20 @@ end
 Update a dataframe of parameters.
 Length of `u` is expected to match number of columns in `df`.
 """
-function map_to_discrete!(df::DataFrame, u::Union{AbstractVector,Tuple})::Nothing
+function map_to_discrete!(df::DataFrame, u::Union{AbstractVector{Union{Int64,Float64}},Tuple})::Nothing
```
Zapiano, line 36: Function name bigger than 92 characters

src/ecosystem/Ecosystem.jl
```diff
@@ -33,20 +33,20 @@ end
...
 end
 
-struct EnvironmentalLayer{P} <: EcoModel
+struct EnvironmentalLayer{P<:Param} <: EcoModel
     dhw_scenario::P
     wave_scenario::P
 end
 
-function EnvironmentalLayer(dhw::AbstractArray, wave::AbstractArray)
+function EnvironmentalLayer(dhw::AbstractArray{T}, wave::AbstractArray{T2})::EnvironmentalLayer where {T<:Union{Float32,Float64},T2<:Union{Float32,Float64}}
```
Zapiano, line 47: Function definition with more than 92 characters

src/io/ResultSet.jl
```diff
@@ -15,15 +15,15 @@ const ENV_STATS = "env_stats"
...
-struct ResultSet{S,T1,T2,F,A,B,C,D,G,D1,D2,D3,DF}
-    name::S
-    RCP::S
-    invoke_time::S
-    ADRIA_VERSION::S
+struct ResultSet{T1,T2,A,B,C,D,G,D1,D2,D3,DF}
+    name::String
+    RCP::String
+    invoke_time::String
+    ADRIA_VERSION::String
```
Zapiano, line 22: I'm curious about this change. Is this to improve performance in some way?

src/utils/text_display.jl
```diff
@@ -13,9 +13,8 @@ Returns a copy of original array so input is not modified.
 # Returns
 - converted_names : array[str], of cleaned parameter names
 """
-function human_readable_name(names::Array{String}; title_case::Bool=false)::Array{String}
-    converted_names = names[:]
-    converted_names = map((x) -> replace(x, "_" => " "), converted_names)
+function human_readable_name(names::Vector{String}; title_case::Bool=false)::Vector{String}
+    converted_names::Vector{String} = map((x) -> replace(x, "_" => " "), names[:])
```
Zapiano, line 17: Another option would be:

`replace.(names[:], "_" => " ")`

I am not sure which one is better, though.

ODE_preferences.jl
```diff
@@ -0,0 +1,11 @@
...
+using OrdinaryDiffEq
+
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileNonStiff" => true)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileStiff" => false)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileAutoSwitch" => false)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileLowStorage" => false)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileDefaultSpecialize" => true)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileAutoSpecialize" => false)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileFunctionWrapperSpecialize" => false)
+set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileNoSpecialize" => false)
```
Zapiano, line 11: Maybe this is one of those situations where we want to be reasonable and not follow the 92 character limit (as an exception). What do you think?

Project.toml
```diff
@@ -104,6 +103,7 @@ NCDatasets = "0.12"
 NamedDims = "1"
 OnlineStats = "1.5"
 OrderedCollections = "1"
+OrdinaryDiffEq = "6"
```
Zapiano, line 106: Just out of curiosity, why did you change this package?

Zapiano:
@ConnectedSystems I only saw that you wrote "For the most part, I have treated changes to update code style to Blue as out of scope for the purpose of this PR." after I did my review. Feel free to ignore and just resolve the comments about style guide if you want.

src/utils/text_display.jl
ConnectedSystems: This is better, I wasn't thinking clearly when making this update

ODE_preferences.jl
ConnectedSystems: Yep, this is run separately anyway to create `LocalPreferences.toml` - only needs to be done once. This script is only kept in case adjustments need to be made.

src/io/ResultSet.jl
ConnectedSystems: It's a minor change, probably won't do much but I think it would reduce initial compile times a little.

Each field "type" specified like so:

`ResultSet{T1,T2,A,B,C,D,G,D1,D2,D3,DF}`

Effectively resolves to `Any` because no other information is given. So the Julia compiler does not create any optimised code at compile time - this is deferred to runtime instead. At the same time, it means the Julia compiler has to be ready to compile combinations of any type. 

Specifying the known types directly helps reduce the number of possible combinations.

It's currently set up like this because I wasn't sure exactly what the structure of `ResultSet` would be, and it's likely to continue changing as our needs/use context evolve.

See: 
1. https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-type
2. https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-containers

Zapiano (approved):

PR #470, opened 2023-09-10
Add option to show GBRMPA management zones

ConnectedSystems:
Add option to mapping to show management zones (defaults to false.

```julia
ADRIA.viz.map(dom; opts=Dict{Symbol,Any}(:show_management_zones=>true))
```

image

Awkward dictionary-based option setting needs to be addressed somehow.
Specifying the types in the dictionary is necessary, otherwise Julia tries to optimize by restricting the Dict only to `Symbol=>Bool` pairs, which causes an issue when other options with different types get set (such as `Symbol=>String`).


Zapiano (approved):
🚀

PR #471, opened 2023-09-11
Update sensitivity module to follow Blue style

ConnectedSystems:
As in title.

### DIFF
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -488,10 +488,13 @@ function outcome_map(
     return outcome_map(rs.inputs, y, rule, names(rs.inputs); S, n_boot, conf)
 end
 
-function _map_outcomes(y::AbstractArray, rule::Union{BitVector,Vector{Int64}})
+function _map_outcomes(
+    y::AbstractArray{<:Real},
+    rule::Union{BitVector,Vector{Int64}}
+)::Union{BitVector,Vector{Int64}}
     return rule
 end
-function _map_outcomes(y::AbstractArray, rule::Function)
+function _map_outcomes(y::AbstractArray{<:Real}, rule::Function)::Vector{Int64}
     _y = col_normalize(y)
     all_p_rule = findall(rule, eachrow(_y))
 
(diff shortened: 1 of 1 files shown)


src/analysis/sensitivity.jl
```diff
@@ -1,9 +1,17 @@
 module sensitivity
 
-using Logging
-using DataFrames, NamedDims, AxisKeys
-using Distributions, HypothesisTests, Bootstrap, StaticArrays, FLoops
-
+using
```
Zapiano, line 3: Blue Style Guide requires one using per line in alphabetical order (https://github.com/invenia/BlueStyle#module-imports)

Zapiano:
@ConnectedSystems  since this is updating this whole file: on line 491, the function `_map_outcomes` accepts a parameter `y` which is not being used AND doesn't specify the returning type. Also: why have a function that returns its argument without doing nothing?

src/analysis/sensitivity.jl
ConnectedSystems: I know this, but I think the function import by theme suggestion is a better fit here ("Avoid splitting a single export over multiple lines; either define one export per line, or group them by theme.").

That said, I did a poor job of grouping them by theme.

Zapiano: 1. Options 1 and 2 below use the same space, so why not the second?

```
# 1
using A
      B
      
# 2
using A
using B
```

2. You can have blocks of `using`'s grouped by theme but inside each block group `using`'s in alphabetical order.

ConnectedSystems:
> @ConnectedSystems since this is updating this whole file: on line 491, the function `_map_outcomes` accepts a parameter `y` which is not being used AND doesn't specify the returning type. Also: why have a function that returns its argument without doing nothing?

It's for flexibility and compatibility.

This line is important:
`all_p_rule = _map_outcomes(y, rule)`

User can pass in a `BitVector` / `Vector{Int64}` instead of a function, thus making `rule` effectively the rule already applied to `y`.

src/analysis/sensitivity.jl
ConnectedSystems: Yes, doing that now! 🤣

ConnectedSystems:
Ready for another look

Zapiano (approved):

PR #475, opened 2023-09-12
Update viz.rules_scatter to accept targeted clusters

### LINKED ISSUE #474: Update rule ADRIA.viz.rules_scatter to accept target_clusters vector instead of clusters
The way `ADRIA.viz.rules_scatter` works now is that is accepts as an argument a `clusters` and an `outcomes` NamedDimsArray and it calls `target_clusters` inside it. This is bad because we may want to target clusters using different algorithms. Instead, now this function should receive only the already targeted clusters vector and let the target to the user.

Zapiano:
As explained here: https://github.com/open-AIMS/ADRIA.jl/issues/474

Closes #474 .


ConnectedSystems (approved):
Approving as is, as I realised all my comments were to do with Blue style.

PR #478, opened 2023-09-13
Rename time_series_clustering and ts_cluster

Zapiano:
Those two functions names were not according to blue style guide and were not very meaningful (in particular `ts_cluster`). We should have functions whose names tell "what they do" (like `cluster_series`) or what they return (like `clustered_scenarios`).

### DIFF
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -91,9 +91,9 @@ function complexity_invariance_distance(
 end
 
 """
-    time_series_clustering(data::AbstractMatrix{T}, n_clusters::Int64)::Vector{Int64} where {T<:Real}
+    cluster_series(data::AbstractMatrix{T}, n_clusters::Int64)::Vector{Int64} where {T<:Real}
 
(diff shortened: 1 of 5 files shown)


src/analysis/clustering.jl
```diff
@@ -91,9 +91,9 @@ function complexity_invariance_distance(
 end
 
 """
-    time_series_clustering(data::AbstractMatrix{T}, n_clusters::Int64)::Vector{Int64} where {T<:Real}
+    cluster_series(data::AbstractMatrix{T}, n_clusters::Int64)::Vector{Int64} where {T<:Real}
 
-Hierarchical clustering between \$S\$ scenarios with \$T\$ time steps each.
+Hierarchical cluster \$S\$ scenarios with \$T\$ time steps each.
```
ConnectedSystems, line 96:
```suggestion
Hierarchically cluster \$S\$ scenarios with \$T\$ time steps each.

```

ConnectedSystems (approved):

Zapiano:
@ConnectedSystems ready for a second review.

PR #479, opened 2023-09-14
Remove locations with no space from consideration for interventions

ConnectedSystems:
In certain cases it seems the MCDA is selecting locations with no available space for seeding deployments.

This PR ensures such locations are removed from consideration.


ConnectedSystems:
@Rosejoycrocker I'm going to merge directly but tagging you here so you are aware of this change.

PR #482, opened 2023-09-14
Update viz.clustered_scenarios to allow confidence interval plot

### LINKED ISSUE #464: Summarized vizualisation for time series clustering
The current version of the function `ADRIA.viz.ts_cluster` plots all series of each cluster:

image

This is not ideal sometimes, so we need a way of only plotting a summarized version, with the median and the confidence intervals only for each cluster.

Zapiano:
- Updated functions _clusters_colors and _cluster_color to _get_colors and _get_alphas
- Added return type to all functions in this file
- Adapted the whole file to the style guide

There is an option to fill or not the area inside the confidence interval - the best visualization option may depend on the situation. The default value of `:fill_confint` is currently set to `true`:

clustered_scenarios_true

This is how it looks with `:fill-confint` set to `false`:

clustered_scenarios_false

Closes #464

### DIFF
--- a/ext/AvizExt/viz/clustering.jl
+++ b/ext/AvizExt/viz/clustering.jl
@@ -56,16 +56,16 @@ function _plot_clusters_series!(
     clusters::Vector{Int64};
     opts::Dict=Dict(),
 )::Nothing
+    unique_clusters = sort(unique(clusters))
     alphas = _get_alphas(clusters)
     clusters_colors = unique(_get_colors(clusters, alphas))
 
     leg_entry::Vector{Any} = [
         series!(ax, data[:, clusters .== cluster]'; solid_color=clusters_colors[idx_c]) for
-        (idx_c, cluster) in enumerate(unique(clusters))
+        (idx_c, cluster) in enumerate(unique_clusters)
     ]
 
-    n_clusters = length(unique(clusters))
-    Legend(g[1, 2], leg_entry, "Cluster " .* string.(1:n_clusters); framevisible=false)
+    Legend(g[1, 2], leg_entry, "Cluster " .* string.(unique_clusters); framevisible=false)
 
     return nothing
 end
@@ -85,13 +85,15 @@ function _plot_clusters_confint!(
         slice_dimension = filter(x -> x != :timesteps, dimnames(data))[1]
     end
 
+    unique_clusters = sort(unique(clusters))
+
     band_colors = unique(_get_colors(clusters, 0.5))
     line_colors = unique(_get_colors(clusters))
-    legend_entry = Vector{Any}(undef, length(unique(clusters)))
+    legend_entry = Vector{Any}(undef, length(unique_clusters))
 
     x_timesteps::UnitRange{Int64} = 1:length(timesteps(data))
 
-    for (idx_c, cluster) in enumerate(unique(clusters))
+    for (idx_c, cluster) in enumerate(unique_clusters)
         cluster_data = data[:, clusters .== cluster]
         data_slices = Slices(cluster_data, NamedDims.dim(cluster_data, slice_dimension))
 
@@ -105,8 +107,9 @@ function _plot_clusters_confint!(
         legend_entry[idx_c] = scatterlines!(ax, y_median; color=line_color, markersize=5)
     end
 
-    n_clusters = length(unique(clusters))
-    Legend(g[1, 2], legend_entry, "Cluster " .* string.(1:n_clusters); framevisible=false)
+    Legend(
+        g[1, 2], legend_entry, "Cluster " .* string.(unique_clusters); framevisible=false
+    )
 
     return nothing
 end
--- a/ext/AvizExt/viz/clustering.jl
+++ b/ext/AvizExt/viz/clustering.jl
@@ -1,4 +1,5 @@
-using JuliennedArrays
+using JuliennedArrays: Slices
+using Statistics
 
 """
     clustered_scenarios(data::AbstractMatrix, clusters::Vector{Int64}; fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
@@ -17,12 +18,13 @@ function ADRIA.viz.clustered_scenarios(
     data::AbstractMatrix,
     clusters::Vector{Int64};
     fig_opts::Dict=Dict(),
-    axis_opts::Dict=Dict()
-)
+    axis_opts::Dict=Dict(),
+    opts::Dict=Dict()
+)::Figure
     f = Figure(; fig_opts...)
     g = f[1, 1] = GridLayout()
 
-    ADRIA.viz.clustered_scenarios!(g, data, clusters; axis_opts=axis_opts)
+    ADRIA.viz.clustered_scenarios!(g, data, clusters; axis_opts=axis_opts, opts=opts)
 
     return f
 end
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
> This is how it looks with :fill-confint set to false:

I'm struggling to see where this is useful - I suggest `:aggregate` or `:summarize` (defaulting to true) which plots either the individual series or the aggregated version with filled confidence intervals.

Incidentally, I'm not against `confint`, but also putting `ci` (or `CI`) as an option as it's an established acronym.

ext/AvizExt/viz/clustering.jl
```diff
@@ -1,4 +1,5 @@
-using JuliennedArrays
+using JuliennedArrays: Slices
+using Statistics
```
ConnectedSystems, line 3: Double space between imports and code please

ext/AvizExt/viz/clustering.jl
```diff
@@ -17,53 +18,104 @@ function ADRIA.viz.clustered_scenarios(
...
+    data::AbstractMatrix,
+    clusters::Vector{Int64};
+    opts::Dict=Dict()
+)::Nothing
+    band_colors = unique(_get_colors(clusters, 0.5))
+    line_colors = unique(_get_colors(clusters))
+    leg_entry = Any[]
+
+    # TODO if data is not a NamedDimsArray this won't work
+    xs::Vector{Float32} = data.timesteps
```
ConnectedSystems, line 92: Could wrap this in a try/catch

ext/AvizExt/viz/clustering.jl
```diff
@@ -17,53 +18,104 @@ function ADRIA.viz.clustered_scenarios(
...
+    clusters::Vector{Int64};
+    opts::Dict=Dict()
+)::Nothing
+    alphas = _get_alphas(clusters)
+    clusters_colors = unique(_get_colors(clusters, alphas))
     leg_entry = Any[]
-    for cluster in unique(clusters)
-        cluster_color = _cluster_color(unique_cluster_colors, clusters, cluster)
+
+    for (idx_c, cluster) in enumerate(unique(clusters))
```
ConnectedSystems, line 63: Trying to get rid of the `Any[]`:

Something like:

```julia
leg_entry = [
    series!(ax, data[:, clusters.==cluster]', solid_color=clusters_colors[idx_c]) 
        for (idx_c, cluster) in enumerate(unique(clusters))
]
```

Not sure what type the vector would be but good to specify that too

ext/AvizExt/viz/clustering.jl
```diff
@@ -17,53 +18,104 @@ function ADRIA.viz.clustered_scenarios(
...
+        else
+            lines!(ax, xs, y_lower, linestyle=:dash, color=band_colors[idx_c])
+            lines!(ax, xs, y_upper, linestyle=:dash, color=band_colors[idx_c])
+        end
+
+        y_median = median.(data_slices)
+        push!(
+            leg_entry,
+            scatterlines!(ax, xs, y_median, color=line_colors[idx_c], markersize=5)
+        )
```
ConnectedSystems, line 112: Same here as previous comment here - would be nice to get rid of the `push!()`

ConnectedSystems (changes requested):
Thanks Pedro, I think some minor tweaks (mostly to get rid of `push!()`) are needed.

Zapiano:
> > This is how it looks with :fill-confint set to false:
> 
> I'm struggling to see where this is useful - I suggest `:aggregate` or `:summarize` (defaulting to true) which plots either the individual series or the aggregated version with filled confidence intervals.

100% agree. 

> Incidentally, I'm not against `confint`, but also putting `ci` (or `CI`) as an option as it's an established acronym.

I know about `ci`, but I really think it's better to avoid acronyms when possible.

Zapiano:
| :summarize = true | :summarize = false |
|------|-------|
|image | clustered_scenarios_summarize-false|

ext/AvizExt/viz/clustering.jl
Zapiano: Instead of using a try/catch, why not change the type of `data` to `NamedDimsArray`? Because that is basically what the try catch will check.

ConnectedSystems: Aaah, yes and also we should use the `timesteps()` function to handle the extraction I think.

I was thinking we want it flexible and applicable to both Matrix and NamedDimsArray, but I think that's not necessary.

Zapiano:

About the `timestep()` function: 
This function is already being used (line 39) when the Axis is created, BUT I have to pass an array to `band!` with x-axis values. You can see that the previous plots doesn't have x-axis labels. I had to use this range (line 84) to fix that. This is the final result:

| :summarize = true | :summarize = false |
|------|-------|
|clustered_scenarios_summarize-true|clustered_scenarios_summarize-false|

ext/AvizExt/viz/clustering.jl
```diff
@@ -14,56 +16,91 @@ Visualize clustered time series of scenarios.
...
     g::Union{GridLayout,GridPosition},
-    data::AbstractMatrix,
+    data::NamedDimsArray,
     clusters::Vector{Int64};
-    axis_opts::Dict=Dict()
-)
-    # Ensure last year is always shown in x-axis
+    axis_opts::Dict=Dict(),
+    opts::Dict=Dict()
+)::Union{GridLayout,GridPosition}
```
ConnectedSystems, line 38: I just noticed this, please standardize order of kwargs to be consistent.

It should be in the order of: `opts`, `fig_opts`, and `axis_opts`

ext/AvizExt/viz/clustering.jl
```diff
@@ -14,56 +16,91 @@ Visualize clustered time series of scenarios.
...
+    ax::Axis,
+    data::NamedDimsArray,
+    clusters::Vector{Int64};
+    opts::Dict=Dict()
+)::Nothing
+    band_colors = unique(_get_colors(clusters, 0.5))
+    line_colors = unique(_get_colors(clusters))
+    legend_entry = Vector{Any}(undef, length(unique(clusters)))
+
+    x_timesteps::UnitRange{Int64} = 1:length(data.timesteps)
```
ConnectedSystems, line 84:
```suggestion
    x_timesteps::UnitRange{Int64} = 1:length(timesteps(data))

```

Reason we'd want to rely on an interface is in case we change the underlying object/type or dimension name.

This might happen as we make the visualizations "smarter" to get plots to show additional info, like we spoke about (display name of metric, unit of measure, etc).

ext/AvizExt/viz/clustering.jl
```diff
@@ -118,52 +155,66 @@ function ADRIA.viz.map!(
...
+function _get_alphas(clusters::Vector{Int64})::Vector{Float64}
+    alphas::Vector{Float64} = zeros(Float64, length(unique(clusters)))
 
-    # Compute line weight
-    color_weight = max(min((1.0 / (n_scens * 0.05)), 0.6), 0.1)
+    for (i, cluster) in enumerate(unique(clusters))
+        n_scens = count(clusters .== cluster)
+        base_alpha = 1.0 / (n_scens * 0.05)
+        alphas[i] = max(min(base_alpha, 0.6), 0.1)
+    end
```
ConnectedSystems, line 212: Thinking about this now, shouldn't the weighting be based on total number of scenarios?
What we have right now could lead to a smaller cluster hiding a larger cluster (unless I'm misinterpreting the code here?)

ConnectedSystems (changes requested):
Some things to think about, but overall looks good!

ext/AvizExt/viz/clustering.jl
Zapiano: I don't think a small cluster would hide a large one because alpha is, at the most, 0.6 (for small clusters) but in this case the cluster has only a few lines.

Considering the result (figure above) it seems to me that this works fine. 

That said, if I would do this now, I would probably try something like:

```julia
base_alpha = 1 - count(clusters .== cluster)/length(clusters)
alphas[i] = max(min(base_alpha, 0.6), 0.1)
```

Zapiano: But I think this change is out of the scope of this PR so if you think we should change this I suggest we open an issue, what do you think?

ConnectedSystems: No need for an issue if you're going to submit a PR directly after (but please do if your not planning to do it immediately).

By hiding, take a look at the example figures provided here. See how the orange trajectories are blocked from view.

https://github.com/open-AIMS/ADRIA.jl/pull/492#issue-1902179842

Maybe we plot smaller clusters last?

Zapiano: You wrote "smaller cluster hiding a larger cluster" but in this figure what I see is a larger cluster hiding a smaller cluster, right? But yes, I think a good improvement would be to plot clusters in order from larger to smaller. 

Let me open a different PR to do that and also try to change the way this alpha is computed and test some different cases.

ConnectedSystems: > You wrote "smaller cluster hiding a larger cluster" but in this figure what I see is a larger cluster hiding a smaller cluster, right?

Yes, correct, accidentally swapped those, sorry for the miscommunication

Zapiano:
@ConnectedSystems 

I just made a small change, so this summary visualization works with other types of timeseries, not only scenarios. This is how it looks for coral cover and different sites (target cluster == 1 and non-target == 2):

|Coral Cover (target high values) | Shelter Volume (target low values) |
|---------------------------------|------------------------------------|
| tac_outcome_series_agg_false | asv_outcome_series_agg_false |
| tac_outcome_series_agg | asv_outcome_series_agg |

ext/AvizExt/viz/clustering.jl
Zapiano: Maybe the problem with the figures you sent from PR 492 wasn't the alpha weight.

If you see my last commit - and the commit message - you will see what happened:

Currently we are iterating over the cluster with a `for cluster in unique(clusters)` but if clusters vector is something like `[2, 1, 1, 3, 2]`, then `unique(clusters)` will return `[2, 1, 3]`, and then cluster number `2`, for instance, will get the weight that should be assigned to cluster `1`, because we are using the index of the cluster inside the vector as it was the number of the cluster itself.

Zapiano: Update: I just tested here and that doesn't solve the visualization problem. I think plotting larger clusters before smaller will solver though.

ConnectedSystems (approved):

ConnectedSystems:
Hi @Zapiano , you don't need to merge in unrelated changes if they aren't relevant to the PR.

Zapiano:
> Hi @Zapiano , you don't need to merge in unrelated changes if they aren't relevant to the PR.

I know... it was by accident

PR #491, opened 2023-09-19
Fix bug in how `sensitivity.outcome_map()` treats categorical variables.

Rosejoycrocker:
In `sensitivity.outcome_map()` factor splittings are determined using the number of bins provided and the function `quantile()` from the Statisitcs.jl package. This is an issue for categorical functions as `quantile()` treats the lowest and highest values in the set of samples provided as the upper and lower bounds of the splitting. This means, with the number of bins being equal to the number of categories, there are duplicate values because there are not enough splittings for the number of categorical variables. Providing the number of bins as (number of categories) +1 does not solve this either because quantile tries to split the range of values into smaller splittings without adjusting the upper and lower bounds. 

The fix provided here checks whether the variable is categorical and if so, sets the number of bins as the (number of categories) + 1. The splittings are then set as [0 , [ordered set of categories]]. The lower bound of zero means that when the bounds checks are applied scenarios are correctly categorised.

Setting this as a draft however because it means that categorical and continuous variables cannot be calculated (or plotted)  together as the size of the storage matrix may not be compatible for both. Do we want to a) keep it this way so the two must be considered separately (and have an error message if requested)? b) make the size of the storage as the maximum number of splittings between categorical factors and the n_bins provided? c) another alternative ?

### DIFF
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -314,8 +314,7 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
             f_name = Symbol(factors[curr])
             f_vals = rs.inputs[:, f_name]
             
-            if f_name .== :guided
-                fv_labels = ["unguided", "cf", last.(split.(string.(ADRIA.methods_mcda),"."))...]
+            if f_name == :guided
                 fv_s = collect(1:length(fv_labels))
             else
                 fv_s = round.(quantile(f_vals, b_slices), digits=2)
@@ -336,6 +335,7 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
                 fv_s,
                 outcomes(factors=f_name, CI=:mean), markersize=15
             )
+
             if f_name == :guided
                 ax.xticks = (fv_s,fv_labels)
                 ax.xticklabelrotation = pi/4
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -310,7 +310,11 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
         for c in 1:n_cols
             f_name = Symbol(factors[curr])
             f_vals = rs.inputs[:, f_name]
-            fv_s = round.(quantile(f_vals, b_slices), digits=2)
+            if all(mod.(f_vals, 1.0) .== 0.0)
+                fv_s = sort(unique(f_vals))
+            else
+                fv_s = round.(quantile(f_vals, b_slices), digits=2)
+            end
 
             ax::Axis = Axis(
                 g[r, c],
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -304,6 +304,9 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
     bin_slices, factor_list, CIs = axiskeys(outcomes)
     b_slices = parse.(Float64, bin_slices)
 
+    if any(f_names .== :guided)
+        fv_labels = ["unguided", "cf", last.(split.(string.(ADRIA.methods_mcda), "."))...]
+    end
     curr::Int64 = 1
     axs = Axis[]
     for r in 1:n_rows
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -310,8 +310,10 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
         for c in 1:n_cols
             f_name = Symbol(factors[curr])
             f_vals = rs.inputs[:, f_name]
-            if all(mod.(f_vals, 1.0) .== 0.0)
-                fv_s = sort(unique(f_vals))
+            
+            if f_name .== :guided
+                fv_labels = ["unguided", "cf", last.(split.(string.(ADRIA.methods_mcda),"."))...]
+                fv_s = collect(1:length(fv_labels))
             else
                 fv_s = round.(quantile(f_vals, b_slices), digits=2)
             end
@@ -331,6 +333,10 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
                 fv_s,
                 outcomes(factors=f_name, CI=:mean), markersize=15
             )
+            if f_name == :guided
+                ax.xticks = (fv_s,fv_labels)
+                ax.xticklabelrotation = pi/4
+            end
 
             push!(axs, ax)
             curr += 1
--- a/src/Domain.jl
+++ b/src/Domain.jl
@@ -152,7 +152,7 @@ function update_params!(d::Domain, params::Union{AbstractVector,DataFrameRow})::
         end
     end
 
-    to_floor = (p_df.ptype .== "integer")
+    to_floor = (p_df.ptype .== "integer") .| (p_df.ptype .== "categorical")
     if any(to_floor)
         p_df[to_floor, :val] .= map_to_discrete.(p_df[to_floor, :val], Int64.(getindex.(p_df[to_floor, :bounds], 2)))
     end
--- a/src/Domain.jl
+++ b/src/Domain.jl
@@ -152,7 +152,7 @@ function update_params!(d::Domain, params::Union{AbstractVector,DataFrameRow})::
         end
     end
 
-    to_floor = (p_df.ptype .== "integer") .| (p_df.ptype .== "categorical")
+    to_floor = _check_discrete(p_df.ptype)
     if any(to_floor)
         p_df[to_floor, :val] .= map_to_discrete.(p_df[to_floor, :val], Int64.(getindex.(p_df[to_floor, :bounds], 2)))
     end
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -459,7 +459,7 @@ function outcome_map(
             fact_idx = foi_spec.fieldname .== fact_t
             lb = foi_spec.lower_bound[fact_idx][1]
             ub = foi_spec.upper_bound[fact_idx][1]
-            X_q .= round.(quantile(lb:1:ub, steps)) .- 1
+            X_q .= round.(quantile(lb:ub, steps)) .- 1
         else
             X_q .= quantile(X[:, fact_t], steps)
         end
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -406,7 +406,7 @@ function outcome_map(
     X::DataFrame,
     y::AbstractVecOrMat{<:Real},
     rule::Union{Function,BitVector,Vector{Int64}},
-    target_factors::Vector,
+    target_factors::Vector{String},
     model_spec::DataFrame;
     S::Int64=20,
     n_boot::Int64=100,
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -412,6 +412,7 @@ function outcome_map(
     n_boot::Int64=100,
     conf::Float64=0.95
 )::NamedDimsArray
+
     if !all([tf in model_spec.fieldname for tf in target_factors])
         missing_factor = .!([tf in model_spec.fieldname for tf in target_factors])
         error("Invalid target factors: $(target_factors[missing_factor])")
@@ -455,9 +456,9 @@ function outcome_map(
     for (j, fact_t) in enumerate(target_factors)
         ptype = model_spec.ptype[model_spec.fieldname .== fact_t][1]
         if ptype == "categorical"
-            fact_indx = foi_spec.fieldname .== fact_t
-            lb = foi_spec.lower_bound[fact_indx][1]
-            ub = foi_spec.upper_bound[fact_indx][1]
+            fact_idx = foi_spec.fieldname .== fact_t
+            lb = foi_spec.lower_bound[fact_idx][1]
+            ub = foi_spec.upper_bound[fact_idx][1]
             X_q .= round.(quantile(lb:1:ub, steps)) .- 1
         else
             X_q .= quantile(X[:, fact_t], steps)
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -406,7 +406,8 @@ function outcome_map(
     X::DataFrame,
     y::AbstractVecOrMat{<:Real},
     rule::Union{Function,BitVector,Vector{Int64}},
-    target_factors::Vector{String};
+    target_factors::Vector,
+    model_spec::DataFrame;
     S::Int64=20,
     n_boot::Int64=100,
     conf::Float64=0.95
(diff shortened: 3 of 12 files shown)


ext/AvizExt/viz/sensitivity.jl
```diff
@@ -310,7 +310,11 @@ function ADRIA.viz.outcome_map!(g::Union{GridLayout,GridPosition}, rs::ResultSet
         for c in 1:n_cols
             f_name = Symbol(factors[curr])
             f_vals = rs.inputs[:, f_name]
-            fv_s = round.(quantile(f_vals, b_slices), digits=2)
+            if all(mod.(f_vals, 1.0) .== 0.0)
+                fv_s = sort(unique(f_vals))
```
ConnectedSystems, line 314: I'm not sure if I fully understand what's being done here - my guess is you're trying to identify factors that only use whole numbers?

If so, the model specification (which is accessible here via `rs.model_spec`) holds information on which factors are categorical/discrete (i.e., `ptype` is integer). Looping through all factor values to check their modulo/modulus is not very efficient for this purpose.

I'm also not sure of the implications for factors that aren't categorical but do use discrete values, such as number of corals to be deployed.

The above aside, when code is a little cryptic best practice is to write a comment explaining the why of the code.

src/analysis/sensitivity.jl
```diff
@@ -330,7 +330,12 @@ function outcome_map(X::DataFrame, y::AbstractVecOrMat{T}, rule::V, target_facto
 
     X_q = zeros(S + 1)
     for (j, fact_t) in enumerate(target_factors)
-        X_q .= quantile(X[:, fact_t], steps)
+        if all(typeof.(X[:, fact_t]) .== Int64)
```
ConnectedSystems, line 333: What if values are whole numbers but expressed as Float? (for whatever reason)

src/analysis/sensitivity.jl
```diff
@@ -330,7 +330,12 @@ function outcome_map(X::DataFrame, y::AbstractVecOrMat{T}, rule::V, target_facto
...
+            X_q .= [0, sort(unique(X[:, fact_t]))...]
```
ConnectedSystems, line 334: Does it matter if values are negative whole numbers, or if values include 0?

ConnectedSystems (commented):
You raise some potential alternatives to consider in the PR description.

I vote for option "c".

My suggestion would be to refactor the visualization methods to consider the model spec. This way we're not relying on awkward checks for integers or modulo operations.

On aligning the results matrix, currently the method skips `missing` - you could mark the unneeded x position as such. 

On a cleaner approach, @Zapiano and I discussed an new approach to handling model results which involves a new data structure that allows metadata to be stored alongside the matrix/array. In effect, a more robust `Metric` type.

If this approach is taken, we could attach metadata to indicate positions in the resulting matrix/array that are relevant for given categorical/discrete factor.

A little more involved but allows analysis of mixed factor types without relying on `missing`.

ext/AvizExt/viz/sensitivity.jl
ConnectedSystems: Does the sorting mess up the ordering between x and y?

Rosejoycrocker:
> You raise some potential alternatives to consider in the PR description.
> 
> I vote for option "c".
> 
> My suggestion would be to refactor the visualization methods to consider the model spec. This way we're not relying on awkward checks for integers or modulo operations. Ideally, for categorical factors you'd replace the x values with their associated names where appropriate (e.g., for `guided` it would be the method name)
> 
> On aligning the results matrix, currently the method skips `missing` - you could mark the unneeded x positions as such.
> 
> On a cleaner approach, @Zapiano and I discussed an new approach to handling model results which involves a new data structure that allows metadata to be stored alongside the matrix/array. In effect, a more robust `Metric` type.
> 
> If this approach is taken, we could attach metadata to indicate positions in the resulting matrix/array that are relevant for given categorical/discrete factor.
> 
> A little more involved but allows analysis of mixed factor types without relying on `missing`.

Hey @ConnectedSystems . By refactor the visualisation methods to include model_spec, do you mean add an argument which allows you to extract the model spec (rs or dom)? Also, note these are changes to the outcome_map sesitivity function as well as the visualisation as both bin the categorical variables incorrectly.

Rosejoycrocker:
Plotting for guided variable now looks like this:

outcome_map_guided_test

Rosejoycrocker:
Plots of other variables for comparison:

outcome_map_seedfoi_test

Rosejoycrocker:
Plot of other categorical use_dist to show plotting also works for this:
outcome_map_use_dist_test

Rosejoycrocker:
Hey @ConnectedSystems, I think this is ready for review

src/io/result_io.jl
```diff
@@ -263,7 +263,9 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
 
     # Store post-processed table of input parameters.
     # +1 skips the RCP column
-    integer_params = findall(domain.model[:ptype] .== "integer")
+    is_cat = domain.model[:ptype] .== "categorical"
+    is_int = domain.model[:ptype] .== "integer"
+    integer_params = findall(is_int .| is_cat)
     map_to_discrete!(scen_spec[:, integer_params.+1], getindex.(domain.model[:bounds], 2)[integer_params])
     inputs[:, :] = Matrix(scen_spec)
```
ConnectedSystems, line 270: Oooo, this needs to be fixed.

The assumption that the `RCP` entry is before any other categorical/discrete factor may not be true.
I'll open a new issue.

src/io/sampling.jl
```diff
@@ -284,7 +284,9 @@ function _deactivate_interventions(to_update::DataFrame)::Nothing
         _row = to_update.fieldname .== c
         _bnds = length(to_update[_row, :bounds][1]) == 2 ? (0.0, 0.0) : (0.0, 0.0, 0.0)
 
-        dval = to_update[_row, :ptype][1] == "integer" ? 0 : 0.0
+        is_int = to_update[_row, :ptype][1] == "integer"
+        is_cat = to_update[_row, :ptype][1] == "categorical"
+        dval = (is_int .| is_cat) ? 0 : 0.0
```
ConnectedSystems, line 289:
```suggestion
        # Handle special cases for discrete valued factors.
        # This will break if meanings/possible values of `ptype` changes.
        is_int = to_update[_row, :ptype][1] == "integer"
        is_cat = to_update[_row, :ptype][1] == "categorical"
        dval = (is_int .| is_cat) ? 0 : 0.0
```

src/Domain.jl
```diff
@@ -152,7 +152,7 @@ function update_params!(d::Domain, params::Union{AbstractVector,DataFrameRow})::
         end
     end
 
-    to_floor = (p_df.ptype .== "integer")
+    to_floor = (p_df.ptype .== "integer") .| (p_df.ptype .== "categorical")
```
ConnectedSystems, line 155: I suggest we define a constant that specifies `ptype`s that are discrete, and write a function that does this check as it is a common task that is repeated.

src/analysis/sensitivity.jl
```diff
@@ -434,8 +453,16 @@ function outcome_map(
 
     X_q = zeros(S + 1)
     for (j, fact_t) in enumerate(target_factors)
-        X_q .= quantile(X[:, fact_t], steps)
-        for (i, s) in enumerate(X_q[1:end-1])
+        ptype = model_spec.ptype[model_spec.fieldname.==fact_t][1]
+        if ptype == "categorical"
+            lb = foi_spec.lower_bound[foi_spec.fieldname.==fact_t][1]
+            ub = foi_spec.upper_bound[foi_spec.fieldname.==fact_t][1]
+            X_q .= round.(quantile(lb:1:ub, steps)) .-1
```
ConnectedSystems, line 460: We're re-checking against `foi_spec.fieldname.==fact_t` here.

Assign to a variable for readability

ConnectedSystems:
```suggestion
        ptype = model_spec.ptype[model_spec.fieldname.==fact_t][1]
        if ptype == "categorical"
            lb = foi_spec.lower_bound[foi_spec.fieldname.==fact_t][1]
            ub = foi_spec.upper_bound[foi_spec.fieldname.==fact_t][1]
            X_q .= round.(quantile(lb:1:ub, steps)) .- 1
```

Missing space.

ConnectedSystems (changes requested):
Some minor changes

src/Domain.jl
Rosejoycrocker: Should this constant/function be in Domain.jl?

src/analysis/sensitivity.jl
```diff
@@ -434,8 +453,17 @@ function outcome_map(
 
     X_q = zeros(S + 1)
     for (j, fact_t) in enumerate(target_factors)
-        X_q .= quantile(X[:, fact_t], steps)
-        for (i, s) in enumerate(X_q[1:end-1])
+        fact_indx = model_spec.fieldname .== fact_t
+        ptype = model_spec.ptype[][1]
```
ConnectedSystems, line 457: What does the empty selector do? I'm guessing it dereferences the entry or something but not sure why...

ConnectedSystems (commented):
One more comment!

src/analysis/sensitivity.jl
Rosejoycrocker: My mistake, I was in the process of replacing `model_spec.fieldname .== fact_t` with a variable because it is used 3 times, but I must have not copied and pasted correctly. Fixed now

Rosejoycrocker:
Hey @ConnectedSystems, I think this is ready for final review

src/io/result_io.jl
```diff
@@ -263,7 +263,7 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
 
     # Store post-processed table of input parameters.
     # +1 skips the RCP column
-    integer_params = findall(domain.model[:ptype] .== "integer")
+    integer_params = findall([_check_discrete(dd) for dd in domain.model[:ptype]])
```
ConnectedSystems, line 266:
```suggestion
    integer_params = findall(_check_discrete.(domain.model[:ptype]))
```

src/io/sampling.jl
```diff
@@ -6,6 +6,8 @@ import Surrogates: sample
 import Surrogates.QuasiMonteCarlo: SobolSample
 
 
+const DISCRETE_FACTOR_TYPES = ["integer", "categorical"]
```
ConnectedSystems, line 9: I think move to `Ecosystem.jl` which defines the abstract `EcoModel` type which all components are a subtype of.

`EcoModel` is a horrible name, we'll change it later.

src/io/sampling.jl
```diff
@@ -369,6 +373,19 @@ function _check_bounds(lower, upper)
...
+"""
+Check ptype for discrete variable types.
+
+Returns true if discrete, false otherwise.
+
+# Arguments
+- `ptype` : String representing variable type.
+"""
+function _check_discrete(p_type::String)
+    check = any(p_type .== DISCRETE_FACTOR_TYPES)
```
ConnectedSystems, line 385: I think the code is readable enough without using an intermediate variable.

```suggestion
    return any(p_type .== DISCRETE_FACTOR_TYPES)
```

ConnectedSystems (commented):
A few more changes, but after those I think this is ready to go.

src/analysis/sensitivity.jl
```diff
@@ -406,11 +406,30 @@ function outcome_map(
...
     rule::Union{Function,BitVector,Vector{Int64}},
-    target_factors::Vector{String};
+    target_factors::Vector{String},
+    model_spec::DataFrame;
     S::Int64=20,
     n_boot::Int64=100,
     conf::Float64=0.95
 )::NamedDimsArray
+    if !all([tf in model_spec.fieldname for tf in target_factors])
+        missing_factor = .!([tf in model_spec.fieldname for tf in target_factors])
```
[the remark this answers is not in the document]
Rosejoycrocker: I think they're different sizes because `missing_factor` is the inverse of what is inside `target_factors` (and so will have the length of target_factors) and `factors_to_assess` is what is inside `model_spec.fieldname` from `target_factors` (and so will have the same size as `model_spec.fieldname`, unless I've misinterpreted? But on checking they seem to be different sizes

Rosejoycrocker:
I've addressed your requested changes, except the one about `factors_to_assess` as it seems they're different sizes

src/analysis/sensitivity.jl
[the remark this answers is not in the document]
ConnectedSystems: Oh I see, the check is to ensure the factors named in `target_factors` are valid or not.

I feel there's a simpler way of doing this but fine to ignore my comment then.

src/analysis/sensitivity.jl
```diff
@@ -434,8 +454,17 @@ function outcome_map(
...
     X_q = zeros(S + 1)
     for (j, fact_t) in enumerate(target_factors)
-        X_q .= quantile(X[:, fact_t], steps)
-        for (i, s) in enumerate(X_q[1:end-1])
+        ptype = model_spec.ptype[model_spec.fieldname .== fact_t][1]
+        if ptype == "categorical"
+            fact_idx = foi_spec.fieldname .== fact_t
+            lb = foi_spec.lower_bound[fact_idx][1]
+            ub = foi_spec.upper_bound[fact_idx][1]
+            X_q .= round.(quantile(lb:1:ub, steps)) .- 1
```
ConnectedSystems, line 462: Is specifying the step value (`:1:`) really necessary? Isn't that the default?

src/ecosystem/Ecosystem.jl
```diff
@@ -7,11 +7,26 @@ using Distributions
...
+const DISCRETE_FACTOR_TYPES = ["integer", "categorical"]
+
+"""
+Check ptype for discrete variable types.
+
+Returns true if discrete, false otherwise.
+
+# Arguments
+- `ptype` : String representing variable type.
+"""
```
ConnectedSystems, line 19: Fix docstring structure

```suggestion
"""
    _check_discrete(p_type::String)::Bool

Check ptype for discrete variable types.
Returns true if discrete, false otherwise.

# Arguments
- `ptype` : String representing variable type.
"""
```

src/ecosystem/Ecosystem.jl
```diff
@@ -7,11 +7,26 @@ using Distributions
...
+function _check_discrete(p_type::String)
```
ConnectedSystems, line 20: Add return type

```suggestion
function _check_discrete(p_type::String)::Bool
```

src/io/sampling.jl
```diff
@@ -284,7 +284,9 @@ function _deactivate_interventions(to_update::DataFrame)::Nothing
         _row = to_update.fieldname .== c
         _bnds = length(to_update[_row, :bounds][1]) == 2 ? (0.0, 0.0) : (0.0, 0.0, 0.0)
 
-        dval = to_update[_row, :ptype][1] == "integer" ? 0 : 0.0
+        # Handle special cases for discrete valued factors.
+        # This will break if meanings/possible values of `ptype` changes.
+        dval = _check_discrete(to_update[_row, :ptype][1]) ? 0 : 0.0
```
ConnectedSystems, line 289: Remove now unnecessary note

```suggestion
        dval = _check_discrete(to_update[_row, :ptype][1]) ? 0 : 0.0
```

ConnectedSystems (changes requested):
Few more changes please

src/analysis/sensitivity.jl
Rosejoycrocker: I don't remember this, thanks for spotting

Rosejoycrocker:
I've addressed the requested changes @ConnectedSystems

src/analysis/sensitivity.jl
```diff
@@ -406,11 +406,31 @@ function outcome_map(
...
     y::AbstractVecOrMat{<:Real},
     rule::Union{Function,BitVector,Vector{Int64}},
-    target_factors::Vector{String};
+    target_factors::Vector{String},
+    model_spec::DataFrame;
     S::Int64=20,
     n_boot::Int64=100,
     conf::Float64=0.95
 )::NamedDimsArray
+
```
ConnectedSystems, line 415: Remove unnecessary space

ConnectedSystems (approved):
One minor change, then good to go I think

PR #463, opened 2023-09-07
Add function to find robust scenarios across several metrics

Zapiano:
Find scenarios that perform well among a list of given metrics.

Close #493

PR #492, opened 2023-09-19
Change target_clusters to accept bool optional param target_highest

### LINKED ISSUE #451: Target clusters with lower values of some metric
Currently, when we select the `target_clusters` it looks for clusters with maximum value of some metric. We need to be able to look at clusters with minimum value as well.

Zapiano:
Allow targeting clusters with lowest value of given metric. Default is targeting highest value.

Results are shown above: cluster 1 is target and 2 is non-target.

| target_highest = true | target_highest = false |
|-----------------------|------------------------|
| target_clusters_max |  target_clusters_min |

Close #451


Zapiano:
@ConnectedSystems I propose we close this PR and Issue#451 as this is already addressed in PR #463

ConnectedSystems:
Very efficient - thanks @Zapiano

PR #496, opened 2023-09-19
Fix _time_labels missing last label

### LINKED ISSUE #495: `_time_labels` gives duplicate labels in some cases
Depending on the labels passed to `_time_labels` it might miss or not the last label, but now we are always adding it - which causes an error because of duplicated labels sometimes.

`tick_pos` gives the tick label positions + 1
`tick_label` finds the corresponding values, but then always adds the last label again

One option is to check to see if the last label is a duplicate before adding.

Zapiano:
Prevent missing last label if (length(labels) - 1) is not multiple of label_step.

Here are some examples plotting `ADRIA.metrics.total_absolute_cover(rs)[timesteps=1:last_year]` for different values of `last_year`. I have clustered scenarios, targeted, and I am plotting using `ADRIA.viz.clustered_scenarios()`:

| last_year = 26 |  last_year = 27 |
|----------------|-----------------|
| tac_outcome_series_26 | tac_outcome_series_27 |

| last_year = 28 | last_year = 29 |
|----------------|----------------|
|tac_outcome_series_28 | tac_outcome_series_29 |

| last_year = 30 | last_year = 31 |
|----------------|----------------|
| tac_outcome_series_30 | tac_outcome_series_31 |

Close #495

### DIFF
--- a/ext/AvizExt/viz/viz.jl
+++ b/ext/AvizExt/viz/viz.jl
@@ -11,7 +11,7 @@ using GLMakie.Colors
 
 Extract time step labels, ensuring last entry is always included.
 """
-function _time_labels(labels; label_step=5)
+function _time_labels(labels; label_step=5)::Tuple{Vector{Int64},Vector{String}}
     labels_length = length(labels)
     labels_strings = string.(labels)
 
--- a/ext/AvizExt/viz/viz.jl
+++ b/ext/AvizExt/viz/viz.jl
@@ -11,9 +11,14 @@ using GLMakie.Colors
 
 Extract time step labels, ensuring last entry is always included.
 """
-function _time_labels(labels)
-    tick_pos = vcat(1:5:length(labels), [length(labels)])
-    tick_label = vcat(string.(labels)[1:5:end], string.(labels)[end])
+function _time_labels(labels; label_step=5)
+    tick_pos = collect(1:label_step:length(labels))
+    tick_label = collect(string.(labels)[1:label_step:end])
+
+    # Prevent missing last label if (length(labels) - 1) is not multiple of label_step
+    if (length(labels) - 1) % label_step != 0
+        return vcat(tick_pos, length(labels)), vcat(tick_label, string.(labels)[end])
+    end
 
     return tick_pos, tick_label
 end
--- a/ext/AvizExt/viz/viz.jl
+++ b/ext/AvizExt/viz/viz.jl
@@ -12,15 +12,18 @@ using GLMakie.Colors
 Extract time step labels, ensuring last entry is always included.
 """
 function _time_labels(labels; label_step=5)
-    tick_pos = collect(1:label_step:length(labels))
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
Thanks for this.

Had a thought: it would be nice to always include the initial time step as well.

I'm okay merging this (after a light review) but just wanted to give you the option of adding the initial time step in this PR rather than making another one.

Zapiano:
> Had a thought: it would be nice to always include the initial time step as well.

But isn't the initial step always included..? In which case it would not be included? Can you give me an example?

ConnectedSystems:
Hmm, that's interesting, must've dreamt what I saw. Please ignore my last comment then

ext/AvizExt/viz/viz.jl
```diff
@@ -11,9 +11,14 @@ using GLMakie.Colors
...
-function _time_labels(labels)
-    tick_pos = vcat(1:5:length(labels), [length(labels)])
-    tick_label = vcat(string.(labels)[1:5:end], string.(labels)[end])
+function _time_labels(labels; label_step=5)
+    tick_pos = collect(1:label_step:length(labels))
+    tick_label = collect(string.(labels)[1:label_step:end])
+
+    # Prevent missing last label if (length(labels) - 1) is not multiple of label_step
+    if (length(labels) - 1) % label_step != 0
+        return vcat(tick_pos, length(labels)), vcat(tick_label, string.(labels)[end])
```
ConnectedSystems, line 20: Extract `string.(labels)` and `length(labels)` to separate variables, just for readability.

ConnectedSystems (changes requested):

Zapiano:

ext/AvizExt/viz/viz.jl
```diff
@@ -11,11 +11,19 @@ using GLMakie.Colors
...
-function _time_labels(labels)
-    tick_pos = vcat(1:5:length(labels), [length(labels)])
-    tick_label = vcat(string.(labels)[1:5:end], string.(labels)[end])
+function _time_labels(labels; label_step=5)
+    labels_length = length(labels)
+    labels_strings = string.(labels)
 
-    return tick_pos, tick_label
+    tick_position = collect(1:label_step:labels_length)
+    tick_label = collect(labels_strings[1:label_step:end])
```
ConnectedSystems, line 19: Are these `collect()` necessary?

ConnectedSystems (approved):
One minor comment to double check, but otherwise ready for merge.

ext/AvizExt/viz/viz.jl
Zapiano: It depends. If I remove them the code still works, but the `typeof` these `1:label_step:labels_length` is `StepRange{Int64, Int64}` and that means that this function would return different types of objects depending on the `(labels_length - 1) % label_step != 0` being `true` or `false`.

So, in my opinion, we should keep these.

Issue #465, opened 2023-09-08
Standardize all the things


ConnectedSystems:
<img src="https://i.imgflip.com/7yez4r.jpg" title="made at imgflip.com" width=200 height=150/></a>

ADRIA expect a standard order of arguments for most of the basic usage functions

For example:

```julia
dom = ADRIA.load_domain("some domain")

scens = ADRIA.sample(dom, 128)

rs = ADRIA.load_results("some result set")
ADRIA.metrics.scenario_total_cover(rs)
ADRIA.metrics.scenario_relative_cover(rs)
```

Note that the domain (`dom`) or result set (`rs`) is expected as the first argument.
Except in cases where it doesn't:

```julia
# domain is no longer the first argument!
ADRIA.run_scenarios(scens, dom, "45")
```

Inconsistent usage patterns like the above are micro-annoyances the user has to account for.
ADRIA API/functions should be standardized to follow a single approach/pattern.

Later on, this standard approach could facilitate piping:

```julia
using ADRIA: sample, run_scenarios
using ADRIA.metrics: scenario_total_cover

rs = sample(dom, 128) |> run_scenarios("45") |> scenario_total_cover
```

---

Names of metric names are also not standardized.

- `scenario_total_cover`
- `scenario_asv` (asv = absolute shelter volume)

There is potential to remove all `scenario`-based metrics and instead apply a macro, e.g:

```julia
@scenario total_absolute_cover(rs)
```

On this subject, the identifiers "total" and "cover" never made much sense to me, particularly as we begin to consider other ecosystems (cover of what? total in what sense?)

The below make more sense and make for a more seamless set of function names:

- `absolute coral cover`
- `relative coral cover`

Units for each could be standardized for conceptual alignment. Some use m^2, but are more readily communicated in km^2 or hectares for example.

---

On visualization:

`ADRIA.viz.scenario(...)`

Should in fact be `ADRIA.viz.scenarios()` as we are plotting a set of scenarios, not a singular trajectory.

There are other inconsistencies across the board.

These should all be addressed.

ConnectedSystems:
- [ ] Deprecate `run_scenarios(df, dom, ...)`
- [ ] Deprecate `viz.scenario()` in favour of `viz.scenarios()`
- [ ] Agree to consistent naming scheme for metrics
- [ ] Standardize metrics

PR #500, opened 2023-09-22
Refactor viz.scenario

Zapiano:
- Rename viz.scenario to viz.scenarios
- Extract some parts of it to external functions
- Adapt to blue style guide
- Other minor refactorings


ConnectedSystems:
Part of #465

src/viz/viz.jl
```diff
@@ -12,8 +12,8 @@ for existing functions/methods.
 function explore() end
 
 # Scenario plotting methods
-function scenario() end
-function scenario!() end
+function scenarios() end
+function scenarios!() end
```
ConnectedSystems, line 16:
```suggestion
function scenarios() end
function scenarios!() end

function scenario(args...; kwargs...)
    @warn """
    viz.scenario() is now deprecated and will be removed in ADRIA v1.0
    
    Instead, use:
        `viz.scenarios()`
    """
    
    return scenarios(args...; kwargs...)
end
```

ConnectedSystems (changes requested):
Add deprecation warning and indication of usage update.

src/viz/viz.jl
Zapiano: Yes, it works.

image

Zapiano:
@ConnectedSystems Ready for another review.

ConnectedSystems (approved):

Issue #510, opened 2023-09-26
Fix bug when loading Domain due to DHWs containing missing values


Rosejoycrocker:
When loading the Moore-2023-08-17 Domain with RCP "45" there is an error due to DHWs not fitting in the Float64 or Float32 type. Expand input type to Union{Float64,Float32,Missing} to amend this.

PR #511, opened 2023-09-26
Fix bug causing Domain loading with RCP to error

Rosejoycrocker:
Error occurs due to incorrect input type for EnvironmentalLayer. This PR adds Missing to possible input types in EnvironmentalLayer

Resolves issue #510 .


ConnectedSystems (approved):
Looks good

Just noting that the error occurs when loading Domain data packages with an RCP specified:

`ADRIA.load_domain("some Domain", "45")`

PR #514, opened 2023-09-28
Remove top_n parameter from usage 

Rosejoycrocker:
top_n was previously used to set the maximum number of sites to sort through to find sufficiently far apart highly ranked sites in site selection. However, this parameter would not be used much and can conflate sensitivity analysis. This PR removes top_n as the benefit of reducing the parameter space is greater than it's necessity in distance sorting.


ConnectedSystems (approved):
LGTM

PR #521, opened 2023-10-04
Add support series and map binary clusters plots

### LINKED ISSUE #494: Properly handle plotting binary (target vs non-target) clusters
The functions that plot cluster series (`clustered_scenarios` and `map`) accept a vector of cluster numbers for each scenario as one of its parameters. For example, when clustering 10 scenarios in 3 clusters, this vector may look like:

```julia
clusters = [1,3,2,3,1,1,2,1,1,3]
```

But this plot functions don't work well when we have clusters numbered 0. That happens when we use `target_clusters`, for example:

```julia
ADRIA.analysis.target_clusters(clusters, s_tac)
# [1,1,0,1,0,0,0,0,1,1,1,1,0,...]
```

These two functions need to be adapted (or we write new ones) to be able to plot clusters both when their indexes are `[1,2,3,...]` or when they are `[0,1]`.

I am marking this as a bug because right now if you try to use these to plot the result of `target_clusters` it raises an error.

Zapiano:
These were partially supported before, but this implements a few improvements like:

- Allow passing a BitVector of clusters
- Move some common functions to theme.jl
- When plotting a BitVector cluster, use "Target" and "Non-Target" as legend labels

Close #494

### DIFF
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -181,7 +181,7 @@ end
 function cluster_labels(
     clusters::Vector{Int64}, data::AbstractVector{<:Real}
 )::Vector{String}
-    cluster_names::Vector{String} = ["Cluster $(cluster)" for cluster in clusters]
+    cluster_names::Vector{String} = "Cluster " .* string.(clusters)
     return cluster_labels(cluster_names, data)
 end
 function cluster_labels(clusters::BitVector, data::AbstractVector{<:Real})::Vector{String}
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -84,20 +84,22 @@ function scenario_colors!(obs_color::Observable, color_map::Vector, scen_types::
     obs_color[] = color_map
 
 """
+    cluster_colors(clusters::Vector{Int64}, unique_colors::Vector{RGBA{Float32}})::Vector{RGBA{Float32}}
     cluster_colors(clusters::Vector{Int64})::Vector{RGBA{Float32}}
+    cluster_colors(clusters::BitVector)::Vector{RGBA{Float32}}
 
-Computes the colors for each clustered scenario
+Gets colors for each clustered scenario.
 
 # Arguments
 - `clusters` : Vector with scenario cluster numbers
 
 # Returns
-Vector with one color for each clustered scenario
+Vector with one color for each clustered scenario.
 """
-function cluster_colors(clusters::Vector{Int64})::Vector{RGBA{Float32}}
-    unique_clusters = unique(clusters)
-    unique_colors = categorical_colors(:seaborn_bright, length(unique_clusters))
-
+function cluster_colors(
+    clusters::Vector{Int64}, unique_colors::Vector{RGBA{Float32}}
+)::Vector{RGBA{Float32}}
+    unique_clusters::Vector{Int64} = sort(unique(clusters))
     colors_map::Dict{Int64,RGBA{Float32}} = Dict(
         c => unique_colors[i] for (i, c) in enumerate(unique_clusters)
     )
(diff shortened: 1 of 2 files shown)


ext/AvizExt/theme.jl
```diff
@@ -82,4 +85,118 @@ function scenario_colors!(obs_color::Observable, color_map::Vector, scen_types::
...
+    return cluster_labels(cluster_names, data)
+end
+function cluster_labels(
+    cluster_names::Vector{String}, data::AbstractVector{<:Real}
+)::Vector{String}
+    legend_labels = Vector{String}(undef, length(unique(cluster_names)))
+    for (idx, name) in enumerate(unique(cluster_names))
+        # TODO Extract to external function and accept as argument
+        cluster_metric = mean(data[name .== cluster_names])
+        metric_formatted = @sprintf "%.1e" cluster_metric
```
ConnectedSystems, line 198: I'm okay with this for now - I think later when we have metrics and arrays with metadata, we can get rid of this and format according to appropriate units (e.g., convert to km^2 or similar)

ext/AvizExt/viz/clustering.jl
```diff
@@ -84,33 +83,38 @@ function _plot_clusters_confint!(
...
-    for (idx_c, cluster) in enumerate(unique_clusters)
-        band_color = band_colors[idx_c]
-        line_color = line_colors[idx_c]
-
-        y_lower, y_median, y_upper = eachcol(
-            ADRIA.analysis.series_confint(
-                data[:, clusters .== cluster]; agg_dim=slice_dimension
-            ),
+    confints = zeros(n_timesteps, length(unique(sorted_clusters)), 3)
+    for (idx_c, cluster) in enumerate(unique(sorted_clusters))
```
ConnectedSystems, line 96: repeated use of `unique(sorted_clusters)`

ext/AvizExt/theme.jl
```diff
@@ -82,4 +85,118 @@ function scenario_colors!(obs_color::Observable, color_map::Vector, scen_types::
...
+end
+
+"""
+    cluster_alphas(clusters::Vector{Int64})::Vector{Float64}
+    cluster_alphas(clusters::BitVector)::Vector{Float64}
+
+Get color alphas for each cluster weighted by number of scenarios.
+
+# Arguments
+- `clusters` : Vector with scenario cluster numbers
```
ConnectedSystems, line 140:
```suggestion
- `clusters` : Vector with scenario cluster ids
```

ext/AvizExt/theme.jl
```diff
@@ -82,4 +85,118 @@ function scenario_colors!(obs_color::Observable, color_map::Vector, scen_types::
...
+function cluster_labels(clusters::Vector{Int64})::Vector{String}
+    return "Cluster " .* string.(unique(clusters))
+end
+function cluster_labels(clusters::BitVector)::Vector{String}
+    return [BINARY_LABELS[cluster] for cluster in unique(clusters)]
+end
+function cluster_labels(
+    clusters::Vector{Int64}, data::AbstractVector{<:Real}
+)::Vector{String}
+    cluster_names::Vector{String} = ["Cluster $(cluster)" for cluster in clusters]
```
ConnectedSystems, line 184:
```suggestion
    cluster_names::Vector{String} = cluster_labels(clusters)
```

or if non-unique elements are allowed:

```julia
cluster_names = "Cluster" .* string.(clusters)
```

ConnectedSystems (changes requested):
Some minor changes for your consideration

ext/AvizExt/theme.jl
Zapiano: Absolutely!

ext/AvizExt/theme.jl
Zapiano: I think `id` is also ambiguous - that's why I avoided it in first place. It could be misinterpreted by the vector ID. What do you think of `tag`, `label` or `key`?

Zapiano: `label` can also be ambiguous because of the axis labels.

Zapiano: That said, I think I like `cluster_labels` more

Zapiano:
@ConnectedSystems done.

ConnectedSystems (approved):
Thanks!

PR #523, opened 2023-10-09
Fix scenarios subsetting now working

### LINKED ISSUE #522: [Bug] Plotting subset of scenarios fail as not all scenario types may be included
Scenario plots break when targeting specific scenario conditions.

The usual approach seems to be broken now:

https://github.com/open-AIMS/ADRIA.jl/blob/2cc3f6f91becdcb9f815b011dd9921879c4c4a7d/ext/AvizExt/viz/scenarios.jl#L236

MWE:

```julia
rs = ADRIA.load_results("...")

target_scens = (rs.inputs.guided .== 0) .& (rs.inputs.SRM .< 1.0)

# Subset using Scenario IDs
s_tac = ADRIA.metrics.scenario_total_cover(rs; scenarios=findall(target_scens))

ADRIA.viz.scenarios(rs, s_tac)
# ERROR: BoundsError: attempt to access 75×271 Matrix{Float64} at index [1:75, 4096-element BitVector]
```

But even if the error above is fixed, you'll run into an issue with this line:
https://github.com/open-AIMS/ADRIA.jl/blob/2cc3f6f91becdcb9f815b011dd9921879c4c4a7d/ext/AvizExt/viz/scenarios.jl#L96C1-L96C1

It expects all scenario types to be present in a given set of scenario outcomes.

MWE:

```julia
dom = ADRIA.load_domain("...")
scens = ADRIA.sample(dom, 4096)
# rs = ADRIA.run_scenarios(dom, scens, "45")
rs = ADRIA.load_results("...")

# Let's look at unguided scenarios with low SRM
target_scens = (rs.inputs.guided .== 0) .& (rs.inputs.SRM .< 1.0)

# Hacky workaround until we support plotting subsets of scenarios again
# Copy original scens as a back up
orig_scens = copy(rs.inputs)

# Filter dataframe down to target
rs.inputs .= filter!(row -> (row.SRM < 1.0) & (row.guided == 0), rs.inputs)

# Get subset of scenario outcomes
s_tac = ADRIA.metrics.scenario_total_cover(rs; scenarios=findall(target_scens))

ADRIA.viz.scenarios(rs, s_tac)
# ERROR: BoundsError: attempt to access 0-element Vector{Tuple{Symbol, Float64}} at index [1]
```

Zapiano:
As in title.

Because this issue involved changing viz.scenarios code, that had a lot code duplicated from viz.clustered_scenarios, also refactored both to re-use the same functions.

Following are some plotting results: 

- scenario_total_cover

| scenario_total_cover :summarized = true | scenario_total_cover :summarized = false|
|--|--|
|scenarios__summarize_true|scenarios__summarize_false|

- scenario_total_cover with the filter `target_scens = (rs.inputs.guided .== 0) .& (rs.inputs.SRM .< 1.0)`

| :summarized = true | :summarized = false|
|--|--|
|scenarios_filtered__summarize_true|scenarios_filtered__summarize_false|

- scenario_total_cover clustered with 4 clusters

| :summarized = true | :summarized = false|
|--|--|
|clusters_summarize_true| clusters_summarize_false|

- scenario_total_cover targeting the best cluster

| :summarized = true | :summarized = false|
|--|--|
|target__summarize_true|target__summarize_false|

Closes #522.

### DIFF
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -192,3 +192,12 @@ function cluster_labels(
     end
     return legend_labels
 end
+
+function alphas(scenario_types::Dict{Symbol,BitVector})::Dict{Symbol,Float64}
+    return Dict(k => alpha(v) for (k, v) in scenario_types)
+end
+
+function alpha(scenario_types::BitVector)::Float64
+    base_alpha = (1 - count(scenario_types) / length(scenario_types))
+    return max(min(base_alpha, 0.3), 0.1)
+end
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -13,71 +13,34 @@ const COLORS::Dict{Symbol,Symbol} = Dict(
 )
 
 const BINARY_LABELS::Dict{Bool,String} = Dict(0 => "Non-Target", 1 => "Target")
-function scenario_type(rs::ResultSet; scenarios=(:))
-    counterfactual = ADRIA.analysis.counterfactual(rs)
-    unguided = ADRIA.analysis.unguided(rs)
-    guided = ADRIA.analysis.guided(rs)
 
-    return (
-        counterfactual=ADRIA.analysis.counterfactual(rs_input)[scenarios],
-        unguided=ADRIA.analysis.unguided(rs_input)[scenarios],
-        guided=ADRIA.analysis.guided(rs_input)[scenarios],
-    )
-function scenario_colors(rs::ResultSet, weight::Float64)::Vector{Tuple{Symbol,Float64}}
-    return [(c, weight) for c in scenario_colors(rs)]
+function colors(scen_groups::Dict{<:Any,BitVector})::Dict{Symbol,Symbol}
+    return Dict(group => COLORS[group] for group in keys(scen_groups))
 end
-function scenario_colors(rs::ResultSet, weight::Float64, hide::BitVector)
-    color_map::Vector{Tuple{Symbol,Float64}} = scenario_colors(rs, weight)
+function colors(
+    scen_groups::Dict{Symbol,BitVector}, weight::Float64
+)::Vector{Tuple{Symbol,Float64}}
+    groups = collect(keys(scen_groups))
+    n_scens = length(scen_groups[groups[1]])
+    _colors = Vector{Symbol}(undef, n_scens)
+    scen_colors = colors(scen_groups)
 
-    if length(hide) > 0
-        color_map[hide] .= ((:white, 0.0),)
+    for (group, scens) in scen_groups
+        _colors[scens] .= [scen_colors[group]]
     end
 
-    return color_map
-end
-function scenario_colors(rs::ResultSet)::Vector{Symbol}
-    return scenario_colors(rs.inputs)
-end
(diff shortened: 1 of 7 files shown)


ConnectedSystems:
Wow, this looks great @Zapiano . I'll have a closer look at the code tomorrow.

@Rosejoycrocker note the unguided and `SRM .< 1.0` case.

Using a more appropriate MCDA method increases the lower bound, so the upper bound is driven entirely by SRM it seems.

MicrosoftTeams-image

src/analysis/scenario.jl
```diff
@@ -0,0 +1,50 @@
+SCENARIO_TYPES = [:counterfactual, :unguided, :guided]
```
ConnectedSystems, line 1:
```suggestion
const SCENARIO_TYPES = [:counterfactual, :unguided, :guided]
```

ext/AvizExt/viz/scenarios.jl
```diff
@@ -1,8 +1,12 @@
 using JuliennedArrays: Slices
+using ADRIA.analysis: series_confint
+
+include("../../../src/utils/text_display.jl")
```
ConnectedSystems, line 4: ☹️

Zapiano: Sorry, that was committed by accident, I'm not even using it...

ext/AvizExt/theme.jl
```diff
@@ -7,69 +7,66 @@ const COLORS::Dict{Symbol,Symbol} = Dict(
...
-    scen_type = scenario_type(rs)
-    counterfactual = scen_type.counterfactual
-    unguided = scen_type.unguided
 
-    color_map[counterfactual] .= ((COLORS[:counterfactual], weight),)
-    color_map[unguided] .= ((COLORS[:unguided], weight),)
-
-    return color_map
+function alphas(scen_groups::Dict{Symbol,BitVector})::Dict{Symbol,Float64}
+    return Dict(name => alpha(scens) for (name, scens) in scen_groups)
```
ConnectedSystems, line 44: I think there's an overuse of dictionaries in this (Dicts are comparably higher in terms of computational cost, as we discussed) but I'm okay with this, at least for now. The effect is probably very low, unless this function is called multiple times in a loop...

ext/AvizExt/viz/scenarios.jl
```diff
@@ -162,80 +200,47 @@ function _plot_scenarios_hist(
...
+            scen_types;
+            by=type -> sum(var(outcomes[:, scenario_types[type]]; dims=2)),
+            rev=true,
+        )
+    elseif by == :size
+        return sort(
+            scen_types; by=type -> size(outcomes[:, scenario_types[type]], 2), rev=true
+        )
+    else
+        throw(ArgumentError("Invalid value of 'by' kwarg"))
```
ConnectedSystems, line 244:
```suggestion
        throw(ArgumentError("Invalid 'by' option. Must be one of: [:variance, :size]"))
```

ConnectedSystems (approved):
Will merge after one minor change

Zapiano:
@ConnectedSystems done.

ConnectedSystems (approved):
Thank you

PR #527, opened 2023-10-10
Add interface function to query number of scenarios in result set

### LINKED ISSUE #517: Implement n_scenarios(rs)
As the name suggests, this function should return the number of scenarios given a result set.

ConnectedSystems:
As in title.

Closes #517

### DIFF
--- a/src/io/ResultSet.jl
+++ b/src/io/ResultSet.jl
@@ -333,6 +333,18 @@ function n_locations(rs::ResultSet)
 end
 
 """
+    n_scenarios(rs::ResultSet)::Int64
+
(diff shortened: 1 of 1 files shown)


Zapiano (commented):
Do you think we might also want a version that accepts a `scenarios` DataFrame?

ConnectedSystems:
> Do you think we might also want a version that accepts a `scenarios` DataFrame?

I think not. It makes sense to make interface functions for our own types. When working with dataframes, we should use dataframe functions.

Zapiano (approved):

PR #537, opened 2023-10-16
Add summarize and turn per_loc and loc_trajectory into alias

Zapiano:
Summarize uses Slices to compute summary metric over some dataset. The same operation was being done inside `per_loc` and `loc_trajectories` and we sometimes need to do similar things but `along` different axis.


ConnectedSystems:
Are `tac_per_location` and `tac_loc_median` scalar values?

If not, could you see what is returned with the broadcasted check?
Double checking this as I'm not sure what happens when you compare NamedDimsArrays like that.

```julia
all(tac_loc_median .== tac_per_location)
```

ConnectedSystems (approved):
LGTM!

PR #540, opened 2023-10-16
Extract call to run scenarios inside tests to external function 

### LINKED ISSUE #513: Refactor tests so we have an available ResultSet instance
What is now run inside `test/example_run.jl` should be run before the other tests so we can write tests that need an instance of ResultSet object. These tests should also check to see if that was run, so it could be run standalone.

Zapiano:
As in title.

Close #513


test/runtests.jl
```diff
@@ -2,64 +2,41 @@ using Test
...
 
-
 const ADRIA_DIR = pkgdir(ADRIA)
 const TEST_DATA_DIR = joinpath(ADRIA_DIR, "test", "data")
 const EXAMPLE_DOMAIN_PATH = joinpath(ADRIA_DIR, "examples", "Example_domain")
 
+function test_rs()
+    orig_loc = pwd()
+    this_loc = @__DIR__
+    cd(this_loc)
```
ConnectedSystems, line 12: This is now bad practice - use the constants defined above to

1. Replace the relative paths (Line 20 - 21) and use `joinpath` with `EXAMPLE_DOMAIN_PATH`
2. Remove all calls to `cd()`

test/growth.jl
```diff
@@ -2,43 +2,66 @@ using Test
 using Distributions
 using ADRIA
 
+@testset "proportional adjustment" begin
```
ConnectedSystems, line 5: I'm assuming this is all just auto-formatting :)

ConnectedSystems (changes requested):
Only one minor change, thank you

Zapiano:
@ConnectedSystems ready for another round.

ConnectedSystems (approved):
Thank you

PR #542, opened 2023-10-17
Resolve Out-of-Memory issues when working with larger spatial areas

ConnectedSystems:
Attempt to reduce high number of memory allocations to a more manageable level 
(recalling that memory allocations are not the same as memory use - although a small reduction is a nice side-effect here).

Much of the effort is in streamlining the `settler_DHW_tolerance!()` function.

High memory use is somewhat expected in GBR-scale simulations (3806 reefs).
Currently, `settler_DHW_tolerance!()` is doing a naive check for any locations with $k$ > 0 - and it turns out all represented reefs have $k$ > 0.

Note: Please ignore changes to `settler_cover()` - I had to cherry pick some changes from another branch (for turning all calculations to be relative to $k$ area) and these were brought over by accident. The changes there don't affect anything.

### DIFF
--- a/src/ExtInterface/ReefMod/Domain.jl
+++ b/src/ExtInterface/ReefMod/Domain.jl
@@ -79,8 +79,8 @@ function load_domain(::Type{ReefModDomain}, fn_path::String, RCP::String)::ReefM
 
     conn_data = load_connectivity(ReefModDomain, data_files, loc_ids)
     in_conn, out_conn, strong_pred = ADRIA.connectivity_strength(
-        conn_data, 
-        vec(site_data.area .* site_data.k), 
+        conn_data,
+        vec(site_data.area .* site_data.k),
         similar(conn_data)
     )
 
@@ -250,7 +250,7 @@ function load_connectivity(::Type{ReefModDomain}, data_path::String, loc_ids::Ve
     end
 
     # Mean over all years
-    conn_data::SparseMatrixCSC = sparse(dropdims(mean(tmp_mat, dims=3), dims=3))
+    conn_data::Matrix{Float64} = dropdims(mean(tmp_mat, dims=3), dims=3)
     return NamedDimsArray(conn_data, source=loc_ids, sinks=loc_ids)
 end
 
--- a/src/ExtInterface/ReefMod/Domain.jl
+++ b/src/ExtInterface/ReefMod/Domain.jl
@@ -1,5 +1,6 @@
 using NamedDims, AxisKeys, CSV
-using CSV, DataFrames, Statistics, Distributions
+using DataFrames, Statistics, Distributions
+using SparseArrays
 import GeoDataFrames as GDF
 
 using ModelParameters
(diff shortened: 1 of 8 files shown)


Zapiano:
@ConnectedSystems are you still going to add more things to this PR?

ConnectedSystems:
> @ConnectedSystems are you still going to add more things to this PR?

Yes, I'll work on the changes we discussed earlier today.

ConnectedSystems:
Ready for first review I think @Zapiano

ConnectedSystems:
Note, I will be reverting back to dense matrices from sparse matrices.

While the reduction of memory use from using a sparse matrix for connectivity data is initially beneficial, Julia has to implicitly convert to dense matrices to do some of the calculations resulting in more than necessary memory allocations. So the benefits of having this dataset in a sparse matrix is outweighed by the conversion overheads.

ConnectedSystems:
Actually, please hold off on the review, I need to double check a few things

ConnectedSystems:
Ready for review @Zapiano.

```
32.362324 seconds (45.52 M allocations: 13.177 GiB, 11.05% gc time)
```

There are many other improvements in the `consistent-cover` branch that should reduce these timings further, but the changes in this PR addresses the key issue of large memory allocations for large spatial contexts.

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -1,5 +1,6 @@
 using NamedDims, AxisKeys, CSV
-using CSV, DataFrames, Statistics, Distributions
+using DataFrames, Statistics, Distributions
+using SparseArrays
```
Zapiano, line 3: I remember you said you were not going to use SparseArrays anymore. Should this line be removed then?

src/ecosystem/connectivity.jl
```diff
@@ -174,11 +174,14 @@ function connectivity_strength(TP_base::AbstractMatrix{Float64})::NamedTuple
...
+    area_weighted_TP::AbstractMatrix{Float64},
+    cover::Vector{Float64},
+    TP_cache::AbstractMatrix{Float64}
 )::NamedTuple
+
     # Accounts for cases where there is no coral cover
-    tp = (area_weighted_TP .* cover)
-    tp .= tp ./ maximum(tp)
+    TP_cache .= (area_weighted_TP .* cover)
+    TP_cache .= TP_cache ./ maximum(TP_cache)
```
Zapiano, line 184: Just checking: is it possible that maximum(TP_cache) is 0? Because in this case this will be NaN.

Zapiano (approved):
Just two questions.

src/ecosystem/connectivity.jl
ConnectedSystems: Hypothetically? Yes.

Realistically? No, there should always be some connectivity.

src/ExtInterface/ReefMod/Domain.jl
ConnectedSystems: Thanks for the catch

PR #547, opened 2023-10-18
Changes to mcda functions to reflect coral cover being in terms of k area

Rosejoycrocker:
Adds changes to mcda to reflect changes to coral cover being relative to k area. This includes cover and space left criteria being relative to 1 and only k area being required for input to the shading matrix. Changes to cover in connectivity are dealt with outside of the mcda functions and hence don't require changing.

I had to rebase with consistent cover and it has introduced some other changes so apologies for this.


ConnectedSystems:
@Rosejoycrocker again, in future please branch off the relevant branch and use rebase instead of merge. This PR has a number of duplicate commits unrelated to the changes you worked on.

Rosejoycrocker:
> @Rosejoycrocker again, in future please branch off the relevant branch and use rebase instead of merge. This PR has a number of duplicate commits unrelated to the changes you worked on.

Hi @ConnectedSystems, I just realised I think this occured because I rebased with main before rebasing with the consistent-cover branch. I'm happy to start another branch off of consistent cover and cherry pick my changes (there's not many) if that is easier.

ConnectedSystems:
I already did a rebase.

src/decision/dMCDA.jl
```diff
@@ -771,7 +800,15 @@ Here, `max_cover` represents the max. carrying capacity for each site (the `k` v
 - `available_space` : vector/matrix : space available at each site (`k` value)
 - `depth` : vector of site ids found to be within desired depth range
 """
-function unguided_site_selection(prefseedsites, prefshadesites, seed_years, shade_years, n_site_int, available_space, depth)
+function unguided_site_selection(
+    prefseedsites,
+    prefshadesites,
```
ConnectedSystems, line 805: Missing docstring for `prefshadesites`.

While you're here, you might as well standardize variable names to snake case.

ConnectedSystems:
Hi Rose,

Did you try running any simulations? I'm seeing instability warnings from the ODE which were not there before. 

From past experience, this is usually caused by mismatches in how area is handled, leading to inappropriate values being used as input to the ODE.

In `create_shade_matrix()` circa L 246:

`SH[:, 8] = (max_area .- A[:, 8])  # total area of coral cover`

When I step into the function and look at what the values are:

```julia
infil> max_area[1:5]'
1×5 adjoint(::Vector{Float64}) with eltype Float64:
 31629.0  13095.0  19859.9  21562.4  15133.7

infil> A[:, 8][1:5]'
1×5 adjoint(::Vector{Float64}) with eltype Float64:
 70109.1  39316.9  72367.2  47753.2  1.30353e5

infil> all((max_area .- A[:, 8]) .< 0.0)
true
```

So clearly this all leads to negative values and later on, negative values are set to 0.

Probably not the issue causing instabilities, but it's an issue...

Is `max_area` correct as well? Doesn't seem to be the maximum area?

Rosejoycrocker:
> Hi Rose,
> 
> Did you try running any simulations? I'm seeing instability warnings from the ODE which were not there before.
> 
> From past experience, this is usually caused by mismatches in how area is handled, leading to inappropriate values being used as input to the ODE.
> 
> In `create_shade_matrix()` circa L 246:
> 
> `SH[:, 8] = (max_area .- A[:, 8]) # total area of coral cover`
> 
> When I step into the function and look at what the values are:
> 
> ```julia
> infil> max_area[1:5]'
> 1×5 adjoint(::Vector{Float64}) with eltype Float64:
>  31629.0  13095.0  19859.9  21562.4  15133.7
> 
> infil> A[:, 8][1:5]'
> 1×5 adjoint(::Vector{Float64}) with eltype Float64:
>  70109.1  39316.9  72367.2  47753.2  1.30353e5
> 
> infil> all((max_area .- A[:, 8]) .< 0.0)
> true
> ```
> 
> So clearly this all leads to negative values and later on, negative values are set to 0.
> 
> Probably not the issue causing instabilities, but it's an issue...
> 
> Is `max_area` correct as well? Doesn't seem to be the maximum area?

I'll check it. It's supposed to be the total area of coral, so on reflection it should probably be defined as this in A and then the calculation of space available happen in the seeding matrix creation rather than vice versa.

src/decision/dMCDA.jl
```diff
@@ -751,27 +806,45 @@ function site_selection(domain::Domain, scenario::DataFrameRow, w_scens::NamedDi
     out_conn = out_conn[site_ids]
     strong_pred = strong_pred[site_ids]
 
-    (_, _, ranks) = guided_site_selection(mcda_vars, scenario.guided, true, true, prefseedsites, prefshadesites, rankingsin, in_conn, out_conn, strong_pred)
+    (_, _, ranks) = guided_site_selection(
+        mcda_vars,
+        scenario.guided,
+        true,
+        true,
```
ConnectedSystems, line 813: Avoid magic values - assign these to variables that at least indicate why they are true (or false)

src/decision/dMCDA.jl
```diff
@@ -412,40 +432,41 @@ function create_shade_matrix(A, max_area, wt_conn_shade, wt_waves, wt_heat, wt_p
     wsh .= mcda_normalize(wsh)
 
     SH[:, 4] = (1.0 .- A[:, 4]) # complimentary of wave damage risk
-    SH[:, 8] = (max_area .- A[:, 8]) # total area of coral cover
+    SH[:, 8] = A[:, 8] # total area of coral cover
```
ConnectedSystems, line 435: `SH` is already a copy of `A`? Wouldn't it already have this data?

src/decision/dMCDA.jl
```diff
@@ -412,40 +432,41 @@ function create_shade_matrix(A, max_area, wt_conn_shade, wt_waves, wt_heat, wt_p
     wsh .= mcda_normalize(wsh)
```
ConnectedSystems, line 433: What was the reasoning behind the normalization again?

It was removed for seeding.

Rosejoycrocker: The set of weights should be normalised. It looks like you removed the seeding weight normalisation 6 months ago because of duplication, but maybe the original was also removed. I just checked JMcDM though and it looks like JMcDM does normalisation for weights and matrices internally also, so we can get rid of matrix and weights normalisation.

Rosejoycrocker: Sorry, correction- we can remove normalisation when we use JMcDM methods. So I'll move it up to only be used if a non-JMcDM method is selected.

src/decision/dMCDA.jl
```diff
@@ -289,8 +296,8 @@ function create_decision_matrix(site_ids, in_conn, out_conn, sum_cover, max_cove
     # priority zone predecessors and sites
     A[:, 7] .= zones_criteria
 
-    # Proportion of empty space (no coral) compared to max possible cover
-    A[:, 8] = max.((max_cover - sum_cover), 0.0) .* area
+    # Area of coral cover in m^2
+    A[:, 8] = sum_cover .* k_area
```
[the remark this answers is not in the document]
Rosejoycrocker: `sum_cover` should be the proportional cover (so between 0 and 1) and `k_area` should be in m^2 so the result will be in m^2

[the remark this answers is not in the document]
ConnectedSystems: Would `prop_cover` be more appropriate then?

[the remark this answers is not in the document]
ConnectedSystems: I don't mind either way, I'm just trying to reduce possible confusion.

Could add a comment?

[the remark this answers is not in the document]
Rosejoycrocker: Yeah `prop_cover` probably makes more sense, I'll change the name and add a comment

src/decision/dMCDA.jl
```diff
@@ -447,8 +447,8 @@ function create_shade_matrix(
         wt_hi_cover,
     ]
 
-    SH[:, 4] = (1.0 .- A[:, 4]) # complimentary of wave damage risk
-    SH[:, 8] = A[:, 8] # total area of coral cover
+    SH[:, 4] = (1.0 .- SH[:, 4]) # complimentary of wave damage risk
+    SH[:, 8] = SH[:, 8] # total area of coral cover
```
ConnectedSystems, line 473: `SH[:, 8] = SH[:, 8] # total area of coral cover`

Want to think about this one for a little bit? 😝

ConnectedSystems: two spaces after code for inline comments please

Rosejoycrocker: whoops didn't even see this :D

ConnectedSystems (commented):
Some things to check

src/decision/location_selection.jl
```diff
@@ -28,7 +28,7 @@ function _location_selection(domain::Domain, mcda_vars::DMCDA_vars, guided::Int6
     # Determine connectivity strength
     # Account for cases where no coral cover
     in_conn, out_conn, strong_pred = connectivity_strength(
-        domain.TP_data .* site_k_area(domain), collect(mcda_vars.sum_cover)
+        domain.TP_data .* site_k_area(domain), collect(mcda_vars.prop_cover)
```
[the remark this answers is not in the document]
Rosejoycrocker: The collect is because the cover is stored in the domain as a NamedDimsArray. For now I've moved the collect outside into rank_locations(), but open to other ways of doing this

Rosejoycrocker:
Hey @ConnectedSystems, I believe this is ready for review. Let me know whether you think sum_cover should be input as an array instead or converted to one elsewhere. I changed things around a bit as it seems available space is the main quantity calculated in scenario.jl, so now the main inputs are k_area and leftover_space for the mcda.

.gitignore
```diff
@@ -12,3 +12,5 @@ examples/Outputs/
 JuliaSysimage.dll
 
 docs/build/
+
+/test/jl*
```
ConnectedSystems, line 16: This will ignore any file starting with `jl` in a directory called `test` at the root of the file system. I don't think this works for Windows but I'm also not sure what you're trying to ignore here?

Rosejoycrocker: When I run tests, they generate a large number of files starting with jl in the test folder. I assumed this was to do with the new tests for running scenarios

src/decision/dMCDA.jl
```diff
@@ -469,49 +482,61 @@ Tuple (SH, wsh)
...
     SH = copy(A)
 
     # Remove consideration of site depth as shading not accounted for in bleaching model yet
-    SH = SH[:, 1:end-1]
+    SH = SH[:, 1:(end - 1)]
 
-    wsh = [wt_conn_shade, wt_conn_shade, wt_waves, wt_heat, wt_predec_shade, wt_predec_zones_shade, wt_hi_cover]
-    wsh .= mcda_normalize(wsh)
+    # Get area of coral cover (m²) by subtracting leftover space from total area for coral.
+    SH[:, 8] = k_area .- SH[:, 8]
```
ConnectedSystems, line 512: Looking at this again, why not just pass in the absolute coral cover (which is already calculated elsewhere) rather than redoing the calculation again?

.gitignore
ConnectedSystems: Yes, but don't they get auto-cleared?

I think my point above still stands though, `/test` points to the root of the file system, it's not local to the project directory and on Windows systems it's a nonsensical file path (I think) so gitignore is ignoring a non-existent path. 

See `docs/build/` above this entry - there's no leading slash.

Rosejoycrocker: How do you get them to auto-clear?

src/decision/dMCDA.jl
Rosejoycrocker: What is absolute cover called in scenario now? I can see `loc_k_area`, which is the k area, `C_t`, `C_Cover` and `loc_coral_cover `which are all forms of relative cover. I tried searching for variables with `m\^2` , or `abs` in them but couldn't find anything so I used `leftover_space_m2` instead

.gitignore
ConnectedSystems: It doesn't create files starting with `jl`, but it does create a folder with `jl` - you were correct in that these are temporary directories to store the example run that occurs as part of a test.

The directory should be automatically cleared/deleted after tests finish running (and this happens for me, I just checked).

The failing test is new though, doesn't happen on `Main`.

ConnectedSystems: The failing test also doesn't happen in `consistent-cover`, but a different error to do with `proportional_adjustment!()` does happen 🤔 

```julia
proportional adjustment: Error During Test at C:\Users\tiwanaga\projects\ADRIA.jl\test\growth.jl:5
  Got exception outside of a @test
  MethodError: no method matching proportional_adjustment!(::Matrix{Float64}, ::Vector{Float64}, ::Vector{Float64})

  Closest candidates are:
    proportional_adjustment!(::Union{SubArray{T}, Matrix{T}}, ::Vector{T}) where T<:Float64
     @ ADRIA C:\Users\tiwanaga\projects\ADRIA.jl\src\ecosystem\corals\growth.jl:43
    proportional_adjustment!(::Union{SubArray{T}, Matrix{T}}) where T<:Float64
     @ ADRIA C:\Users\tiwanaga\projects\ADRIA.jl\src\ecosystem\corals\growth.jl:66
```

src/decision/dMCDA.jl
ConnectedSystems: Ah okay, so we don't have it already calculated then. Okay to ignore this comment then, thank you.

.gitignore
Rosejoycrocker: hmm weird I'll remove the line and try rebasing and see if the test still fails

Rosejoycrocker: ok I think maybe something weird happened when I rebased with consistent-cover the first time, I copied and pasted the growth.jl test from consistent-cover and now I'm only getting the proportional adjustment error.

ConnectedSystems: Did you compare this revision with the previous one to see what the difference was and what was causing the issue?

Rosejoycrocker: Yeah so I just looked into this further. I don't think it's a rebase issue now, the ODE test error is still there if I comment out the proportional cover test. The proportional cover test error is intoduced because tmp is added as an additional argument in `proportional_adjustment!` on line 11 of the growth.jl test in consistent-cover. If I remove the `tmp` argument the test passes (which is how it was in my branch before I compared with consistent-cover).

Rosejoycrocker: I'll look into a fix for the the ODE issue myself (as you suggested just now on #551.

Rosejoycrocker:
All tests passing now. There was an additional issue where in the no recruitment growth ode test we were getting negative covers hich has been fixed now.

src/decision/dMCDA.jl
```diff
@@ -412,22 +415,31 @@ function create_seed_matrix(
...
+    SE[:, 5] = (1 .- SE[:, 5])  # compliment of heat risk
 
     # Coral real estate as total area, sites with ≤ min_area to be seeded available filtered out
-    SE[vec(A[:, 8] .<= min_area), 8] .= NaN
+    # This will also filter out sites with 0 space
+    SE[SE[:, 8] .<= min_area, 8] .= NaN
 
     # Mark "hot" locations for filter
-    SE[vec(A[:, 5] .>= 0.75), 5] .= NaN
+    #SE[vec(A[:, 5] .>= 0.75), 5] .= NaN
```
ConnectedSystems, line 442: In terms of impact, this change is very big and unrelated to the purpose of the PR.

Could you shift this to a different PR please.

test/growth.jl
```diff
@@ -8,7 +8,7 @@ using ADRIA
     max_cover = rand(20)
 
     for i in axes(Y, 1)
-        ADRIA.proportional_adjustment!(Y[i, :, :], tmp, max_cover)
+        ADRIA.proportional_adjustment!(Y[i, :, :], max_cover)
```
ConnectedSystems, line 11: I don't know what happened here, but the original was correct. It looks like the changes were left out during an incorrect merge/rebase.

ConnectedSystems (approved):
I'm going to approve and merge into `consistent-cover`. We can fix all errors there instead.

PR #552, opened 2023-10-27
Add ability to perform convergence analysis

Rosejoycrocker:
Adds the function `convergence' which divides a set of scenarios into randomised sets of increasing size (non-overlapping, where the number of sets is an input) and calculates the sensitivity.

Also adds corresponding viz functions `ADRIA.viz.convergence!` and `ADRIA.viz.convergence` which plot the sensitivity of specified factors for an increasing number of scenarios. Plotting includes the option to plot multiple factors overlayed in a single figure or as a grid.


src/analysis/sensitivity.jl
```diff
@@ -196,6 +201,80 @@ function pawn(
     return pawn(rs.inputs, y; S=S)
 end
 
+"""
+    convergence(X::DataFrame, y::NamedDimsArray, target_factors::Vector{Symbol}; n_steps::Int64=10)
+
+    Calculates the PAWN sensitivity index for an increasing number of scenarios where the maximum
+        is the total number of scenarios in scens. Number of scenario subsets determined by N_steps.
```
ConnectedSystems, line 208: Fix indentation levels

src/analysis/sensitivity.jl
```diff
@@ -196,6 +201,80 @@ function pawn(
     return pawn(rs.inputs, y; S=S)
 end
 
+"""
+    convergence(X::DataFrame, y::NamedDimsArray, target_factors::Vector{Symbol}; n_steps::Int64=10)
```
ConnectedSystems, line 205: Add return type to both documented and implemented function signature please

src/analysis/sensitivity.jl
```diff
@@ -196,6 +201,80 @@ function pawn(
...
+        Array{Float64}(zeros(length(target_factors), 8, length(N_it)));
+        factors=target_factors,
+        Si=[:min, :lb, :mean, :median, :ub, :max, :std, :cv],
+        n_scenarios=N_it,
+    )
+    scens_idx = randperm(N)
+
+    for nn in N_it
+        pawn_store(; n_scenarios=nn) .= col_normalize(
+            pawn(X[scens_idx[1:nn], :], Array(y[scens_idx[1:nn]])),
```
ConnectedSystems, line 240: Could the method be passed in as an argument so we can use any sensitivity analysis method in the form of `f(X, y) -> Si`?

src/analysis/sensitivity.jl
```diff
@@ -196,6 +201,80 @@ function pawn(
...
+
+# Arguments
+- `X` : Model inputs
+- `y` : Model outputs
+- `target_factors` : Names of target factors represented by columns in `X`
+- `n_steps` : Number of steps to cut the total number of scenarios into.
+
+# Returns
+NamedDimsArray, of min, mean, median, max, std, and cv summary statistics for an increasing
+number of scenarios.
```
ConnectedSystems, line 218: Update description to include lower/upper bound

src/analysis/sensitivity.jl
Rosejoycrocker: I can do this, I'll just have to add lower and upper quantiles as outputs for the other methods

Rosejoycrocker: Hi @ConnectedSystems , is the plan to allow only other "single output" metrics like pawn, or also methods like outcome_map and rsa which have different values for different ranges? If we want the latter I'll have to generalise the storage and some other things so let me know.

ConnectedSystems: No just methods that return a scalar (or set of scalars), thank you

Rosejoycrocker:
I think this is ready for review again @ConnectedSystems

docs/src/usage/analysis.md
```diff
@@ -200,6 +200,41 @@ save("tsa.png", tsa_fig)
...
+
+```julia
+
+outcome = ADRIA.mean(ADRIA.metrics.scenario_total_cover(rs); dims=:timesteps)[timesteps=1]
+
+# Plot heat map of sensitivities for increasing number of scenarios for 
+# individual factors (foi).
+foi = [:dhw_scenario, :wave_scenario, :guided]
+Si_N = convergence(scens, outcome, foi)
+ADRIA.viz.convergence(Si_N, Symbol.(foi))
```
ConnectedSystems, line 217: No indication of where `convergence` comes from?

Is it not within a submodule?
Did you make it so it is not within a submodule and so it leaks into the global namespace?

docs/src/usage/analysis.md
```diff
@@ -200,6 +200,41 @@ save("tsa.png", tsa_fig)
...
 
+### Convergence Analysis for Sensitivity
+
+When undertaking sensitivity analysis it is importnant to have a sufficient sample size that the sensitivity measure has converged and is relatively stable. To understand convergence behaviour a 
+large sample can be drawn and split into smaller sub-samples. Plotting a sensitivity measure against increasing sample size can help to understand whether the sample size is sufficient for convergence.
+The function `convergence` can be used to calculated a sensitivity measure for an increasing number of samples. The result can then be plotted as band plots or a heat map using `viz.convergence`.
+
+```julia
+
+outcome = ADRIA.mean(ADRIA.metrics.scenario_total_cover(rs); dims=:timesteps)[timesteps=1]
```
ConnectedSystems, line 211: Why `ADRIA.mean()` ? User should load the Statistics package themselves.

docs/src/usage/analysis.md
```diff
@@ -200,6 +200,41 @@ save("tsa.png", tsa_fig)
 ![Plots of Temporal Sensitivities](/ADRIA.jl/dev/assets/imgs/tsa.png?raw=true "Temporal Sensitivity Analysis")
 
 
+### Convergence Analysis for Sensitivity
+
+When undertaking sensitivity analysis it is importnant to have a sufficient sample size that the sensitivity measure has converged and is relatively stable. To understand convergence behaviour a 
+large sample can be drawn and split into smaller sub-samples. Plotting a sensitivity measure against increasing sample size can help to understand whether the sample size is sufficient for convergence.
+The function `convergence` can be used to calculated a sensitivity measure for an increasing number of samples. The result can then be plotted as band plots or a heat map using `viz.convergence`.
```
ConnectedSystems, line 207:
```suggestion
When undertaking sensitivity analysis it is important to have sufficient number of samples such that the sensitivity measure converges to a stable state. To assess whether sufficient samples have been taken a convergence analysis can be conducted. One approach is to draw a large sample and then iteratively assess stability of the sensitivity metric using increasingly number of sub-samples. The sensitivity metric is described as having "converged" if there is little to no fluctuations/variance for a given sample size. The analysis can help determine if too little (or too many) samples have taken for the purpose of sensitivity analysis.

The function `convergence` can be used to calculated a sensitivity measure for an increasing number of samples. The result can then be plotted as band plots or a heat map using `viz.convergence`.
```

ConnectedSystems, line 207: You say `convergence` and later say `viz.convergence` to differentiate between the analysis and visualization methods.

Should the first be `sensitivity.convergence` or `analysis.convergence` instead?

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -244,6 +273,199 @@ function ADRIA.viz.rsa(rs::ResultSet, si::NamedDimsArray, factors::Vector{String
...
+- `axis_opts` : Additional options to pass to adjust Axis attributes
+  See: https://docs.makie.org/v0.19/api/index.html#Axis
+
+# Returns
+GLMakie figure
+"""
+function ADRIA.viz.convergence!(
+    g::GridPosition,
+    Si_conv::NamedDimsArray,
+    factors::Vector{Symbol},
```
ConnectedSystems, line 307: How does this differentiate between factors and groups of factors related to a specific component?
I don't see how the usage in the docs works.

I'm not sure if its clear to you, but any method specific option should be in `opts`, not as a function argument.
Specifically, `plot_overlay` should be passed in as `opts=Dict(:plot_overlay=>true)`

I think `Dict(:viz_type=>:heatmap)` and `Dict(:viz_type=>:series)` is more explicit.

I would use a barrier function and separate out the functionality according to whether its a heatmap or a series plot.

```
function _series_convergence()
    ...
end

function _heatmap_convergence()
    ...
end

function ADRIA.viz.convergence!(
    ....
    opts=Dict(:viz_type=>:series)
)
    viz_type = get(opts, :viz_type, :series)
    
    try
        func = eval(Symbol("_$(viz_type)_convergence"))
    catch err
        if !(err isa UndefVarError)
            rethrow(err)
        end

        error("Convergence plot type $(viz_type) is not supported")
    end

    return func(pass_in_relevant_args_here)
end
```

Adding support for other kinds of analyses would then just be a matter of adding more private functions with the pattern `_*_convergence()`, where `*` is the name of the approach.

ConnectedSystems (changes requested):
I may have other comments...

docs/src/usage/analysis.md
Rosejoycrocker: It's inside the ADRIA.sensitivity submodule. I'll add the .sensitivity at the front. I think in the example I copied from I imported the whole model so I didn't need to.

ext/AvizExt/viz/sensitivity.jl
Rosejoycrocker: Hey @ConnectedSystems, thanks for the suggestion. I've made changes to implement this but am having trouble with the scope of eval(). It throws an undefined error due to function inputs (g, factors etc.) not being in scope. I tried to avoid this by switching to `func = getfield(@__MODULE__, Symbol("_$(viz_type)_convergence"))`, but then it evaluates the function fine (as in inputs are in scope) but throws an error on return because func is undefined outside of the `try-catch` block. It works if I do `global func = getfield(@__MODULE__, Symbol("_$(viz_type)_convergence"))` but this is a hacky fix. Any idea how to solve this?

Rosejoycrocker: Thanks, I think I'll switch to `if/elseif/else` for now because I can't think of another plot type case. If we add one in future I can always change it.

Rosejoycrocker:
I've addressed your most recent comments @ConnectedSystems

docs/src/usage/analysis.md
```diff
@@ -200,6 +200,42 @@ save("tsa.png", tsa_fig)
...
+![Plots of Convergence](/ADRIA.jl/dev/assets/imgs/colormap_convergence_components.png?raw=true "Convergence Analysis- model components")
+
+```julia
+# Plot covergence as grid of band plots
+Si_N = convergence(scens, outcome, foi)
+ADRIA.viz.convergence(Si_N, foi, plot_overlay=false)
+
+# Plot covergence as grid of overlayed band plots
+ADRIA.viz.convergence(Si_N, foi, plot_overlay=true)
+```
```
ConnectedSystems, line 237: `plot_overlay` not a valid argument anymore?

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -235,14 +257,253 @@ function ADRIA.viz.rsa!(g::Union{GridLayout,GridPosition}, rs::ResultSet, si::Na
...
+                Si_conv(; Si=:median)(; factors=factors[step]);
+                color=(_colors[factors[step]], _alphas[factors[step]]),
+            )
+            band!(
+                ax,
+                n_scenarios,
+                Si_conv(; Si=:lb)(; factors=factors[step]),
+                Si_conv(; Si=:ub)(; factors=factors[step]);
+                color=(_colors[factors[step]], _alphas[factors[step]]),
+            )
```
ConnectedSystems, line 355: I say drop the bands for now.

This is showing the upper/lower bounds of the results at conditioning points. There are only 10 of these, so the bounds shown here is likely misleading.

The usual approach is to bootstrap the entire analysis.

Rosejoycrocker: So just plot the lines?

ConnectedSystems: Ehm, on second thought maybe put in a big warning about what the bounds are showing?

We'll want to add bootstrapping at some point.

The figures are showing something weird at the moment though

ConnectedSystems: image


It seems the upper/lower bounds are swapping places?


image

Rosejoycrocker: ooh that's weird, let me check it

Rosejoycrocker: From what I can see the lowera nd upper bounds are giving some weird values, for example:
```julia
infil> Si_conv(; Si=[:lb, :median, :ub])
3-dimensional NamedDimsArray(KeyedArray(...)) with keys:
↓   factors ∈ 3-element Vector{Symbol}
→   Si ∈ 3-element view(::Vector{Symbol},...)
◪   n_scenarios ∈ 10-element Vector{Int64}
And data, 3×3×10 view(::Array{Float64, 3}, :, [2, 4, 5], :) with eltype Float64:
[showing 3 of 10 slices]
[:, :, 1] ~ (:, :, 25):
                    (:lb)      (:median)  (:ub)
  (:dhw_scenario)    0.765149   0.871221   1.0
  (:wave_scenario)   0.144206   0.631884   0.58991
  (:guided)          0.172928   0.705589   0.755389

[:, :, 6] ~ (:, :, 150):
                    (:lb)      (:median)  (:ub)
  (:dhw_scenario)    0.558062   0.629598   0.837923
  (:wave_scenario)   0.603173   0.728472   0.45569
  (:guided)          0.657582   0.591617   0.714464

[:, :, 10] ~ (:, :, 250):
                    (:lb)      (:median)  (:ub)
  (:dhw_scenario)    0.54634    0.573742   0.743484
  (:wave_scenario)   0.554806   0.56338    0.478
  (:guided)          0.563533   0.605634   0.66728
```
In this case the upper bound is less than the median for `:dhw_scenario`. There are some similar anomalies for other variables. Could this happen if the sample size is too small maybe? I couldn't see anything wrong with how the lower and upper bounds are calculated in `pawn()`. I haven't done a very large sample yet but I can try that to see if it is the issue.

Rosejoycrocker: I looked into it further- it's the column-wise normalisation which is causing this in `analysis.convergence`.  Because we normalise relative to other factors the normalised upper bound can be lower than the normalised median. Maybe I just do median +/-std*0.5 or something for now and implement bootstrapping later?

ConnectedSystems: Okay, I think drop the column normalization - it's inappropriate in this case. The col-norm was purely so you could assess factor ranks for each statistic of interest. I should have caught that earlier, sorry.

Rosejoycrocker: Thanks, I'll do that

Rosejoycrocker:
I think I've addressed your comments now. All tests are passing on my end.

docs/src/usage/analysis.md
```diff
@@ -236,6 +236,53 @@ save("tsa.png", tsa_fig)
...
+The function `sensitivity.convergence` can be used to calculate a sensitivity measure for an
+increasing number of samples. The result can then be plotted as band plots or a heat map
+using `viz.convergence`.
+
+```julia
+using Statistics
+
+outcome = dropdims(mean(ADRIA.metrics.scenario_total_cover(rs); dims=:timesteps), dims=:timesteps)
+
+# Display convergence for specific factors of interest ("foi") as heat map
```
ConnectedSystems, line 260:
```suggestion
# Display convergence for specific factors of interest ("foi") as a heat map
```

docs/src/usage/analysis.md
```diff
@@ -236,6 +236,53 @@ save("tsa.png", tsa_fig)
...
+using Statistics
+
+outcome = dropdims(mean(ADRIA.metrics.scenario_total_cover(rs); dims=:timesteps), dims=:timesteps)
+
+# Display convergence for specific factors of interest ("foi") as a heat map
+foi = [:dhw_scenario, :wave_scenario, :guided]
+Si_conv = ADRIA.sensitivity.convergence(scens, outcome, foi; opts=Dict(:viz_type=>:heatmap))
+ADRIA.viz.convergence(Si_conv, foi)
+
+# Convergence analysis of factors grouped by model component as heat map
```
ConnectedSystems, line 265:
```suggestion
# Convergence analysis of factors grouped by model component as a heat map
```

docs/src/usage/analysis.md
```diff
@@ -236,6 +236,53 @@ save("tsa.png", tsa_fig)
...
+![Grouped convergence analysis](/ADRIA.jl/dev/assets/imgs/colormap_convergence_components.png?raw=true "Convergence Analysis - model components")
+
+```julia
+# Display convergence plot within a single figure
+ADRIA.viz.convergence(Si_conv, foi)
+
+# Create a grid of figures for each factor of interest
+ADRIA.viz.convergence(Si_conv, foi; opts=Dict(:plot_overlay=>false))
+
+```
```
ConnectedSystems, line 281: Please add note/warning about what the confidence interval represents here

ConnectedSystems (changes requested):
Just one more addition please - adding a note about what the confidence interval actually represents, at least until we have bootstrapping in.

Rosejoycrocker:
> Just one more addition please - adding a note about what the confidence interval actually represents, at least until we have bootstrapping in.

Thanks, I added a note

ConnectedSystems:
> > Just one more addition please - adding a note about what the confidence interval actually represents, at least until we have bootstrapping in.
> 
> Thanks, I added a note

I don't think it captures what we discussed above but it's fine, I'll add it and merge.

ConnectedSystems (approved):

PR #555, opened 2023-10-30
Coral cover consistently represented as relative to coral habitable area

ConnectedSystems:
A whole slew of changes to change representation of coral cover to be relative to available habitable area ($k$).

This, combined with changes in #542, enables comparatively faster simulations and reduced memory use.

The tests we currently have all pass.

@Zapiano @Rosejoycrocker I think the most important thing to check:

1. All metrics are calculated correctly. I did my best but would benefit from a second/third pair of eyes.
2. All workflows can be run as expected without error (e.g., those listed in the cookbook and analysis pages in the docs)


ConnectedSystems:
I should mention - I reviewed the changes in `dMCDA.jl` separately (#547), if you want to do a separate re-review, that's fine with me.

Zapiano (approved):
Only some minor comments.

src/Domain.jl
```diff
@@ -232,7 +260,7 @@ end
 Get maximum coral cover area as a proportion of site area.
```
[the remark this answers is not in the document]
ConnectedSystems: Hmm, yes the description is correct but poorly worded.

It means maximum possible coral cover. I'll simplify.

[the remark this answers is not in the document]
ConnectedSystems: Took the opportunity to deprecate the function in favour of `location_k()`

src/ExtInterface/ADRIA/Domain.jl
```diff
@@ -114,6 +114,7 @@ function Domain(name::String, dpkg_path::String, rcp::String, timeframe::Vector,
 
     # Filter out missing entries
     site_data = site_data[coalesce.(in.(conn_ids, [site_conn.site_ids]), false), :]
+    site_data.k .= site_data.k / 100.0  # Make `k` non-dimensional (provided as a percent)
```
Rosejoycrocker, line 117: It doesn't really matter but what's the reason for moving division by 100 outside of the function?

src/decision/dMCDA.jl
```diff
@@ -604,53 +631,68 @@ function guided_site_selection(
     # if seeding, create seeding specific decision matrix
     if log_seed
         SE, wse = create_seed_matrix(
-            A, d_vars.min_area, d_vars.wt_in_conn_seed, d_vars.wt_out_conn_seed,
-            d_vars.wt_waves, d_vars.wt_heat, d_vars.wt_predec_seed, d_vars.wt_zones_seed,
-            d_vars.wt_lo_cover
+            A,
```
Rosejoycrocker, line 634: Maybe I did this in the mcda branch but might be more consistent to have every input on a new line, or 2 lines of the same length for inputs instead of some on a new line and some on the same line.

src/scenario.jl
```diff
@@ -580,40 +573,48 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
-        leftover_space_m² = leftover_space_prop .* total_loc_area
+        loc_coral_cover = sum(C_t, dims=1)  # dims: 1 * nsites
+        leftover_space_m² = relative_leftover_space(loc_coral_cover) .* loc_k_area
 
         # Recruitment represents additional cover, relative to total site area
-        # Gets used in ODE
-        p.rec .= settler_cover(fec_scope, TP_data, leftover_space_prop,
-            sim_params.max_settler_density, sim_params.max_larval_density, basal_area_per_settler)
+        # Recruitment/settlement occurs after the full moon in October/November
+        p.rec[:, valid_locs] .= settler_cover(
```
Rosejoycrocker, line 599: Wasn't p.rec taken out of the ODE?

src/ExtInterface/ADRIA/Domain.jl
ConnectedSystems: The reason is to ensure the value in `k` is always loaded as $\in [0, 1]\$ regardless of the source of the data for consistency.

If it is inside `location_k`, it needs to be divided by 100 every time, and we need separate functions to handle sources that provide it as [0, 1] vs [0, 100], or some future hypothetical source that provides absolute units (m^2, km^2, hectares, etc).

Best to do it once in the relevant `load_domain()`.

src/scenario.jl
ConnectedSystems: Yes, it was - `p.rec` is now just re-used as a cache to store the updated values (down below it is used to update coral cover for time $t$).

Agree that it makes sense to move `rec` into the `cache` object, I guess I was being lazy.

Good pick up.

src/ExtInterface/ADRIA/Domain.jl
Rosejoycrocker: Thanks, makes sense

Rosejoycrocker:
I ran the examples in analysis.md and cookbook.md and these all run fine besides some unrelated things that need to be updated. Some of these maybe you could update @Zapiano :
- I got an error in ADRIA.viz.map where it is used to plot spatial clustering because "cluster_colors" doesn't exist (I think maybe the function name has changed to _colors but thought I'd check first).
- I got an error due to `time_series_clustering` not existing. I think again the name has been updated.

- I got an error in `outcome_map` due to invalid target factor "RCP". I think this is due to changes to how RCP is indexed?

These aren't related to changes in consistent_cover but thought I'd just flag as some of them are too small to warrant an issue.

Rosejoycrocker (approved):
Looks good to me, I've noted some unrelated errors when running examples.

ConnectedSystems:
All requested changes made, ready for someone to press the big green button.

ConnectedSystems:
Code reviewed in this branch coveres #488, hence why merging this PR will close that issue.

Zapiano:
> All requested changes made, ready for someone to press the big green button.

Actually the button is blue.

PR #558, opened 2023-10-31
Avoid assuming where the RCP column is when adjusting scenario spec

### LINKED ISSUE #518: Skip RCP entry from scenario specification by name rather than assuming position
Oooo, this needs to be fixed.

The assumption that the `RCP` entry is before any other categorical/discrete factor may not be true.
I'll open a new issue.

_Originally posted by @ConnectedSystems in https://github.com/open-AIMS/ADRIA.jl/pull/491#discussion_r1344871217_

ConnectedSystems:
The column indicating the RCP was originally the first entry but can be the first or last column depending on how/who the scenario spec was generated (and what version of ADRIA).

Remove the assumption that the RCP column is before any other discrete-valued parameter.

Fixed a bug where values set as constants by the user were being floored as well.

Closes #518

### DIFF
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -51,7 +51,7 @@ Length of `u` (the upper bounds) is expected to match number of columns in `df`.
 """
 function map_to_discrete!(
     df::Union{DataFrame,SubDataFrame},
-    u::Union{AbstractVector{Union{Int64,Float64}},Tuple}
+    u::Union{AbstractVector{<:Union{Int64,Float64}},Tuple}
 )::Nothing
     for (idx, b) in enumerate(u)
         df[!, idx] .= map_to_discrete.(df[!, idx], b)
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -44,12 +44,15 @@ function map_to_discrete(v::Union{Int64,Float64}, u::Union{Int64,Float64})::Int6
 end
 
 """
-    map_to_discrete!(df::DataFrame, u::AbstractArray)::Nothing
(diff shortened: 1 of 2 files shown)


src/io/result_io.jl
```diff
@@ -262,9 +262,16 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
-    # +1 skips the RCP column
-    integer_params = findall(_check_discrete.(domain.model[:ptype]))
-    map_to_discrete!(scen_spec[:, integer_params.+1], getindex.(domain.model[:bounds], 2)[integer_params])
+    ms = model_spec(domain)
+
+    # Find discrete parameters that are not constant
+    discrete_param_idx = findall(_check_discrete.(ms.ptype) .& .!ms.is_constant)
+    discrete_params = @view scen_spec[!, Not(:RCP)][!, discrete_param_idx]
+
+    # Map continuous values to discrete, and fill scenario spec with updated values.
```
ConnectedSystems, line 271: This comment may not be that useful 🤷

Rosejoycrocker: Yeah I guess the function name explains what it does already

src/ecosystem/Ecosystem.jl
```diff
@@ -44,12 +44,15 @@ function map_to_discrete(v::Union{Int64,Float64}, u::Union{Int64,Float64})::Int6
 end
 
 """
-    map_to_discrete!(df::DataFrame, u::AbstractArray)::Nothing
+    map_to_discrete!(df::Union{DataFrame,SubDataFrame}, u::Union{AbstractVector{Union{Int64,Float64}},Tuple})::Nothing
```
Rosejoycrocker, line 47: I haven't come across a SubDataFrame type before- does this include things like DataFrameRow?

Rosejoycrocker (approved):
Looks good to me. I ran some scenarios and they ran fine. An unrelated test errored but I'll look into this separately (it was for guided site selection).

PR #563, opened 2023-11-01
Fix integer dims reference in location_selection_frequency()

### LINKED ISSUE #562: Make all indices used in site selection aggregation functions symbols
In `location_selection_frequency` a non-symbol indexing for dims is used when summing over ranks. This is no longer correct and should be changed to a symbol to avoid future errors.

Rosejoycrocker:
Closes #562

A dims reference in `location_selection_frequency()` was incorrect. This has been changed to a symbol to avoid any future issues.


ConnectedSystems:
Hi Rose, just FYI that if you have a fix already implemented there's no need to open an issue and then a PR that closes the issue straight afterwards.

Just submit the PR.

ConnectedSystems (approved):

PR #570, opened 2023-11-08
Visualizations to be agnostic to Makie backend

ConnectedSystems:
GLMakie may be preferred when creating visualizations locally, but allows WGLMakie to be installed instead for web/browser-based plots.

By default, figures will appear in the VS Code plot pane:

image

To switch backends, either:

```julia
] remove GLMakie
] add WGLMakie
```

or have both installed and activate the desired backend:

```julia
using WGLMakie
WGLMakie.activate!()

# or

using GLMakie
GLMakie.activate!()
```

https://docs.makie.org/stable/explanations/backends/

### DIFF
--- a/Aviz/build/precompile_script.jl
+++ b/Aviz/build/precompile_script.jl
@@ -1,7 +1,7 @@
 using CSV, DataFrames, GeoDataFrames, GeoInterface
 using Statistics, Distributions
-using GLMakie.GeometryBasics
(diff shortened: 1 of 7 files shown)


Zapiano:
Should we also update the parts of the doc that say `using GLMakie`?

image

ConnectedSystems:
Done.

Zapiano (approved):

PR #571, opened 2023-11-08
Add functions used for location selection to separate `decision` module

### LINKED ISSUE #556: Deployment location analysis methods/metrics leak into package namespace
I just noticed that location analysis methods and related metrics are imported directly from the package rather than being contained into appropriate submodules. 

```
using ADRIA:
    rank_locations,
    ranks_to_frequencies,
    location_selection_frequencies,
    selection_score
```

Rosejoycrocker:
Adds a module called `decision` which contains functions and other definitions relevant to location selection. Also renames `Criteria` type to `CriteriaWeights` to reduce confusion. Note `CriteriaWeights.jl` is not included in the `decision` module due to being used in many places other than location selection, but can be included if this is best.

Closes #556


src/ADRIA.jl
```diff
@@ -56,19 +56,18 @@ include("ecosystem/corals/Corals.jl")
...
 include("ecosystem/connectivity.jl")
 
+include("decision/CriteriaWeights.jl")
+
 include("Domain.jl")
 include("io/inputs.jl")
 
-include("decision/mcda_methods.jl")
-include("decision/Criteria.jl")
 include("decision/dMCDA.jl")
```
ConnectedSystems, line 64: Either move the include for `decision/dMCDA.jl` into the submodule or group the includes together - which ever is most appropriate.

test/spec.jl
```diff
@@ -18,7 +18,7 @@ end
     x = ADRIA.component_params(dom.model, ADRIA.Coral)
     @test size(x, 1) > 0
 
-    x = ADRIA.component_params(dom.model, Criteria)
+    x = ADRIA.component_params(dom.model, ADRIA.CriteriaWeights)
     @test size(x, 1) > 0
 
 end
```
ConnectedSystems, line 24:
```suggestion
end

```

ConnectedSystems (approved):
One minor change to group related includes together.

src/ADRIA.jl
Rosejoycrocker: I think I put them in this order because `CriteriaWeights.jl` is needed in `Domain.jl` and `Domain.jl` is needed in `dMCDA.jl`.

Currently the submodule `decision` is inside `dMCDA.jl`. Should I instead make a separate file called `decision.jl` which contains the module and all the imorts?

ConnectedSystems: > I think I put them in this order because `CriteriaWeights.jl` is needed in `Domain.jl` and `Domain.jl` is needed in `dMCDA.jl`.

This is a smell that suggests you need to rethink the structure then.

Why is `domain` needed in `dMDCA`? Can you separate out the dependency?

Rosejoycrocker: The Domain type is used to extract parameters used in the `DMCDA_vars` constructor. This includes `sim_constants`, `TP_data`, `median_site_distance` and `site_distances`. I could make these separate inputs but it would mean increasing the number of input arguments. Should I do this?

ConnectedSystems: Hang on - `DMCDA_vars` essentially acts as a data container. 

Do we need that at all?

If we do need it, everything is being shoved into it and only some of the data is relevant to each of the steps/methods that take it as an argument. Could you not split that up to something more appropriate and smaller?

ConnectedSystems: Okay, it seems you've uncovered a big smell that points to a required redesign. 
It's going to be too big for this PR so leave things be.

**Add a comment that the include order has to be like it is and why.**

Rosejoycrocker: The plan in the changes for making the MCDA more generic was come up with an alternative to MCDA_vars because it is a veryt inefficient way of doing things. I think in future changes the scenario struct itself should be passed in (as MCDA_vars is predominantly a container criteria weights), with a NamedDimsArray for criteria (essentially the decision matrix as a NamedDimsArray) for storing criteria values. But agree this is part of a larger problem.

Rosejoycrocker:
Conflicts are resolved and tests all passing on my end, ready for the final look over @ConnectedSystems

Issue #559, opened 2023-11-01
Add MCDA weighting normalisation to scenario sampling post-processing 


Rosejoycrocker:
Currently MCDA weighting parameters are sampled from within uniform distributions and then normalised inside `guided_site_selection` so that the sum of weightings for each of the intervention decision matrices is 1. This should instead be part of the post-processing of scenario samples prior to running scenarios in the ecological model, so that the normalisation is not repeated multiple times within `guided_site_selection` when running scenarios and the inputs reflect the magnitude of the parameters as they are used within the model.

PR #573, opened 2023-11-08
Add separate weighting parameter for depth criteria

### LINKED ISSUE #565: Separate weighting parameters for `heat_stress` and depth in the MCDA
Currently the depth criteria, used in the seeding decision matrix for location selection,  is weighted using the same weighting parameter for heat stress. The depth criteria should be given it's own weighting for clarity and ease of interpretation of sensitivity analysis.

Rosejoycrocker:
Previously the depth criteria was weighted with the same weighting as heat stress. This PR adds a separate weighting for the depth criteria.

Closes #565


src/decision/Criteria.jl
```diff
@@ -44,6 +44,15 @@ Base.@kwdef struct Criteria{P,N} <: EcoModel
         name="Outgoing Connectivity (Seed)",
         description="Higher values give preference to locations with high outgoing connectivity (i.e., provides larvae to other sites) for enhanced coral deployments.",
     )
+    seed_depth::P = Param(
```
Zapiano, line 47: Everywhere else you are calling this `depth` so, for consistency, I would change to `depth` here as well.

Also, a question: this is not the seed depth, but the location depth, right? (otherwise maybe everywhere else should change to `seed_depth` instead..)

src/decision/Criteria.jl
```diff
@@ -44,6 +44,15 @@ Base.@kwdef struct Criteria{P,N} <: EcoModel
...
+        1.0;
+        ptype="real",
+        bounds=(0.0, 1.0),
+        default_bounds=(0.0, 1.0),
+        dists="unif",
+        name="Depth (Seed)",
+        description="Higher values give preference to depper locations for enhanced coral deployments.",
```
Zapiano, line 54: Here you are describing how you intend to use, but I think it should be a description of what this is. Something like "The depth of a location". Because we can use for other things later on, right?

src/decision/dMCDA.jl
```diff
@@ -109,6 +110,7 @@ function DMCDA_vars(
         criteria("shade_connectivity"),
         criteria("wave_stress"),
         criteria("heat_stress"),
+        criteria("seed_depth"),
```
Zapiano, line 113: If you do change the name, remember to change it here as well.

Zapiano (changes requested):
Just some minor comments about naming.

Zapiano:
@Rosejoycrocker  also, tests are not passing:

image

src/decision/Criteria.jl
Rosejoycrocker: This is not the depth of a location, it is the weighting for the depth of a location criteria in the location selection (scalar). It is only used in deciding where to seed.

src/decision/Criteria.jl
Rosejoycrocker: For the other weights if they are not used for multiple interventions, we tend to have the intervention it is used for in the name. This is not completely consistent across parameters but I intend to fix this in #559

Zapiano: But then why not changing `wt_depth`  to `wt_seed_depth`?

Rosejoycrocker:
> @Rosejoycrocker also, tests are not passing:
> 
> image

Weird, they were passing for me, I'll double check.

src/decision/Criteria.jl
Rosejoycrocker: I can do that :)

Rosejoycrocker:
> @Rosejoycrocker also, tests are not passing:
> 
> image

I checked and it's an unrelated error in the growth ODE which is erroring, perhaps due to changes from `consistent-cover`. All relevant tests are passing (domain loading, running scenarios, site_selection, mcda)

Zapiano:
> > @Rosejoycrocker also, tests are not passing:
> > image
> 
> I checked and it's an unrelated error in the growth ODE which is erroring, perhaps due to changes from `consistent-cover`. All relevant tests are passing (domain loading, running scenarios, site_selection, mcda)

Hum, but at the main branch all tests pass... why do you think so?

Rosejoycrocker:
> > > @Rosejoycrocker also, tests are not passing:
> > > image
> > 
> > 
> > I checked and it's an unrelated error in the growth ODE which is erroring, perhaps due to changes from `consistent-cover`. All relevant tests are passing (domain loading, running scenarios, site_selection, mcda)
> 
> Hum, but at the main branch all tests pass... why do you think so?

Strange, I'll rebase and see if that works

src/decision/dMCDA.jl
```diff
@@ -380,6 +382,7 @@ Create seeding specific decision matrix from criteria matrix. The weight criteri
 - `wt_out_conn_seed` : Seed connectivity weight for seeding
 - `wt_waves` : Wave stress weight
 - `wt_heat` : Heat stress weight
+- `wt_depth` : Median depth weight
```
Zapiano, line 385: Shouldn't this change as well?

test/mcda.jl
```diff
@@ -49,7 +49,9 @@ end
 
 
 @testset "MCDA seed matrix creation" begin
-    wtconseedout, wtconseedin, wt_waves, wt_heat, wt_predec_seed, wt_zones_seed, wt_lo_cover = [1.0, 1.0, 0.7, 1.0, 0.6, 0.6, 0.6]
+    wtconseedout, wtconseedin, wt_waves, wt_heat, wt_predec_seed, wt_zones_seed, wt_lo_cover, wt_depth = [
```
Zapiano, line 52: change `wt_depth` to `wt_depth_seed` here too?

test/mcda.jl
```diff
@@ -88,7 +90,18 @@ end
...
+        A,
+        min_area,
+        wtconseedin,
+        wtconseedout,
+        wt_waves,
+        wt_heat,
+        wt_predec_seed,
+        wt_zones_seed,
+        wt_lo_cover,
+        wt_depth,
```
Zapiano, line 103: And here too?

Rosejoycrocker:
Hi @ConnectedSystems, just a question. I'm getting an error in tests which is due to site covers being a very small amount over 1 
```julia
┌ Warning: Cover exceeded bounds, constraining to be within available space, but this indicates an issue with the model.
└ @ ADRIA C:\Users\rcrocker\Documents\GitHub\ADRIA.jl\src\ecosystem\corals\growth.jl:50
┌ Warning: Cover - Max Cover: 4.440892098500626e-15
```
It's weird because the changes in this PR are unrelated and shouldn't effect this test. Have you come across this before? I tried rebasing but still the same error despite no changes to any related functions or values.

Rosejoycrocker:
Hey @Zapiano, tests are all passing for me now. I spoke to Takuya about it and it's a rounding error introduced due to the seed changing when the randomised matrix is created in the proportional adjustment test. I've adjusted the creation of the matrix so the rounding errors shouldn't affect the test any more.

src/decision/Criteria.jl
Zapiano, line 47: Just one last comment: in all other places you wrote `wt_depth_seed`, but here you wrote `seed_depth`. Do you think it would be ok to change this one to `depth_seed` as well, just to keep the pattern?

Zapiano (approved):
I only have one comment for you to consider, but other than that it looks good to me.

src/decision/Criteria.jl
Rosejoycrocker: I can do this. There's not consistency within Criteria with whether seed is at the end or beginning of the name but I can change it for this parameter

Zapiano: It's never too late to start 😅

PR #574, opened 2023-11-09
Add conversion to scientific notation

Zapiano:
As in title.


Zapiano:
@Rosejoycrocker and @ConnectedSystems  I think the most important thing here is to check if I am missing some test scenario - I found a lot of problems after writing these tests.

Zapiano:
Also: I think it would be good to implement some functions that use this to convert numbers like "18745234" to "18 Billion" (or something similar, the specifics are open to discussion). I thin:

1. This needs more discussion (so after this is approved I can open a discussion on Github where we can agree on which different formats we need and what are the expected returns)
2. Once this PR is approved, implementing those visualizations will be a `good first issue` ;)

src/utils/text_display.jl
```diff
@@ -28,4 +28,74 @@ function human_readable_name(name::String; title_case::Bool=false)::String
...
+
+"""
+    get_scientific_factors(x::Float64; digits=2)::Tuple
+    get_scientific_factors(x::Int64; digits=2)::Tuple
+
+Return a and b to write any Real number x as x = a * 10^b, where 1.0 <= a < 10.0 and b is
+an integer. The idea is that `x = a * 10^b` => `log(x) = log(a) + b*log(10)`. Which means
+that `log(a)` is the remainder of `log(x)/log(10)` hence `a = exp((log(x)%log(10)))`.
+Because `log(x) % log(10)` returns an approximate value, if we just use this to find `a` we
+will end up writing `40` as `3.9E1`. To avoid that, we take this approximated value of `a`,
```
ConnectedSystems, line 61:
```suggestion
will end up writing `40` as `3.9E1`. To avoid that, we take this approximate value of `a`,
```

ConnectedSystems:
You're going to hate me for this, but what was the issue with `@sprintf` again?

```julia
using Printf

@sprintf("%.2E", 1_000_000)
# "1.00E+06"

@sprintf("%.2E", -1_000_000)
# "-1.00E+06"

@sprintf("%.2E", 0)
# "0.00E+00"
```

Zapiano:
> You're going to hate me for this, but what was the issue with `@sprintf` again?
> 
> ```julia
> using Printf
> 
> @sprintf("%.2E", 1_000_000)
> # "1.00E+06"
> 
> @sprintf("%.2E", -1_000_000)
> # "-1.00E+06"
> 
> @sprintf("%.2E", 0)
> # "0.00E+00"
> ```

1 - You complained about using a macro for this - does it slow down the code?
2 - It's not too flexible. If we want to convert "1234" to "1.2k", or if we don't want the "+" sign when is positive, or if we want to use "10^" (with superscript), etc, we can't.

ConnectedSystems:
Ah, I think the context was different. I prefer avoiding using an import for a single use case (which was essentially just rounding some numbers for display from memory). But here there's a good reason to use it.

That said, I agree with point 2.

And no, well-written macros should not be slow, in fact they should be very fast.

src/utils/text_display.jl
```diff
@@ -28,4 +28,74 @@ function human_readable_name(name::String; title_case::Bool=false)::String
...
+    return to_scientific(Float64(x); digits=digits)
+end
+
+"""
+    get_scientific_factors(x::Float64; digits=2)::Tuple
+    get_scientific_factors(x::Int64; digits=2)::Tuple
+
+Return a and b to write any Real number x as x = a * 10^b, where 1.0 <= a < 10.0 and b is
+an integer. The idea is that `x = a * 10^b` => `log(x) = log(a) + b*log(10)`. Which means
+that `log(a)` is the remainder of `log(x)/log(10)` hence `a = exp((log(x)%log(10)))`.
```
Rosejoycrocker, line 59: Maybe use $a\approx exp((log(x)%log(10)))$. I know it's basically equality but makes it clear.

Rosejoycrocker: Actually I think you're talking about compuational approximation not mathematical sorry, so ignore me

Rosejoycrocker (approved):
Looks good to me, all tests passing on my end

test/utils/test_display.jl
```diff
@@ -0,0 +1,40 @@
+include("../../src/utils/text_display.jl")
```
ConnectedSystems, line 1: This will break depending on how tests are run. We can't assume relative paths will always work, or rely on tests being invoked from the same initial path every time.

If this is absolutely necessary, use `pkgdir(ADRIA)` instead:

```julia
joinpath(pkgdir(ADRIA), "src", "utils", "text_display.jl")
```

ConnectedSystems: I also have a conceptual issue with the file name `test_display`.

It already is in a directory called `test` so naturally everything in it will be related to tests.
Best to name it after what it is testing.

test/utils/test_display.jl
```diff
@@ -0,0 +1,40 @@
...
+
+@testset "text_display" begin
+    @testset "get_scientific_factors" begin
+        get_scientific_factors = ADRIA.get_scientific_factors
+
+        @testset "Edgecases" begin
+            @test get_scientific_factors(0.0) == (0.0, 0)
```
ConnectedSystems, line 8: Meta comment - no action required as such.

```julia
julia> get_scientific_factors(1e6)  # or 1_000_000.0
(10.0, 5)

julia> 10e5
1.0e6
```

Although they are the same, 1e6 is more understandable to me compared to 10e5 but I suppose it's hard to generalize these down further.

ConnectedSystems (approved):
Okay with me. 

I prefer more human friendly printed values (e.g., 1M vs 1e6) but I see the value in this.

Just the one change to the file include in the tests and then I will merge.

test/utils/test_display.jl
Zapiano: I.. don't think I need to include anything at all, let me test removing that to see what happens.

test/utils/test_display.jl
Zapiano: So, that was a typo, my idea was to name this file `test/utils/text_display.jl`. Do you think this is ok?

Zapiano:
> Okay with me.
> 
> I prefer more human friendly printed values (e.g., 1M vs 1e6) but I see the value in this.
> 
> Just the one change to the file include in the tests and then I will merge.

I also think we should have something like 1M instead of 1e6. And this is going to allow us to choose between those different options.

test/utils/test_display.jl
ConnectedSystems: I think you replied to the wrong comment, but okay with me.

ConnectedSystems: On the above: a bit hacky, but what if we parse the value determined by your function to get Julia to auto-infer the simplified value?

```
julia> parse(Float64, "10e5")
1.0e6

julia> string(parse(Float64, "10e5"))
"1.0e6"
```

Zapiano: Hm... let me think about it for a while...

Zapiano: @ConnectedSystems  I think I solved the problem (and I've added some more tests, thanks for catching that). Care to take a look?

ConnectedSystems: Ooo, how about the opposite sign?

```julia
julia> -0.000001
-1.0e-6

julia> to_scientific(-0.000001)
"-10.0E-7"
```

Zapiano: Cool, going to fix that too, thanks!

Zapiano: @ConnectedSystems Done!

PR #576, opened 2023-11-09
Prevent test_rs() from creating tmp dir inside test/data

Zapiano:
As in title.

The way it was, every time we run `include("test\\runtests.jl")` a tmp folder was being created inside `test\data` and was not being deleted.


ConnectedSystems:
Thanks, I didn't realise this was an issue as I typically run tests from scratch via the package manager.

All looks good to me so will merge shortly.

ConnectedSystems (approved):

PR #581, opened 2023-11-14
Update use of progress meter to resolve deprecation warning

ConnectedSystems:
Recent version requires use of named arguments:

```julia
@showprogress desc=run_msg dt=4 ...
```

Otherwise a deprecation warning is thrown:

```julia
┌ Warning: `Progress(n::Integer, desc::AbstractString, offset::Integer = 0; kwargs...)` is deprecated, use `Progress(n; desc = desc, offset = offset, kwargs...)` instead.
│   caller = ip:0x0
└ @ Core :-1
```


Zapiano (approved):
THANK YOU

PR #583, opened 2023-11-15
Replace package NCDatasets by NetCDF

Zapiano:
As in title. This change was made to enable compatibility with YAXArrays package.


src/io/inputs.jl
```diff
@@ -121,76 +122,61 @@ end
...
-        if isa(err, KeyError)
-            n_sites = size(data, 2)
-            @warn "Provided file $(data_fn) did not have the expected dimensions (one of: timesteps, reef_siteid, scenarios)."
-            if n_sites != nrow(site_data)
-                error(
-                    "Mismatch in number of sites ($(data_fn)). Expected $(nrow(site_data)), got $(n_sites)",
-                )
+        try
+            NamedDimsArray(data; zip(dim_names, dim_labels)...)
+        catch err
```
ConnectedSystems, line 134: Where does the NamedDimsArray go? It's not assigned to a variable?

src/io/inputs.jl
```diff
@@ -121,76 +122,61 @@ end
...
     end
 
     return loaded
 end
 
 """
-    _char_to_string(vals)::Vector{String}
+    Return vector of labels for each array dimension. Important: Cannot trust indicated
+    dimension metadata to get site labels, because this can be incorrect. Instead, match by
+    number of sites.
```
ConnectedSystems, line 155: Incorrect indentation?

src/io/inputs.jl
```diff
@@ -121,76 +122,61 @@ end
...
-    if vals isa Matrix
-        vals = map(x -> join(skipmissing(x)), eachcol(vals))
+    try
+        # This will be an issue if two or more dimensions have the same number of elements
+        # as the number of sites, but so far it hasn't happened...
+        sites_idx = first(findall(size(data) .== length(sites)))
+    catch err
+        error(
+            "Error loading $data_fn : could not determine number of locations." *
+            "Detected size: $(size(data)) | Known # sites: $(length(sites))",
```
ConnectedSystems, line 172:
```suggestion
            "Detected size: $(size(data)) | Known number of locations: $(length(sites))",
```

src/io/inputs.jl
Zapiano: Because it's the last thing on that block, it is assigned to `loaded`

src/io/inputs.jl
Zapiano: yes!

ConnectedSystems (changes requested):
Two questions and a suggested change

src/io/inputs.jl
ConnectedSystems: Oh, I didn't see it's part of a `do` block.

I prefer explicit returns or assignment - particularly in cases like this where the returned value is far away from assignment location.

```julia
local data
open(...) do
    try
        data = NamedDimsArray(...)
    catch err
        # ...
    end
end
```

Zapiano: Ok, I'll change that :)

Zapiano:
@ConnectedSystems changes made, ready for another review.

src/io/inputs.jl
```diff
@@ -121,76 +122,65 @@ end
...
 
-    # R's ncdf4 package does not yet support string values
-    # so strip the null terminator from the joined character array.
-    vals = replace.(vals, "\0" => "")
+    dim_labels = Union{UnitRange{Int64},Vector{String}}[1:n for n in size(data)]
+    dim_labels[sites_idx] = sites
 
-    return vals
+    return dim_labels
 end
```
ConnectedSystems, line 184: You sure character arrays are handled properly?

Zapiano: Can you give me an example of what was the problem before? Is there a test for this (I can write one if you give me an example)?

src/io/inputs.jl
```diff
@@ -121,76 +122,65 @@ end
...
+            "Error loading $data_fn : could not determine number of locations." *
+            "Detected size: $(size(data)) | Known number of locations: $(length(sites))",
+        )
     end
 
-    # R's ncdf4 package does not yet support string values
-    # so strip the null terminator from the joined character array.
-    vals = replace.(vals, "\0" => "")
+    dim_labels = Union{UnitRange{Int64},Vector{String}}[1:n for n in size(data)]
+    dim_labels[sites_idx] = sites
```
ConnectedSystems, line 181: How about:


```suggestion
    dim_labels = Union{UnitRange{Int64},Vector{String}}[
        idx != sites_idx ? 1:n : sites for (idx, n) in enumerate(size(data))
    ]
```

ConnectedSystems: The current approach is probably more readable though

src/io/inputs.jl
```diff
@@ -121,76 +122,65 @@ end
...
 
-    return loaded
+    return loaded_data
 end
 
 """
-    _char_to_string(vals)::Vector{String}
+Return vector of labels for each array dimension. Important: Cannot trust indicated
+dimension metadata to get site labels, because this can be incorrect. Instead, match by
+number of sites.
```
ConnectedSystems, line 159:
```suggestion
Return vector of labels for each dimension.

**Important:** Cannot trust indicated dimension metadata to get site labels,
because this can be incorrect. Instead, match by number of sites.
```

ConnectedSystems (changes requested):
Some small comments for consideration

src/io/inputs.jl
Zapiano: Yes... I thought about doing what you wrote, but it's harder to read and I didn't see any big advantage... But I'm ok with changing, what do you think?

src/io/inputs.jl
Zapiano: I have added a `::Vector{String}` to `sites` to be sure it is a vector of strings - and the tests are passing. Do you think this is enough?

`sites::Vector{String} = has_reef_siteid_var ? NetCDF.readvar(nc_file, "reef_siteid") : 1:size(data, 2)`

ConnectedSystems: No because it will just fail when reading a character array, rather than a string array. Wait a tick, I'm mocking up an example.

src/io/inputs.jl
ConnectedSystems: Benefit of doing it in a single pass is small, I'm okay with leaving it as is.

src/io/inputs.jl
ConnectedSystems: I'm making an example netcdf but to give a concrete example, it's the difference between these two:

```julia

a = "a string"
b = ['c', 'h', 'a', 'r', 'a', 'c', 't', 'e', 'r']

typeof(a)
# String

typeof(b)
# Vector{Char} (alias for Array{Char, 1})

Comment (Zapiano) [reply src/io/inputs.jl:186]: Hum.. ok, I understand the difference (of course) but my question is: what was happening before that you wrote that function? NetCDF was reading each string as a vector of chars? Because that is definitely not happening now.

Comment (ConnectedSystems) [reply src/io/inputs.jl:186]: I can confirm it doesn't automatically detect character arrays:

```julia
julia> ncread("./char_R_ncdf_test.nc", "reef_siteid")
14×3 Matrix{NetCDF.ASCIIChar}:
 'a'   's'   's'
 'b'   'o'   'o'
 'c'   'm'   'm'
 'd'   'e'   'e'
 '\0'  't'   't'
 '\0'  'h'   'h'
 '\0'  'i'   'i'
 '\0'  'n'   'n'
 '\0'  'g'   'g'
 '\0'  '\0'  ' '
 '\0'  '\0'  'e'
 '\0'  '\0'  'l'
 '\0'  '\0'  's'
 '\0'  '\0'  'e'
```

[char_R_ncdf_test.zip](https://github.com/open-AIMS/ADRIA.jl/files/13370916/char_R_ncdf_test.zip)

NetCDF.jl now seems to have `nc_char2string()` which handles it:

```julia
julia> nc_char2string(ncread("./char_R_ncdf_test.nc", "reef_siteid"))
3-element Vector{String}:
 "abcd"
 "something"
 "something else"
```

https://juliageo.org/NetCDF.jl/v0.5/strings.html#NC_CHAR-variables-1

ConnectedSystems: > Hum.. ok, I understand the difference (of course) but my question is: what was happening before that you wrote that function? NetCDF was reading each string as a vector of chars? Because that is definitely not happening now.

It's not happening now because you're reading in strings, not a character array. If we happen to get a netcdf which uses character arrays instead of strings, then the changed implementation will not work

Zapiano: @ConnectedSystems I've added some tests together as well. Ready for another look

src/io/inputs.jl
```diff
@@ -121,76 +122,75 @@ end
...
 
-    ds = NCDataset(data_fn, "r")
-    dim_names = keys(dimsize(ds[attr]))
-    nd = ndims(ds[attr])
-
-    # Load all data (need same number of colons as number of dimensions)
-    data = ds[attr].var[[(:) for _ in 1:nd]...]
+    NetCDF.open(data_fn; mode=NC_NOWRITE) do nc_file
+        data::Array{<:AbstractFloat} = NetCDF.readvar(nc_file, attr)
+        dim_names::Vector{Symbol} = [Symbol(dim.name) for dim in nc_file.vars[attr].dim]
```
ConnectedSystems, line 129:
```suggestion
        dim_names::Vector{Symbol} = Symbol[Symbol(dim.name) for dim in nc_file.vars[attr].dim]
```

Just adding a type hint for the Vector as well.

src/io/inputs.jl
```diff
@@ -121,76 +122,75 @@ end
...
-
-    # R's ncdf4 package does not yet support string values
-    # so strip the null terminator from the joined character array.
-    vals = replace.(vals, "\0" => "")
-
-    return vals
+function _site_labels(nc_file::NetCDF.NcFile)::Vector{String}
+    site_ids = NetCDF.readvar(nc_file, "reef_siteid")
+    # Converts character array entries in netCDFs to string if needed
+    return site_ids isa Matrix ? nc_char2string(site_ids) : site_ids
```
ConnectedSystems, line 193: Don't change anything, just a note for now but one comment is that it may not just be the site (location) labels - it is any variable or metadata with text data.

src/io/inputs.jl
```diff
@@ -121,76 +122,75 @@ end
...
-    if vals isa Matrix
-        vals = map(x -> join(skipmissing(x)), eachcol(vals))
-    end
-
-    # R's ncdf4 package does not yet support string values
-    # so strip the null terminator from the joined character array.
-    vals = replace.(vals, "\0" => "")
-
-    return vals
+function _site_labels(nc_file::NetCDF.NcFile)::Vector{String}
```
ConnectedSystems, line 190: Location not sites ?

```suggestion
function _loc_labels(nc_file::NetCDF.NcFile)::Vector{String}
```

test/io/inputs.jl
```diff
@@ -0,0 +1,14 @@
+@testset "NetCDF" begin
+    @testset "Handling Vector of Strings and Matrix{ASCIIChar}" begin
+        char_fp = joinpath(TEST_DATA_DIR, "test_ncdf_char.nc")
+        str_fp = joinpath(EXAMPLE_DOMAIN_PATH, "DHWs", "dhwRCP45.nc")
+
+        ADRIA.NetCDF.open(char_fp; mode=ADRIA.NetCDF.NC_NOWRITE) do char_nc
+            @test ADRIA._site_labels(char_nc) isa Vector{String}
+        end
```
ConnectedSystems, line 8: Best to check if the correct/expected strings are returned too :)

ConnectedSystems (changes requested):
Thanks Pedro, just one small change to the tests and I'll merge.

src/io/inputs.jl
Zapiano: In my opinion we should wait to change to location when we change in other places too. Right now the NamedDimsArray dimension is called `sites` so it makes sense for this one to be `sites_labels` as well and only change to `location_labels` or `loc_labels` when we also change the dimension name.

src/io/inputs.jl
Zapiano: @ConnectedSystems I've added the requested tests.

test/io/inputs.jl
Zapiano: @ConnectedSystems I've added the requested tests.

ConnectedSystems (approved):

PR #593, opened 2023-11-20
Change approach to listing supported MCDA methods

ConnectedSystems:
Replace consts with a function to allow flexibility. Previously modifying the list of supported methods required Julia to be restarted as the method list was assigned to a const.

Rather than listing out methods to ignore, we now explicitly list out the methods to include. This avoids situations where new (unsupported) methods are added to JMcDM.

Tests pass locally but visualizations fail (this is fixed in #589).


Rosejoycrocker (approved):
Looks good to me :)

PR #597, opened 2023-11-21
Handle cases where connectivity directories may be empty

ConnectedSystems:
Addresses the issue we ran into @Zapiano


ConnectedSystems:
Note the specs detailed here:

https://github.com/open-AIMS/ADRIA-data-specs

Zapiano (approved):

PR #582, opened 2023-11-14
Normalize sampled MCDA weights

Rosejoycrocker:
Incorporates changes allowing MCDA weights for fogging and seeding decision matrices to be normalised at the parameter sampling step so that parameter inputs reflect their magnitude as used inside ADRIA.

The heat stress and wave stress weights are also split into separate weights for seeding and fogging so that normalisation can be done correctly (otherwise normalised values for seeding will be overwritten by normalised values for shading and these will not be the same).

Tests added to check normalisation is working as expected.

Note the growth.jl test is failing but this is fixed in PR #573

Closes #559.

Closes #590

PR #598, opened 2023-11-22
Deactivated interventions should be considered separately

ConnectedSystems:
If fog/SRM inteventions are not considered separately, post-processing will be incorrectly applied.

https://github.com/open-AIMS/ADRIA.jl/pull/582#discussion_r1393649977
https://github.com/open-AIMS/ADRIA.jl/pull/582#discussion_r1400087136

Rose, is the change in this PR appropriate?


Rosejoycrocker:
I'll have a deeper look, but it may not be because we still use `shade_years` and `shade_start_year` to decide when to fog and when to apply SRM I think. I think it could work if we separate the timing parameters.

ConnectedSystems:
> Or would we generally only be fogging when we shade?

I think they should be considered separately, applying SRM first.

ConnectedSystems:
Okay, could you do that please?

Rosejoycrocker:
> Okay, could you do that please?

Sure :)

ConnectedSystems (commented):
@Zapiano 

Could you please approve this? No need for a review as the bulk of the changes were reviewed in #602.

Tests, plus figure creation, all pass and are confirmed to work.

Zapiano:
@ConnectedSystems do you also want me to merge?

PR #601, opened 2023-11-23
Fix treatment of categorical factors in regional sensitivity analysis

### LINKED ISSUE #459: Regional sensiyivity analysis gives duplicate values for categorical factors
Raised in issue #440 for `outcome_map()`, this should also be fixed in `rsa()`.

Rosejoycrocker:
Closes #459

### DIFF
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -163,6 +163,7 @@ Plot regional sensitivities of up to 30 factors.
 # Arguments
 - `rs` : ResultSet
 - `si` : Results from ADRIA regional sensitivity analysis
+- `opts` : Additional figure customization options
 - `fig_opts` : Additional options to pass to adjust Figure creation
   See: https://docs.makie.org/v0.19/api/index.html#Figure
 - `axis_opts` : Additional options to pass to adjust Axis attributes
@@ -176,6 +177,7 @@ function ADRIA.viz.rsa!(
     rs::ResultSet,
     si::NamedDimsArray,
     factors::Vector{Symbol};
+    opts::Dict=Dict(),
     axis_opts::Dict=Dict(),
 )
     n_factors::Int64 = length(factors)
@@ -272,12 +274,169 @@ function ADRIA.viz.rsa(
     rs::ResultSet,
     si::NamedDimsArray,
     factors::Vector{Symbol};
+    opts::Dict=Dict(),
     fig_opts::Dict=Dict(),
     axis_opts::Dict=Dict(),
 )
     f = Figure(; fig_opts...)
     g = f[1, 1] = GridLayout()
     ADRIA.viz.rsa!(g, rs, si, factors; axis_opts)
+
+    return f
+end
+
+"""
+    ADRIA.viz.outcome_map(rs::ResultSet, outcomes::NamedDimsArray, factors::Vector{String}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
+    ADRIA.viz.outcome_map!(f::Union{GridLayout,GridPosition}, rs::ResultSet, outcomes::NamedDimsArray, factors::Vector{String}; opts, axis_opts)
+
+Plot outcomes mapped to factor regions for up to 30 factors.
+
+# Arguments
+- `rs` : ResultSet
+- `outcomes` : ADRIA Outcome Mapping results
+- `factors` : The factors of interest to display
+- `opts` : Additional figure customization options
+- `fig_opts` : Additional options to pass to adjust Figure creation
+  See: https://docs.makie.org/v0.19/api/index.html#Figure
+- `axis_opts` : Additional options to pass to adjust Axis attributes
+  See: https://docs.makie.org/v0.19/api/index.html#Axis
+
+# Returns
+Makie figure
+"""
+function ADRIA.viz.outcome_map!(
+    g::Union{GridLayout,GridPosition},
+    rs::ResultSet,
+    outcomes::NamedDimsArray,
+    factors::Vector{Symbol};
+    opts::Dict=Dict(),
+    axis_opts::Dict=Dict(),
+)
+    # TODO: Clean up and compartmentalize as a lot of code here are duplicates of those
+    #       found in `rsa()`
+    n_factors::Int64 = length(factors)
+    if n_factors > 30
+        ArgumentError("Too many factors to plot. Maximum number supported is 30.")
+    end
+
+    n_rows, n_cols = _calc_gridsize(n_factors)
+
+    xlabel = pop!(axis_opts, :xlabel, "Factor Value")
+    ylabel = pop!(axis_opts, :ylabel, "Outcome")
+
+    if :title in keys(axis_opts)
+        title_val = pop!(axis_opts, :title)
+    end
+
+    # min_step = (1 / 0.05)
+    # color_weight = min((1.0 / (length(factors) / min_step)), 0.6)
+
+    # Color by component
+    ms = model_spec(rs)
+    foi = ms.fieldname .∈ [factors]
+    all_comps = ms[foi, :component]
+    f_names = ms[foi, :fieldname]
+    h_names = ms[foi, :name]
+    bounds = ms[foi, :bounds]
+
+    # Hacky special case handling for SSP/RCP
+    if :RCP in factors || :SSP in factors
+        loc = first(findall((factors .== :RCP) .|| (factors .== :SSP)))
+        insert!(all_comps, loc, "EnvironmentalLayer")
+        insert!(f_names, loc, "RCP")
+        insert!(h_names, loc, "SSP/RCP")
+        insert!(bounds, loc, (1, length(unique(rs.inputs.RCP))))
+    end
+
+    bin_slices, factor_list, CIs = axiskeys(outcomes)
+    b_slices = parse.(Float64, bin_slices)
+
+    if any(f_names .== :guided)
+        fv_labels = [
+            "unguided", "cf", last.(split.(string.(ADRIA.decision.mcda_methods()), "."))...
+        ]
+    end
+    curr::Int64 = 1
+    axs = Axis[]
+    for r in 1:n_rows
+        for c in 1:n_cols
+            f_name = Symbol(factors[curr])
+            f_vals = rs.inputs[:, f_name]
+
+            if f_name == :guided
+                fv_s = collect(1:length(fv_labels))
+            else
+                fv_s = round.(quantile(f_vals, b_slices), digits=2)
+            end
+
+            ax::Axis = Axis(
+                g[r, c]; title=h_names[f_names .== factors[curr]][1], axis_opts...
+            )
+
+            band!(
+                ax,
+                fv_s[.!ismissing.(outcomes(; factors=f_name, CI=:lower))],
+                collect(skipmissing(outcomes(; factors=f_name, CI=:lower))),
+                collect(skipmissing(outcomes(; factors=f_name, CI=:upper))),
+            )
+            scatterlines!(ax, fv_s, outcomes(; factors=f_name, CI=:mean); markersize=15)
+
+            if f_name == :guided
+                ax.xticks = (fv_s, fv_labels)
+                ax.xticklabelrotation = pi / 4
+            end
+
+            push!(axs, ax)
+            curr += 1
+
+            if curr > n_factors
+                break
+            end
+        end
+    end
+
+    if n_factors > 1
+        linkyaxes!(axs...)
+        Label(g[n_rows + 1, :]; text=xlabel, fontsize=24)
+        Label(g[:, 0]; text=ylabel, fontsize=24, rotation=pi / 2)
+
+        if @isdefined(title_val)
+            Label(g[0, :]; text=title_val, fontsize=32)
+        end
+    else
+        axs[1].xlabel = xlabel
+        axs[1].ylabel = ylabel
+
+        if @isdefined(title_val)
+            axs[1].title = title_val
+        end
+    end
+
+    try
+        # Clear empty figures
+        trim!(g)
+    catch err
+        if !(err isa MethodError)
+            # GridPosition plots a single figure so does
+            # not need empty figures to be cleared
+            # If any other error is encountered, something else happened.
+            rethrow(err)
+        end
+    end
+
+    return g
+end
+function ADRIA.viz.outcome_map(
+    rs::ResultSet,
+    si::NamedDimsArray,
+    factors::Vector{Symbol};
+    opts::Dict=Dict(),
+    fig_opts::Dict=Dict(),
+    axis_opts::Dict=Dict(),
+)
+    f = Figure(; fig_opts...)
+    g = f[1, 1] = GridLayout()
+    ADRIA.viz.outcome_map!(g, rs, si, factors; opts=opts, axis_opts=axis_opts)
 
     return f
 end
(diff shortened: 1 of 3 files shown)


src/analysis/sensitivity.jl
```diff
@@ -440,8 +468,10 @@ function rsa(X::DataFrame, y::AbstractVector{<:Real}; S::Int64=10)::NamedDimsArr
         NamedDimsArray(r_s; bins=string.(seq[2:end]), factors=Symbol.(names(X)))
     )
 end
-function rsa(rs::ResultSet, y::AbstractVector{<:Real}; S::Int64=10)::NamedDimsArray
-    return rsa(rs.inputs, vec(y); S=S)
+function rsa(
+    rs::ResultSet, y::AbstractVector{<:Real}, model_spec::DataFrame; S::Int64=10
+)::NamedDimsArray
+    return rsa(rs.inputs, vec(y), model_spec; S=S)
 end
```
ConnectedSystems, line 475: `rs` holds a copy of model spec, so I'm not sure why we need it as an argument here.

```suggestion
function rsa(
    rs::ResultSet, y::AbstractVector{<:Real}; S::Int64=10
)::NamedDimsArray
    return rsa(rs.inputs, vec(y), rs.model_spec; S=S)
end
```

src/analysis/sensitivity.jl
```diff
@@ -401,18 +402,45 @@ ADRIA.sensitivity.rsa(X, y; S=10)
...
     N, D = size(X)
-    seq = collect(0.0:(1/S):1.0)
 
     X_di = @MVector zeros(N)
+    sel = trues(N)
+    factors = Symbol.(names(X))
+
+    factors_to_assess = model_spec.fieldname .∈ [factors]
+    foi_attributes = Symbol[:fieldname, :ptype, :lower_bound, :upper_bound]
+    foi_spec = model_spec[factors_to_assess, foi_attributes]
```
ConnectedSystems, line 416: This doesn't work because the factor names in the dataframe are strings, but `factors` are cast to Symbols.

Have you checked if this works, Rose?

src/analysis/sensitivity.jl
```diff
@@ -401,18 +402,45 @@ ADRIA.sensitivity.rsa(X, y; S=10)
...
+
+    foi_cat = (foi_spec.ptype .== "categorical")
+    if any(foi_cat)
+        max_bounds = maximum(
+            foi_spec[foi_cat, :upper_bound] .- foi_spec[foi_cat, :lower_bound]
+        )
+        S = round(Int64, max(S, max_bounds))
+    end
```
ConnectedSystems, line 424: Ideally common/duplicate code that are shared across different functions/methods should be separated out into their own functions.

I'm not going to check this in further detail right now as it looks like it's just copy/pasted from the `outcome_map` method. I'm not sure it will work without any more changes.

Have you checked?

ConnectedSystems (changes requested):
Please double check for correctness

src/analysis/sensitivity.jl
Rosejoycrocker: Hi Takuya, yes I did check. I found that the factors from the domain model spec are symbols but the rs model spec are strings - is this intended? Or do I need to rebase again. I do check my work

ConnectedSystems: > Hi Takuya, yes I did check. I found that the factors from the domain model spec are symbols but the rs model spec are strings - is this intended? Or do I need to rebase again. I do check my work

Okay, then it's a mismatch between how `load_results()` reads in the stored copy of the model specification, and how `Domain` treats it. 

Remember that the `ResultSet` is intended to be a standalone snapshot of what was used to run scenarios. You will not always have the `Domain` available to generate the model spec from.

Rosejoycrocker: Ok I will change the usage.

src/analysis/sensitivity.jl
Rosejoycrocker: As I said, I have checked and would not have submitted a PR if I hadn't. I will group the common code into a function.

Rosejoycrocker:
I've addressed your comments so it expects the result set model_spec now. All tests passing and plotting fine on my end.

src/analysis/sensitivity.jl
ConnectedSystems: No I think there's a misunderstanding here. We have to support both.

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,38 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+"""
+    _get_factor_spec(model_spec::DataFrame, factors::Vector{String})
+
+Get model spec for specified factors.
+"""
+function _get_factor_spec(model_spec::DataFrame, factors::Vector{Symbol})
+    factors_to_assess = model_spec.fieldname .∈ [factors]
+    foi_attributes = Symbol[:fieldname, :ptype, :lower_bound, :upper_bound]
+    return model_spec[factors_to_assess, foi_attributes]
+end
```
ConnectedSystems, line 44: This doesn't return a spec - I'd say this is the factor details, its domain, or factor range (parameter type + lower/upper bounds).

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,38 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+
+"""
+    _set_cat_S(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+
+Get number of binnings for categorical variables.
+"""
+function _set_cat_S(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+    max_bounds = maximum(foi_spec[foi_cat, :upper_bound] .- foi_spec[foi_cat, :lower_bound])
+    return round(Int64, max(S, max_bounds))
+end
```
ConnectedSystems, line 54: Function name is a bit cryptic - I'm not sure I understand what a `cat_S` is.

It also does not set anything?

What about `_category_bins()`?

src/analysis/sensitivity.jl
```diff
@@ -345,8 +377,8 @@ function tsa(rs::ResultSet, y::AbstractMatrix{<:Real})::NamedDimsArray
 end
 
 """
-    rsa(X::DataFrame, y::Vector{<:Real}; S=10)::NamedDimsArray
-    rsa(rs::ResultSet, y::AbstractArray{<:Real}; S::Int64=10)::NamedDimsArray
+    rsa(X::DataFrame, y::Vector{<:Real}, model_spec::DataFrame; S=10)::NamedDimsArray
+    rsa(rs::ResultSet, y::AbstractArray{<:Real}, model_spec::DataFrame; S::Int64=10)::NamedDimsArray
```
ConnectedSystems, line 381:
```suggestion
    rsa(rs::ResultSet, y::AbstractArray{<:Real}; S::Int64=10)::NamedDimsArray
```

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,38 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+    factors_to_assess = model_spec.fieldname .∈ [factors]
+    foi_attributes = Symbol[:fieldname, :ptype, :lower_bound, :upper_bound]
+    return model_spec[factors_to_assess, foi_attributes]
+end
+
+"""
+    _set_cat_S(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+
+Get number of binnings for categorical variables.
+"""
```
ConnectedSystems, line 50: Should detail the arguments and what they refer to

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,38 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+function _set_cat_S(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+    max_bounds = maximum(foi_spec[foi_cat, :upper_bound] .- foi_spec[foi_cat, :lower_bound])
+    return round(Int64, max(S, max_bounds))
+end
+
+"""
+    _get_cat_quantile(foi_spec::DataFrame, fact_t::String, steps::Vector{Float64})
+
+Get quantile for categorical variable, fact_t.
+"""
```
ConnectedSystems, line 60: As above - detail arguments and what they relate to.

src/io/result_io.jl
```diff
@@ -407,6 +407,10 @@ function load_results(result_loc::String)::ResultSet
     # Model specification
     model_spec = CSV.read(joinpath(result_loc, MODEL_SPEC, "model_spec.csv"), DataFrame; comment="#")
 
+    # Standardize fieldnames to Symbol
+    # TODO: Match all other column data types with original model spec
+    model_spec.fieldname .= Symbol.(model_spec.fieldname)
+
```
ConnectedSystems, line 413: I've put in this is a hacky fix to be replaced with a more robust approach later.

ConnectedSystems (changes requested):
A few minor changes requested

src/analysis/sensitivity.jl
Rosejoycrocker: ah ok, sorry I thought you meant you wanted to support the result set one only

src/analysis/sensitivity.jl
Rosejoycrocker: By `cat_S` I meant categorical S, but I'll change to `_category_bins`

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,38 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+function _get_cat_quantile(foi_spec::DataFrame, fact_t::Symbol, steps::Vector{Float64})
+    fact_idx = foi_spec.fieldname .== fact_t
+    lb = foi_spec.lower_bound[fact_idx][1]
+    ub = foi_spec.upper_bound[fact_idx][1]
+    return round.(quantile(lb:ub, steps)) .- 1
+end
```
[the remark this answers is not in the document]
Rosejoycrocker: I've changed it now to `fact_cat` as in categorical factor

Rosejoycrocker:
I think this is ready for review again @ConnectedSystems . All tests passing and both usages (rs input & X,y input) for outcome_map and rsa working. Plotting also working on my end.

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -155,15 +155,14 @@ function ADRIA.viz.tsa(
...
 
 Plot regional sensitivities of up to 30 factors.
 
 # Arguments
 - `rs` : ResultSet
 - `si` : Results from ADRIA regional sensitivity analysis
-- `opts` : Additional figure customization options
 - `fig_opts` : Additional options to pass to adjust Figure creation
   See: https://docs.makie.org/v0.19/api/index.html#Figure
 - `axis_opts` : Additional options to pass to adjust Axis attributes
```
ConnectedSystems, line 168: The `opts` argument - at least for `rsa()` - is unused but purposefully put in to maintain a consistent API as most other plotting methods expects `opts`.

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -530,8 +539,7 @@ function ADRIA.viz.outcome_map!(
     g::Union{GridLayout,GridPosition},
     rs::ResultSet,
     outcomes::NamedDimsArray,
-    factors::Vector{String};
-    opts::Dict=Dict(),
+    factors::Vector{Symbol};
     axis_opts::Dict=Dict(),
 )
```
ConnectedSystems, line 544: Same comment as above - `opts` was in there so there is a consistent interface for users to interact with.

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,52 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+
+"""
+    _category_bins(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+
+Get number of binnings for categorical variables.
+
+# Arguments
+- S : Number of bins
+- foi_spec : Model specification for factors of interest
+- foi_cat : Vector of length `foi_cat.fieldname` which contains true where the factor is categorical and false otherwise
```
ConnectedSystems, line 58:
```suggestion
# Arguments
- `S` : Number of bins
- `foi_spec` : Model specification for factors of interest
- `foi_cat` : Vector of length `foi_cat.fieldname` which contains `true` where the factor is categorical and `false` otherwise
```

On `foi_cat`: to me this would mean "factor of interest category", but it holds true/false indicating whether a factor is categorical or not. How about calling this `is_cat` ("is categorical") ?

ConnectedSystems: Actually on second thought - this seems over-complicated to me.

Why have both the dataframe and the selector as arguments, when the desired subset could be passed in right away?

```julia
_category_bins(S, foi_spec[is_cat, :])
```

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,52 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+- steps : Number of steps for defining binnings
+"""
+function _get_cat_quantile(foi_spec::DataFrame, fact_cat::Symbol, steps::Vector{Float64})
+    fact_idx = foi_spec.fieldname .== fact_cat
+    lb = foi_spec.lower_bound[fact_idx][1]
+    ub = foi_spec.upper_bound[fact_idx][1]
+    return round.(quantile(lb:ub, steps)) .- 1
+end
```
ConnectedSystems, line 80: Generally the same comment as above, but I don't understand this:

If `fact_cat` (which hasn't been updated in the argument list) is a vector of true/false, then how does the below work?

```julia
fact_idx = foi_spec.fieldname .== fact_cat
```

EDIT:
Okay, I see, `fact_cat` is the target factor name. In which case the variable name doesn't make sense to me.

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,52 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+
+"""
+    _get_cat_quantile(foi_spec::DataFrame, fact_cat::String, steps::Vector{Float64})
+
+Get quantile for categorical variable, fact_cat.
+
+# Arguments
+- foi_spec : Model specification for factors of interest
+- foi_cat : Vector of length `foi_cat.fieldname` which contains true where the factor is categorical and false otherwise
+- steps : Number of steps for defining binnings
```
ConnectedSystems, line 73: Arguments need to be updated and backticks added

src/analysis/sensitivity.jl
```diff
@@ -376,6 +422,7 @@ Note: Values of type `missing` indicate a lack of samples in the region.
 - `rs` : ResultSet
 - `X` : scenario specification
 - `y` : scenario outcomes
+- `model_spec` : Model specification, such as ADRIA.model_spec(domain)
```
ConnectedSystems, line 425:
```suggestion
- `model_spec` : Model specification, as extracted by `ADRIA.model_spec()` or from a `ResultSet`
```

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -223,18 +227,25 @@ function ADRIA.viz.rsa!(
...
+            if f_name == :guided
+                fv_s = collect(1:length(fv_labels))
+            else
+                fv_s = round.(quantile(f_vals, b_slices), digits=2)
+            end
 
             ax::Axis = Axis(
                 g[r, c],
-                title=h_names[f_names.==factors[curr]][1];
+                title=h_names[f_names .== factors[curr]][1],
```
ConnectedSystems, line 240:
```suggestion
                title=h_names[f_names .== f_name][1],
```

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -223,18 +227,25 @@ function ADRIA.viz.rsa!(
...
                 axis_opts...
             )
 
-            scatterlines!(ax, fv_s, si(factors=Symbol(factors[curr])), markersize=15)
+            scatterlines!(ax, fv_s, si(; factors=Symbol(f_name)); markersize=15)
```
ConnectedSystems, line 244: `f_name` already a Symbol

```suggestion
            scatterlines!(ax, fv_s, si(; factors=f_name); markersize=15)
```

ConnectedSystems (changes requested):
A few things to fix up

src/analysis/sensitivity.jl
Rosejoycrocker: It's for factor categoical, but I'll think of another name

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,52 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+function _get_factor_spec(model_spec::DataFrame, factors::Vector{Symbol})
+    factors_to_assess = model_spec.fieldname .∈ [factors]
+    foi_attributes = Symbol[:fieldname, :ptype, :lower_bound, :upper_bound]
+    return model_spec[factors_to_assess, foi_attributes]
+end
+
+"""
+    _category_bins(S::Int64, foi_spec::DataFrame, foi_cat::BitVector)
+
+Get number of binnings for categorical variables.
```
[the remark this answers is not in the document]
Rosejoycrocker: Your right, I think I started with a different sentence and didn't change binnings

src/analysis/sensitivity.jl
Rosejoycrocker: Good point!

src/analysis/sensitivity.jl
Rosejoycrocker: Sorry, not sure why I missed this

Rosejoycrocker:
I think I've addressed your comments @ConnectedSystems , functions and plotting tested and tests all passing

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,51 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
 end
 
+"""
+    _get_factor_spec(model_spec::DataFrame, factors::Vector{String})
+
+Get model spec for specified factors.
+
+# Arguments 
+- `model_spec` : Model specification, as extracted by `ADRIA.model_spec(domain)` or from a `ResultSet`
+- `factors` : Parameters considered for sensitivity analysis
```
ConnectedSystems, line 42:
```suggestion
- `factors` : Factors considered for sensitivity analysis
```

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,51 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
     return sqrt(n) * ks.δ
 end
 
+"""
+    _get_factor_spec(model_spec::DataFrame, factors::Vector{String})
```
ConnectedSystems, line 36:
```suggestion
    _get_factor_spec(model_spec::DataFrame, factors::Vector{Symbol})
```

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -155,8 +155,8 @@ function ADRIA.viz.tsa(
 end
 
 """
-    ADRIA.viz.rsa(rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
-    ADRIA.viz.rsa!(f::Union{GridLayout,GridPosition}, rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; opts, axis_opts)
+    ADRIA.viz.rsa(rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
+    ADRIA.viz.rsa!(f::Union{GridLayout,GridPosition}, rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; axis_opts::Dict=Dict())
```
ConnectedSystems, line 159:
```suggestion
    ADRIA.viz.rsa(rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
    ADRIA.viz.rsa!(f::Union{GridLayout,GridPosition}, rs::ResultSet, si::NamedDimsArray, factors::Vector{String}; opts::Dict=Dict(), axis_opts::Dict=Dict())
```

src/analysis/sensitivity.jl
```diff
@@ -32,6 +32,51 @@ function ks_statistic(ks::ApproximateKSTest)::Float64
...
+"""
+function _category_bins(S::Int64, foi_spec::DataFrame)
+    max_bounds = maximum(foi_spec.upper_bound .- foi_spec.lower_bound)
+    return round(Int64, max(S, max_bounds))
+end
+
+"""
+    _get_cat_quantile(foi_spec::DataFrame, factor_c::String, steps::Vector{Float64})
+
+Get quantile for categorical variable, factor_c.
```
ConnectedSystems, line 67: Re the variable name, the larger point here is that the description in the docstring should focus on what the function does. It is very rare for a variable to be called out like this.

ConnectedSystems, line 67:
```suggestion
Get quantile value for a given categorical variable.
```

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -246,7 +257,7 @@ function ADRIA.viz.rsa!(
 
     linkyaxes!(axs...)
     Label(g[end+1, :], text=xlabel, fontsize=32)
-    Label(g[1:end-1, 0], text=ylabel, fontsize=32, rotation=pi / 2)
+    Label(g[1:(end - 1), 0]; text=ylabel, fontsize=32, rotation=xtick_rot)
```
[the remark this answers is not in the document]
ConnectedSystems: Turns out this was the y-axis label, putting this at a 45-degree angle is wrong.

PR #602, opened 2023-11-24
Separate shading and fogging timing parameters

Rosejoycrocker:
As requested in #598


src/scenario.jl
```diff
@@ -478,9 +480,10 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     α = 0.95
     decay = α .^ (1:Int64(param_set("plan_horizon"))+1)
 
-    # Years at which to reassess seeding site selection
+    # Years at which seeding/shading/fogging occurs
     seed_decision_years = fill(false, tf)
     shade_decision_years = fill(false, tf)
+    fog_decision_years = fill(false, tf)
```
ConnectedSystems, line 486: Is the modified comment correct, or has the conceptualization changed?

My understanding was that:

- `*_freq` controls how often location selection occurs (every year, every 3rd year, etc)
- `*_decision_years` then speak to which years in the simulation location selection occurs
- It does not relate to years interventions are deployed

If `fog_freq` is 3, it re-evaluates where interventions should occur every third year.
So `fog_decision_years` holds [false, false, true, false, false, true, false, false, true, ...]

Interventions still occur every year, it's just that _where_ interventions occur changes every third year.

Is this not correct?

Rosejoycrocker: You're correct, I think maybe this comment was originally for `in_seed_years` , `in_shade_years` etc. definitions. I've moved it to where there are defined and added a comment to describe `*_decision_years`

src/metrics/performance.jl
```diff
@@ -117,13 +117,23 @@ This is referred to as \$F\$.
 
 # Returns
 Matrix of `s * 7`, where `s` is the number of scenarios and columns are:
-`N_seed_TA`, `N_seed_CA`, `N_seed_SM`, `fogging`, `SRM`, `seed_years`, `shade_years`
+`N_seed_TA`, `N_seed_CA`, `N_seed_SM`, `fogging`, `SRM`, `seed_years`, `shade_years`, `fog_years`
```
ConnectedSystems, line 120: Number of columns needs to be updated - `s * 8`

src/scenario.jl
```diff
@@ -432,14 +432,16 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     n_species::Int64 = domain.coral_growth.n_species
     n_groups::Int64 = domain.coral_growth.n_groups
 
-    # years to start seeding/shading
+    # years to start seeding/shading/fogging
     seed_start_year::Int64 = param_set("seed_year_start")
     shade_start_year::Int64 = param_set("shade_year_start")
+    fog_start_year::Int64 = param_set("fog_year_start")
 
     fogging::Real = param_set("fogging")  # percent reduction in bleaching mortality through fogging
```
ConnectedSystems, line 440:
```suggestion
    fogging::Real = param_set("fogging")  # proportion of bleaching mortality reduction through fogging
```

Not caused by any changes in this PR but noticed this says "percent", implying values are $\in$ [0, 100].
Suggest changing to "proportion" to avoid this as values should be $\in$ [0, 1].

src/metrics/performance.jl
```diff
@@ -117,13 +117,23 @@ This is referred to as \$F\$.
...
+    interv_cols=Symbol.([
+        "N_seed_TA",
+        "N_seed_CA",
+        "N_seed_SM",
+        "fogging",
+        "SRM",
+        "seed_years",
+        "shade_years",
+        "fog_years",
+    ]),
```
ConnectedSystems, line 135:
```suggestion
    interv_cols=[
        :N_seed_TA,
        :N_seed_CA,
        :N_seed_SM,
        :fogging,
        :SRM,
        :seed_years,
        :shade_years,
        :fog_years,
    ],
```

src/scenario.jl
```diff
@@ -626,8 +638,13 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
 
-        in_shade_years = (shade_start_year <= tstep) && (tstep <= (shade_start_year + shade_years - 1))
-        in_seed_years = (seed_start_year <= tstep) && (tstep <= (seed_start_year + seed_years - 1))
+        # Years at which seeding/fogging/shading occurs
+        in_fog_years =
+            (fog_start_year <= tstep) && (tstep <= (fog_start_year + fog_years - 1))
+        in_shade_years =
+            (shade_start_year <= tstep) && (tstep <= (shade_start_year + shade_years - 1))
+        in_seed_years =
+            (seed_start_year <= tstep) && (tstep <= (seed_start_year + seed_years - 1))
```
ConnectedSystems, line 647: While you're at these changes, could you check if this works?

From memory there was something preventing me from implementing it this way

```suggestion
        in_fog_years =
            fog_start_year <= tstep <= (fog_start_year + fog_years - 1)
        in_shade_years =
            shade_start_year <= tstep <= (shade_start_year + shade_years - 1)
        in_seed_years =
            seed_start_year <= tstep <= (seed_start_year + seed_years - 1)
```

src/scenario.jl
```diff
@@ -671,19 +689,20 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
             # Log site ranks
             # First col only holds site index ids so skip (with 2:end)
             site_ranks[tstep, rankings[:, 1], :] = rankings[:, 2:end]
-        elseif seed_corals && (in_seed_years || in_shade_years)
+        elseif seed_corals && (in_seed_years || in_fog_years)
```
ConnectedSystems, line 692: Hang on - doesn't this mean unguided selection only occurs when there is something to seed?

What if `fogging > 0`?

ConnectedSystems: I think this should be changed to:

```
# somewhere above...
apply_fogging = fogging > 0.0

        elseif (seed_corals && in_seed_years) || (apply_fogging && in_fog_years)
```

src/scenario.jl
```diff
@@ -695,7 +714,7 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
         end
 
         # Fog selected locations
-        if (fogging > 0.0) && in_shade_years && has_shade_sites
+        if (fogging > 0.0) && in_fog_years && has_fog_sites
```
ConnectedSystems, line 717: If `apply_fogging = fogging > 0.0` is adopted:

```suggestion
        if apply_fogging && in_fog_years && has_fog_sites
```

src/scenario.jl
```diff
@@ -478,9 +480,10 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
     α = 0.95
     decay = α .^ (1:Int64(param_set("plan_horizon"))+1)
 
-    # Years at which to reassess seeding site selection
+    # Years at which seeding/shading/fogging decisions are re-evaluated
```
ConnectedSystems, line 483:
```suggestion
    # Years at which intervention locations are re-evaluated
```

src/scenario.jl
```diff
@@ -671,19 +689,20 @@ function run_model(domain::Domain, param_set::NamedDimsArray, corals::DataFrame,
...
+                seed_decision_years[tstep],
+                fog_decision_years[tstep],
                 n_site_int, vec(leftover_space_m²), depth_priority)
 
             site_ranks[tstep, seed_locs, 1] .= 1.0
             site_ranks[tstep, fog_locs, 2] .= 1.0
         end
 
-        has_shade_sites::Bool = !all(fog_locs .== 0)
+        has_fog_sites::Bool = !all(fog_locs .== 0)
```
ConnectedSystems, line 705: Missed opportunity to update to `has_fog_locs` ?

src/scenario.jl
Rosejoycrocker: Yeah I think `in_*_timeframe` stands out more from the `*_decision_years` parameters and is more informative

I'll check what you've suggested

src/scenario.jl
Rosejoycrocker: You're right!

PR #607, opened 2023-11-27
Fix proportional_adjustment vector conversion computational error

Zapiano:
As in title.


ConnectedSystems (approved):
Thanks for fixing, looks fine to me.

PR #608, opened 2023-11-28
Add tests for figure creation

ConnectedSystems:
As in title.

Requires #598 to be merged in first, followed by a rebase onto main.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -128,7 +128,10 @@ Zarr = "0.7, 0.8, 1, 0.9"
 julia = "1"
 
 [extras]
+WGLMakie = "276b4fcb-3e11-5398-bf8b-a0c2d153d008"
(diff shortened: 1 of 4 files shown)


Zapiano:
This seems to be doing much more than just adding tests for figure creation 🤔

ConnectedSystems:
You'll have to forgive the unrelated changes that mark struct fields as constant - I didn't want to submit a PR for three lines.

Zapiano (approved):
I really think we should put these two functions (and call them) in a separate file called `viz.jl` or something like that. Besides being more organized, it'll make it easier in the future if we want to run just this test apart from the rest. 

But other than that, lgtm

ConnectedSystems:
I agree that would be cleaner. This was a quick implementation. Let's iterate towards that perfect solution.

PR #610, opened 2023-11-28
Fix bug where cover instead of leftover space was being used in stand alone site selection

Rosejoycrocker:
Changes in MCDAvars mean leftover space in $m^2$ is now an input to MCDAvars instead of cover (introduced in #547). Cover was still being used as an input in the `MCDA_vars` usage in `rank_locations`, causing incorrect outputs in testing.

### DIFF
--- a/src/Domain.jl
+++ b/src/Domain.jl
@@ -232,7 +232,7 @@ function n_locations(domain::Domain)::Int64
 end
 
 """
-    relative_leftover_space(loc_coral_cover::Matrix{Float64})::Matrix{Float64}
+    relative_leftover_space(loc_coral_cover::Union{Matrix{Float64},Vector{Float32}})::Union{Matrix{Float64},Vector{Float32}}
 
 Get proportion of leftover space, given site_k and proportional cover on each site, summed
 over species.
@@ -243,7 +243,9 @@ over species.
 # Returns
 Leftover space ∈ [0, 1]
 """
-function relative_leftover_space(loc_coral_cover::Matrix{Float64})::Matrix{Float64}
+function relative_leftover_space(
+    loc_coral_cover::Union{Matrix{Float64},Vector{Float32}}
+)::Union{Matrix{Float64},Vector{Float64}}
     return max.(1.0 .- loc_coral_cover, 0.0)
 end
 
--- a/src/Domain.jl
+++ b/src/Domain.jl
@@ -232,7 +232,7 @@ function n_locations(domain::Domain)::Int64
 end
 
 """
-    relative_leftover_space(loc_coral_cover::Union{Matrix{Float64},Vector{Float32}})::Union{Matrix{Float64},Vector{Float32}}
+    relative_leftover_space(loc_coral_cover::AbstractArray)::AbstractArray
 
 Get proportion of leftover space, given site_k and proportional cover on each site, summed
 over species.
@@ -244,8 +244,8 @@ over species.
 Leftover space ∈ [0, 1]
 """
 function relative_leftover_space(
-    loc_coral_cover::Union{Matrix{Float64},Vector{Float32}}
-)::Union{Matrix{Float64},Vector{Float64}}
+    loc_coral_cover::AbstractArray,
+)::AbstractArray
     return max.(1.0 .- loc_coral_cover, 0.0)
 end
 
--- a/src/decision/location_selection.jl
+++ b/src/decision/location_selection.jl
@@ -130,7 +130,7 @@ function rank_locations(
             scen,
             considered_sites,
             relative_leftover_space(
-                AxisKeys.keyless(NamedDims.unname(sum_cover[scen_idx, :]))
+                collect(sum_cover[scen_idx, :]),
             ) .* k_area_locs,
             area_to_seed,
             summary_stat_env(wave_scens[:, :, target_wave_scens], (:timesteps, :scenarios)),
--- a/src/decision/location_selection.jl
+++ b/src/decision/location_selection.jl
@@ -10,7 +10,7 @@ MCDA method.
 
 # Arguments
 - `domain` : The geospatial domain to assess
-- `sum_cover` :  Absolute coral cover at each location
+- `sum_cover` :  Relative coral cover at each location
 - `mcda_vars` : Parameters for MCDA
 - `guided` : ID of MCDA algorithm to apply
 
@@ -66,7 +66,7 @@ Return location ranks for a given domain and scenarios.
 # Arguments
 - `domain` : The geospatial domain locations were selected from
 - `scenarios` : Scenario specification
-- `sum_cover` : Matrix[n_scenarios ⋅ n_sites] containing the total coral cover at each
+- `sum_cover` : Matrix[n_scenarios ⋅ n_sites] containing the total relative coral cover at each
     location, for each scenario
 - `area_to_seed` : Area of coral to be seeded at each time step in km²
 - `agg_func` : Aggregation function to apply, e.g `ranks_to_frequencies` or
--- a/src/decision/location_selection.jl
+++ b/src/decision/location_selection.jl
@@ -116,6 +116,8 @@ function rank_locations(
         target_site_ids = collect(1:length(domain.site_ids))
     end
 
+    leftover_space_scens = relative_leftover_space(sum_cover.data) .* k_area_locs'
+
     for (scen_idx, scen) in enumerate(eachrow(scenarios))
         depth_criteria = within_depth_bounds(
             domain.site_data.depth_med,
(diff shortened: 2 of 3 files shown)


ConnectedSystems:
@Zapiano would you mind doing the review instead? 

Before you do so, just want to check @Rosejoycrocker :

The docstring for `rank_locations()` states that `sum_cover` relates to total cover so from what I understand absolute cover should be passed in as the argument, not relative cover.

Looking at the commit history, it's been like this for a long time (since at least #431). Harder to track what happened before that as you moved/recreated the file before then - but we could dig into it to see what was expected before that PR.

Either all the example usage and tests need to be updated to reflect the above, or the docstring for `rank_locations()` has to be corrected.

src/decision/location_selection.jl
```diff
@@ -124,7 +125,14 @@ function rank_locations(
         depth_priority = findall(depth_criteria)
 
         considered_sites = target_site_ids[findall(in(depth_priority), target_site_ids)]
-        mcda_vars_temp = DMCDA_vars(domain, scen, considered_sites,  sum_cover[scen_idx, :], area_to_seed,
+        mcda_vars_temp = DMCDA_vars(
+            domain,
+            scen,
+            considered_sites,
+            relative_leftover_space(
+                AxisKeys.keyless(NamedDims.unname(sum_cover[scen_idx, :]))
```
ConnectedSystems, line 133: > Sorry, its not the argument to rank_locations which is wrong, it's the argument to MCDA_vars in rank_locations, which now takes leftover_space, rather than cover. rank_locations still takes cover. The update to MCDA_vars was done when we added a depth weight, I believe. I'll make this clearer in my description.

Sorry, I'm still not following - but I think what you're raising is a separate issue.

If `sum_cover` is in absolute units (e.g., meters squared) then this change to L 132 of `location_selection.jl` 
is wrong because `relative_leftover_space` expects relative cover.

https://github.com/open-AIMS/ADRIA.jl/blob/62c5deb4505a424233f2c724117767ae27a059ba/src/Domain.jl#L234-L249

Rosejoycrocker: I think you've spotted another issue, the input to `_location_selection` should now be relative cover because it is use in `connectivity_strength`, so the input to `rank_locations` should also be relative cover ( and was originally I think), I'll add changes to the docs

Zapiano:
I can review this, no worries, just let me know when it's ready!

src/decision/location_selection.jl
ConnectedSystems: >  the input to _location_selection should now be relative cover because it is used in connectivity_strength, so the input to rank_locations should also be relative cover

`connectivity_strength` uses absolute cover. I don't think it should use relative cover, wouldn't create misleading indications of connectivity, as it places greater weight on small locations with high cover?

The transference probability matrix  (TP) is expected to be weighted by $k$ area based on the docstring.

I see yet another issue here - the process of loading the domain for ADRIA uses the TP data directly, whereas loading ReefMod Engine datasets applies the weighting based on k area and absolute cover.

Rosejoycrocker: Right, I just copied the usage in `scenario.jl`, whith uses `TP` weighted with `k_area` for the first argument and relative cover for the second. Should the second argument be absolute cover then?

ConnectedSystems: Okay, could you please check which it should actually be then?

Personally I think my previous change in `scenario.jl` is incorrect.

Rosejoycrocker: Sure, I'll check :)

Rosejoycrocker: I think relative cover in `connectivity_strength` is correct because here it is multiplied byt the k_area weighted TP matrix- https://github.com/open-AIMS/ADRIA.jl/blob/62c5deb4505a424233f2c724117767ae27a059ba/src/ecosystem/connectivity.jl#L218

ConnectedSystems: Okay, so it should be relative cover and the previous change in `scenario.jl` is correct? In which case the docstring needs to be updated

Rosejoycrocker: Ok cool, I'll update the doc strings, then I think it's ready for review @Zapiano

Zapiano:
@Rosejoycrocker just confirming: is this PR ready for review?

src/decision/location_selection.jl
```diff
@@ -124,7 +125,14 @@ function rank_locations(
         depth_priority = findall(depth_criteria)
 
         considered_sites = target_site_ids[findall(in(depth_priority), target_site_ids)]
-        mcda_vars_temp = DMCDA_vars(domain, scen, considered_sites,  sum_cover[scen_idx, :], area_to_seed,
+        mcda_vars_temp = DMCDA_vars(
+            domain,
+            scen,
+            considered_sites,
+            relative_leftover_space(
```
Zapiano, line 132: I personally am very against calling one function as argument to another the way is happening here because of readability. I'm not sure how much performance you lose by assigning this to a variable instead since Julia will have to allocate memory to do this calculation anyways (I THINK). 

@ConnectedSystems thoughts?

Zapiano (approved):
Just one comment. Other than that, LGTM.

src/decision/location_selection.jl
Rosejoycrocker: We could change MCDAvars but the variable will have to be stripped at somepoint because it is put into the decision matrix

ConnectedSystems: I'd then ask: why doesnt the decision matrix use named dimensions? Relying on roe/column indices is a potential source of bugs

Rosejoycrocker: This was one of the changes I was pushing for in the PR to make MCDA generic, but that got broken into smaller PRs and this aspect still hasn't been done. Agree it's a priority, but probably not for this PR.

Actually, I realise now it's because `relative_leftover_space` expects a vector or Matrix that I have to strip the NmaedDimsArray here

Rosejoycrocker: Using `Matrix()` errors because it's a row, so I could use Array or collect instead. I've gone with collect, but let me know if Array is better.

ConnectedSystems: If it is a row then it's Vector is more explicit than Array, no?

Rosejoycrocker: As in `vec()`? - that is still a NamedDimsArray though with a single named dimension

Zapiano: A few comments:

1. Maybe `relative_leftover_space` should expect/return an `AbstractArray`, this way it will work with Vectors, Matrices, NamedDimsArrays, etc. Also, this way, in the future, when you make the decision matrix use named dimensions, this function won't have to be changed.
2. Suppose you have `nda = NamedDimsArray(rand(2,2), x=[1,2], y=[3,4])`, you can do `nda.data.data` to retrieve the raw matrix, and you can also do `nda[1,:].data.data` to get your vector. I think this could be a temporary solution while we wait for the decision matrix and `DMCDA_vars` changes, so maybe creating an issue to come back here and change that later would be interesting?

src/Domain.jl
```diff
@@ -243,7 +243,9 @@ over species.
 # Returns
 Leftover space ∈ [0, 1]
 """
-function relative_leftover_space(loc_coral_cover::Matrix{Float64})::Matrix{Float64}
+function relative_leftover_space(
+    loc_coral_cover::Union{Matrix{Float64},Vector{Float32}}
```
Zapiano, line 247: One other comment: the way it is now, one can pass a `Vector{Float32}` but this function returning type is `Vector{Float64}`. Won't this cause an error?

Rosejoycrocker: I'm just testing and ran into an error with `test_rs_w_fig` which seems to be about the clustering. So I'm just running updates and then I'll test but you're right. Should I implement as Pedro suggests though?

src/decision/location_selection.jl
Rosejoycrocker: Using `nda[1,:].data.data` isn't working for me. Would it be : `sum_cover[scen_idx, :].data.data` ?

Zapiano: Hum.. .yes... why does it work for me then..? Can you try `using NamedDims` and `using AxisKeys` just to confirm? 

Also, when you do `sum_cover[scen_idx, :].data` what happens?

Rosejoycrocker: I'll try this

Rosejoycrocker: Just `sum_cover[scen_idx, :].data` works :) I'll push the changes

Zapiano: All that said, I still think this should be assigned to a separate variable for readability (:

Rosejoycrocker: Thanks @Zapiano, realised I should have pre-allocated `leftover_space` outside the for loop anyway, so I've added that now. All tests passing, let me know if there's anything else but I think it's ready

PR #614, opened 2023-12-02
Support RSA with large sample sizes

ConnectedSystems:
Available GC threads get exhausted for large sample sizes (>= 8192) when using mutable static vectors as cache.
I believe because under the hood the vectors are represented as tuples, and modifying contents triggers GC to run.

Switching to plain vectors resolves the issue and does not drastically hurt performance.

### DIFF
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -392,7 +392,8 @@ function tsa(rs::ResultSet, y::AbstractMatrix{<:Real})::NamedDimsArray
 end
 
 """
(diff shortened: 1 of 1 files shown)


Rosejoycrocker (approved):
This looks good to me, I've run the function and it definitely seems faster + all tests passing. Do you want me to try running a large sample or have you done this?

Rosejoycrocker:
> > Do you want me to try running a large sample or have you done this?
> 
> I certainly have, yes ma'am.

Ok cool haha :) I'll merge

PR #623, opened 2023-12-04
Changes to dhw and wave logs to store scenario and site means and stds

### LINKED ISSUE #619: Change log for dhws and waves to store mean and std for scenarios and sites/reefs
Currently dhw and wave logs store the mean and std of each scenario used in the result set. Visualisation functions for the MCDA criteria (#620) , however, will need the mean and std over scenarios and sites available in the result set for criteria plots.

Rosejoycrocker:
Closes #619

### DIFF
--- a/src/io/result_io.jl
+++ b/src/io/result_io.jl
@@ -42,8 +42,8 @@ Matrix{Float64, 2}, of mean and standard deviation for each environmental scenar
 function summarize_env_data(data::AbstractArray)::Array{Float64}
     # TODO: Update once
     stats_store::Array{Float64} = zeros(2, size(data, 3), size(data, 2))
-    stats_store[1, :, :] .= Array(dropdims(mean(data; dims=1); dims=1)')
-    stats_store[2, :, :] .= Array(dropdims(std(data; dims=1); dims=1)')
+    stats_store[1, :, :] .= dropdims(mean(data; dims=1); dims=1)'
+    stats_store[2, :, :] .= dropdims(std(data; dims=1); dims=1)'
     return stats_store
 end
 
--- a/src/io/result_io.jl
+++ b/src/io/result_io.jl
@@ -362,9 +362,12 @@ function _recreate_stats_from_store(zarr_store_path::String)::Dict{String,Abstra
         store = zopen(sd, fill_as_missing=false)
 
         dims = store.attrs["structure"]
-        row_names = string.(store.attrs["rows"])
-        col_names = string.(store.attrs["cols"])
-        stat_set = NamedDimsArray(store[:, :]; zip(Symbol.(dims), [row_names, col_names])...)
+        stats = string.(store.attrs["stats"])
+        scenario_ids = string.(store.attrs["scenarios"])
+        loc_ids = string.(store.attrs["locations"])
+        stat_set = NamedDimsArray(
+            store[:, :, :]; zip(Symbol.(dims), [stats, scenario_ids, loc_ids])...
+        )
 
         stat_d[rcp_dirs[i]] = stat_set
     end
--- a/src/io/result_io.jl
+++ b/src/io/result_io.jl
@@ -361,12 +361,13 @@ function _recreate_stats_from_store(zarr_store_path::String)::Dict{String,Abstra
     for (i, sd) in enumerate(rcp_stat_dirs)
         store = zopen(sd, fill_as_missing=false)
 
-        dims = store.attrs["structure"]
+        dim_names = Symbol.(store.attrs["structure"])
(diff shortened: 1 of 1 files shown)


src/io/result_io.jl
```diff
@@ -41,10 +41,10 @@ Matrix{Float64, 2}, of mean and standard deviation for each environmental scenar
 """
 function summarize_env_data(data::AbstractArray)::Array{Float64}
     # TODO: Update once
-    dc_mean = dropdims(mean(data, dims=(1, 2)), dims=(1, 2))'
-    dc_std = dropdims(std(data, dims=(1, 2)), dims=(1, 2))'
-
-    return Array{Float64}(vcat(dc_mean, dc_std))
+    stats_store = Array{Float64}(zeros(2, size(data, 3), size(data, 2)))
+    stats_store[1, :, :] .= Array(dropdims(mean(data; dims=1); dims=1)')
+    stats_store[2, :, :] = Array(dropdims(std(data; dims=1); dims=1)')
```
Zapiano, line 46: Why on line 45 you are doing `.=` and on line 46 just `=`?

src/io/result_io.jl
```diff
@@ -68,17 +68,20 @@ N is the number of dhw/wave scenarios.
 function store_env_summary(data_cube::NamedDimsArray, type::String, file_loc::String, rcp::String, compressor::Zarr.Compressor)::ZArray
     stats = summarize_env_data(data_cube)
 
-    stats_store = zcreate(Float32, (2, size(stats, 2))...;
+    stats_store = zcreate(
+        Float32,
```
Zapiano, line 72: `summarize_env_data` creates an `Array{Float64}` and here you are creating a `ZArr{Float32}`. Just want to doublecheck check if maybe you want to change one of them..?  (I know there may be reasons for doing this so, if this is right, feel free to ignore this comment).

src/io/result_io.jl
```diff
@@ -359,9 +362,12 @@ function _recreate_stats_from_store(zarr_store_path::String)::Dict{String,Abstra
         store = zopen(sd, fill_as_missing=false)
 
         dims = store.attrs["structure"]
```
Zapiano, line 364: Alternatively, you could do `Symbol.(store.attrs["structure"])` and call this `dim_names` then when you call `NamedDimsArray` you just pass `dim_names`. (Another comment you can ignore if you prefer the way it is ;) )

Zapiano (approved):
Just a few small comments, but overall LGTM. (:

src/io/result_io.jl
```diff
@@ -41,10 +41,10 @@ Matrix{Float64, 2}, of mean and standard deviation for each environmental scenar
 """
 function summarize_env_data(data::AbstractArray)::Array{Float64}
     # TODO: Update once
-    dc_mean = dropdims(mean(data, dims=(1, 2)), dims=(1, 2))'
-    dc_std = dropdims(std(data, dims=(1, 2)), dims=(1, 2))'
-
-    return Array{Float64}(vcat(dc_mean, dc_std))
+    stats_store = Array{Float64}(zeros(2, size(data, 3), size(data, 2)))
```
Zapiano, line 44:
```suggestion
    stats_store::Array{Float64} = zeros(2, size(data, 3), size(data, 2))
```

src/io/result_io.jl
Rosejoycrocker: Thanks, must have missed pushing it

src/io/result_io.jl
Rosejoycrocker: Not sure, this was in Takuya's original function- @ConnectedSystems is there a reason for one being Float32 and one Float64?

ConnectedSystems: Data is stored as Float32 to save on space. This was an early decision that I ran by the team 2+ years ago. We lose precision by doing this of course, but not sure if it matters.

FYI - The RM team also store all their results as single precision values.

src/io/result_io.jl
ConnectedSystems: Looking at this, two questions come to mind:

1. Aren't these matrices? (`Matrix()`). If so, I prefer the more explicit `Matrix` over the general `Array`. ?
If it is an `Array`, I'd prefer the number of dimensions be specified.

2. Isn't `data` an array type already? Is the `Array()` necessary?

src/io/result_io.jl
Rosejoycrocker: Thanks Takuya

Rosejoycrocker:
I think this is ready for final review @Zapiano, tests all passing on my end

PR #625, opened 2023-12-06
Fix bug introduced by SIRUS v2.0.1

Zapiano:
As in title.


src/analysis/rule_extraction.jl
```diff
@@ -46,10 +46,11 @@ Vector of Rule condition clauses (each one being a vector itself).
...
-        direction = split.direction
-        value = split.splitpoint.value
+
+    for subclause in rules.rules[index].clause.subclauses
+        feature_name::String = subclause.feature_name
+        direction::Symbol = subclause.direction
+        value::Float32 = subclause.splitval
         push!(condition, [feature_name, direction, value])
     end
     return condition
```
ConnectedSystems, line 56: Minor comment - I prefer `return` to be separated by a line but otherwise looks fine to me

ConnectedSystems (approved):

PR #627, opened 2023-12-06
Fix `S` allocation bug in `rsa` and `outcome_map`

### LINKED ISSUE #615: Fix "surprising" readjustment of the number of slices used to perform RSA/outcome mapping
The number of conditioning points / slices / bins to use in RSA and outcome mapping ($S$) get automatically (and silently) adjusted to the maximum number of categorical values found in the model spec.

https://github.com/open-AIMS/ADRIA.jl/blob/83028faca9dbc67f1b447b198afb1867c3b38146/src/analysis/sensitivity.jl#L465-L468

This has implications both to expected runtime, and validity of the analysis (an increase to the number of conditioning points generally increases the required number of samples).

Instead, the number of slices should only be increased for categorical/integer factors, not for everything.

The result matrix should instead of adjusted with column labels indicating the bin/quantile/conditioning point/position. 
Only the columns that are relevant for a given factor should be filled.

This has implications for how the visualisations handle the result matrix so changes to address this needs careful consideration of visualisation processes.

Rosejoycrocker:
Closes #615

Previously, `S`, the number of bins for sensitivity analysis, was set as the max between the input `S` and number of bins needed for a categorical variable. These changes set the size of storage as the maximum between the input `S` and the size of the categorical variables, and then `S` is set as the input `S` for non-categorical variables during analysis.

Setting this as a draft for discussion as the labels need to be changed to work with visualisations still.


Rosejoycrocker:
My thought is that the labels should be set to the input `S` bin steps, because the categorical variable (guided) labels are set with symbols anyway during plotting. If S is smaller than the storage size though, I'm not sure how the unfilled bins should be labelled... Thoughts, @ConnectedSystems ?

Rosejoycrocker:
Closing this as has been superceded by #679

PR #629, opened 2023-12-11
Fix mcda method bugs 

Rosejoycrocker:
This PR fixes a couple of bugs found in mcda methods during testing.

VIKOR method has been returning NaNs in the case where some weights are zero. This is due to the formulation of the Q score, which requires division by (column max - column min). This has been fixed by adding a line in the VIKOR method which removes any zero columns.

ROV method was found to be incorrectly reversing site ranks. An example of this is :
```julia
results.scores =
 0.6371641507442964
 0.6941436816327351
 0.877935879420678
 1.0066403672998092
 1.033670846678704
 1.5902842580982375
 1.7501562536882735
 0.5729368306847074
 0.4478149807103405
 0.4484346568894094

results.bestIndex =2
```

In this case, site 2 is neither the highest nor lowest score. This suggests either a bug in JMcDM, or that the method searches for something other than a max/min score in ranking alternatives. Either way, for now it is best to remove ROV.

### DIFF
--- a/src/decision/mcda_methods.jl
+++ b/src/decision/mcda_methods.jl
@@ -106,7 +106,7 @@ Details of this aggregation method in, for example [1]
 """
 function adria_vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Float64}
 
-    # Remove zero columns as these will return NaNs in the Q score 
+    # Remove columns with constant values. Zero-valued columns will return NaNs in the Q score.
     S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
     F_s = maximum(S)
 
--- a/src/decision/mcda_methods.jl
+++ b/src/decision/mcda_methods.jl
@@ -106,8 +106,8 @@ Details of this aggregation method in, for example [1]
 """
 function adria_vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Float64}
 
-    # Remove zero columns as these will return NaNs in the Q score
-    S = S[:, vec(sum(S .> 0.0; dims=1) .!= 0)]
+    # Remove zero columns as these will return NaNs in the Q score 
+    S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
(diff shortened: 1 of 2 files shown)


ConnectedSystems:
> ROV method was found to be incorrectly reversing site ranks.

Without knowing how the ROV (Range of Value) method works I'm not sure we can determine this is the issue.
Going by the scores alone, it doesn't look like a reversal issue, but judging by the name of the method, it does seem it does not search for a given min/max. Not saying we need to keep it in, but we do need to justify eventually why we ignore/remove it.

Rosejoycrocker:
> > ROV method was found to be incorrectly reversing site ranks.
> 
> Without knowing how the ROV (Range of Value) method works I'm not sure we can determine this is the issue. Going by the scores alone, it doesn't look like a reversal issue, but judging by the name of the method, it does seem it does not search for a given min/max. Not saying we need to keep it in, but we do need to justify eventually why we ignore/remove it.

Agree , I can look into it more. It's not so much the fact that it is working differently to other methods, it's that `results.BestIndex!=maximum(scores)`, which causes the order to be reversed when ranking when it maybe shouldn't be. On  checking the non-reversed order of ranks, it lines up well with the ranks of other methods, suggesting the method is correct if not reversed. But because the 'results.bestIndex' is neither the max or the min, I'm not sure how to interpret this. Perhaps the method is meant to find a compromise/middle value or something.

src/decision/mcda_methods.jl
```diff
@@ -106,6 +106,8 @@ Details of this aggregation method in, for example [1]
 """
 function adria_vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Float64}
 
+    # Remove zero columns as these will return NaNs in the Q score
```
ConnectedSystems, line 109:
```suggestion
    # Remove columns with constant values. Zero-valued columns will return NaNs in the Q score.
```

ConnectedSystems (approved):
I still don't like the check put in place but okay.

PR #631, opened 2023-12-13
Fix set_factor_bounds! out of bounds for discrete factors

### LINKED ISSUE #630: set_factor_bounds! setting value outside range for discrete variables
When we call `set_factor_bounds!` with a discrete param, it does the following:

```
new_bnds = (round(new_lower), round(new_upper) + 1.0)
```

With this, if `new_upper == default_upper` (for example, when we only want to change the lower bound) it sets the upper bound to `default_upper + 1`.

Zapiano:
As in title.

Closes #630

### DIFF
--- a/src/io/sampling.jl
+++ b/src/io/sampling.jl
@@ -615,21 +615,6 @@ function _check_bounds_range(
     return nothing
 end
 
-"""
-Check specified bounds for validity.
-
-Raises error if lower bound values are greater than upper bounds.
-
-# Arguments
-- `lower` : lower bounds
-- `upper` : upper bound values
-"""
-function _check_bounds(lower, upper)
-    if any(lower .> upper)
-        error("Bounds are not legal (upper bound must be greater than lower bound)")
-    end
-end
-
 _offdiag_iter(A) = collect(ι for ι in CartesianIndices(A) if ι[1] ≠ ι[2])
 
 """
--- a/src/io/sampling.jl
+++ b/src/io/sampling.jl
@@ -24,9 +24,18 @@ function _is_discrete_factor(p_type::String)::Bool
 end
 function _is_discrete_factor(dom::Domain, fieldname::Symbol)::Bool
     model::Model = dom.model
-    param_filter::BitVector = collect(model[:fieldname]) .== fieldname
-    ptype::String = model[:ptype][param_filter][1]
+    param_idx::Int64 = findfirst(x -> x == fieldname, model[:fieldname])
+    ptype::String = model[:ptype][param_idx]
     return _is_discrete_factor(ptype)
+end
+
+"""
+    _distribution_type(dom::Domain, fieldname::Symbol)::String
+"""
+function _distribution_type(dom::Domain, fieldname::Symbol)::String
+    model::Model = dom.model
+    param_idx::Int64 = findfirst(x -> x == fieldname, model[:fieldname])
+    return model[:dists][param_idx]
 end
 
 """Set a model parameter value directly."""
@@ -466,12 +475,9 @@ end
 """
     get_bounds(dom::Domain, factor::Symbol)::Tuple
 
-Get factor bounds. If the factor is continuous, it will have a triangular distribution,
-then bounds will be a tuple with three elements (minimum, maximum, peak). If the factor is
-discrete, it will have a uniform distribution, then bounds will be a tuple with two elements
-(lower_bound, upper_bound). Note that, for discrete factors, the actual upper bound
-corresponds to the upper bound saved at the Domain's model_spec minus 1.0 because of how
-sampling works.
+Get factor lower and upper bounds. If the factor has a triangular distribution, it returns
+a 2-elements tuple (without the peack value). Note that, for discrete factors, the actual
+upper bound corresponds to the upper bound saved at the Domain's model_spec minus 1.0.
 
 # Arguments
 - `dom` : Domain
@@ -480,14 +486,9 @@ sampling works.
 function get_bounds(dom::Domain, factor::Symbol)::Tuple
     model::Model = dom.model
     factor_filter::BitVector = collect(model[:fieldname]) .== factor
+    bounds::Tuple{Float64,Float64} = model[:bounds][factor_filter][1]
 
-    bounds = if _is_discrete_factor(dom, factor)
-        lower, upper = model[:bounds][factor_filter][1]
-        (lower, upper - 1.0)
-    else
-        model[:bounds][factor_filter][1]
-    end
-
+    _is_discrete_factor(dom, factor) && return (bounds[1], bounds[2] - 1.0)
     return bounds
 end
 
@@ -503,14 +504,9 @@ Get factor default_bounds. Refer to `get_bounds` for more details of how the bou
 function get_default_bounds(dom::Domain, factor::Symbol)::Tuple
     model::Model = dom.model
     factor_filter::BitVector = collect(model[:fieldname]) .== factor
+    default_bounds::Tuple = model[:default_bounds][factor_filter][1]
 
-    default_bounds = if _is_discrete_factor(dom, factor)
-        default_lower, default_upper = model[:default_bounds][factor_filter][1]
-        (default_lower, default_upper - 1.0)
-    else
-        model[:default_bounds][factor_filter][1]
-    end
-
+    _is_discrete_factor(dom, factor) && return (default_bounds[1], default_bounds[2] - 1.0)
     return default_bounds
 end
 
@@ -518,11 +514,11 @@ end
     set_factor_bounds!(dom::Domain, factor::Symbol, new_bounds::Tuple)::Nothing
     set_factor_bounds!(dom::Domain; factors...)::Nothing
 
-Set new lower and upper bounds for a parameter, `new_bounds = (lower_bound, upper_bound)`.
-All sampled values `s_vals` for this parameter will lie within the range
-`lower_bound ≤ s_vals ≤ upper_bound`. When used to set new bounds for a discrete parameter,
-the upper bound will be set to `upper_bound + 1` to guarantee that the sample values will
-always be less or equal `upper_bound`.
+Set new lower and upper bounds for a parameter. All sampled values `s_vals` for this
+parameter will lie within the range `lower_bound ≤ s_vals ≤ upper_bound`. When used to set
+new bounds for a factor with discrete distribution, upper bound will be set to
+`upper_bound + 1` to guarantee that the sampled values will always be less than or equal
+`upper_bound`. This function will automatically set a new value to `val`.
 
 Note: Changes are permanent. To reset, either specify the original value(s) or reload the
 Domain.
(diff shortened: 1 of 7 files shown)


src/io/sampling.jl
Rosejoycrocker, line 1: Hey @Zapiano, I'm a bit confused- doesn't this mean the maximum value sampled will always be one less than the upper bound?

For example,

```julia
ADRIA.set_factor_bounds!(dom, :guided, (1, 5))
scens = ADRIA.sample_guided(dom, 2^12)
maximum(scens.guided) = 4
ADRIA.set_factor_bounds!(dom, :guided, (1, 10))
scens = ADRIA.sample_guided(dom, 2^12)
maximum(scens.guided) = 9

```

Zapiano: @Rosejoycrocker

I think I might be misunderstanding something. Because when I do 

```
ADRIA.set_factor_bounds!(dom, :guided, (1, 5))
ms = ADRIA.model_spec(dom)
ms[ms.fieldname .== :guided, :].bounds
```

I see that the bounds are (1.0, 5.0). But then after generating the samples, I agree that the maximum value is `4`. Because, in any case, the reason for this change is that, if I try to set the bounds to the default value, with the previous code I would get this:

```
bnds = ms[ms.fieldname .== :guided, :].default_bounds[1] # (-1.0, 19.0)
ADRIA.set_factor_bounds!(dom, :guided, bnds)
ms = ADRIA.model_spec(dom)
ms[ms.fieldname .== :guided, :].bounds # == (-1.0, 20.0)
```

And if I do the same with this PR's code, bounds is set to the right value `(-1.0, 19.0)`.

Zapiano: @Rosejoycrocker Just one other thing: the maximum value after sampling seems to always be one less than the maximum value of `bounds` - even before calling `set_factor_bounds!`. So I don't think what you pointed is a problem (again, unless I'm missing something here).

Rosejoycrocker: Sorry, had a chance to read through this properly now :) The upper bound needs to be set at `upper_bound+1` so that the sampling will include `upper_bound`, because the sampling uses floats which are then rounded to the nearest integer. In Main if you set the bounds and sample, 

```julia
ADRIA.set_factor_bounds!(dom, :guided, (1, 5))
scens = ADRIA.sample_guided(dom, 2^12)
 maximum(scens.guided) = 5
```

You are correct though that the new upper could be set to default_upper +1 and increase above the default upper. Could we instead add to the catch here to disallow this : https://github.com/open-AIMS/ADRIA.jl/blob/c07dc5fd16708dd1e8c59a4e4d6ad56745e51c73/src/io/sampling.jl#L395

Zapiano: @Rosejoycrocker I had a chat with @ConnectedSystems yesterday about this to try to better understand how does the sampling work and how is the user supposed to interact with it. Based on that, I've changed this PR a little bit. The idea here is:

If I understand, when we talk about continuous factors, the lower and upper bounds imply that each sample value (let's call them `s_val`) will be within the range `lower_bound <= s_val <= upper_bound`. In other words, in this case the `s_val` can be equal to `uppwer_bound`. But with discrete factor the `upper_bound` is set to the maximum desired value of `s_val` plus one, so we will have `lower_bound <= s_val < upper_bound`. 

If all that is correct, that means that when the user set a new value for the factor bounds, these factor will have different meanings depending on the type of variable (continuous / discrete). To prevent this confusion, what I did was: I am assuming that when the user calls `set_factor_bounds!` it is interpreting the meaning of bounds as in `lower_bound <= s_val <= upper_bound`, it doesn't matter if the variable is continuous or discrete. 

To prevent the user from getting confused when looking for the bounds directly in the model_spec, I also wrote functions `get_bounds` and `get_default_bounds` that take care of this conversion.

Let me know if there's anything wrong/weird, if you have questions/comments, etc.

Rosejoycrocker: Thanks @Zapiano, I'll have a look today :)

src/io/sampling.jl
```diff
@@ -370,52 +370,116 @@ function fix_factor!(d::Domain; factors...)::Nothing
 end
 
 """
-    set_factor_bounds!(d::Domain, factor::Symbol, new_bnds::Tuple)::Nothing
-    set_factor_bounds!(d::Domain; factors...)::Nothing
+    get_bounds(dom::Domain, factor::Symbol)::Tuple
 
-Fix bounds of a parameter for sampling to those provided.
+Get factor bounds. If the factor is continuous, it will have a triangular distribution,
```
Rosejoycrocker, line 375: Note, some of the continuous factors have a uniform distribution (e.g. https://github.com/open-AIMS/ADRIA.jl/blob/c07dc5fd16708dd1e8c59a4e4d6ad56745e51c73/src/decision/CriteriaWeights.jl#L12 )

Zapiano: Ok, thanks! Question: what determines the type of distribution to be used with each factor?

Rosejoycrocker: The only descriptor that really captures this is `dists="unif", dists="triag"` etc. It's mostly the coral parameters (defined in `Corals.jl`) and intervention parameters (defined in `Interventions.jl`), that have triangular distributions, but there are a lot of exceptions in there, like `:guided'.

Rosejoycrocker (approved):
Looks good to me now :) All tests passing on my end

src/io/sampling.jl
```diff
@@ -370,52 +370,116 @@ function fix_factor!(d::Domain; factors...)::Nothing
...
+function get_bounds(dom::Domain, factor::Symbol)::Tuple
+    model::Model = dom.model
+    factor_filter::BitVector = collect(model[:fieldname]) .== factor
+
+    bounds = if _check_discrete(dom, factor)
+        lower, upper = model[:bounds][factor_filter][1]
+        (lower, upper - 1.0)
+    else
+        model[:bounds][factor_filter][1]
+    end
```
ConnectedSystems, line 395: I'm personally not a fan of these implicit return assignments.

Other than that, I suggest we stick to using [early returns](https://openfolder.sh/returning-early) to improve readability.

```julia
lower, upper = model[:bounds][factor_filter][1]
if _check_discrete(dom, factor)
    return (lower, upper - 1.0)
end

return (lower, upper)
```

Zapiano: The first version of this code I wrote was precisely what you are suggesting. The problem is that sometimes bounds is a 3-values Tuple. But ok, let me try something else here then.

ConnectedSystems: Oh, use `getindex()`

```julia
bounds = model[...]
lower, upper = getindex.(bounds, (1, 2))
```

Might have to use `map` instead of `getindex()` but something like the above is what you want

Zapiano: Huum, cool thanks for the tip!

Zapiano:
@ConnectedSystems  and @Rosejoycrocker  tests are not passing (yet) but just I've commited so you can check what I'm doing.

src/io/sampling.jl
```diff
@@ -585,36 +562,62 @@ function set_factor_bounds!(d::Domain; factors...)::Nothing
...
 
     if _is_discrete_factor(dom, factor)
         # If factor is discrete, sampled values must be within default_lower and (default_upper - 1)
-        out_of_bounds = (new_lower < default_lower) || (new_upper > (default_upper - 1))
+        out_of_bounds = (new_lower < default_lower) || (new_upper > default_upper)
     else
         out_of_bounds = (new_lower < default_lower) || (new_upper > default_upper)
     end
 
+    # Check if new bounds are withing default range
```
ConnectedSystems, line 603: "withing" -> "within"

add "the"

`# Check if new bounds are within the default range`

src/io/sampling.jl
```diff
@@ -585,36 +562,62 @@ function set_factor_bounds!(d::Domain; factors...)::Nothing
...
     if out_of_bounds
         error(
             "Bounds should be within ($default_lower, $default_upper), received: ($new_lower, $new_upper).",
         )
     end
 
+    # Check if number of bounds
```
ConnectedSystems, line 610: Check if number of bounds are what?

src/io/sampling.jl
```diff
@@ -466,12 +475,9 @@ end
...
     get_bounds(dom::Domain, factor::Symbol)::Tuple
 
-Get factor bounds. If the factor is continuous, it will have a triangular distribution,
-then bounds will be a tuple with three elements (minimum, maximum, peak). If the factor is
-discrete, it will have a uniform distribution, then bounds will be a tuple with two elements
-(lower_bound, upper_bound). Note that, for discrete factors, the actual upper bound
-corresponds to the upper bound saved at the Domain's model_spec minus 1.0 because of how
-sampling works.
+Get factor lower and upper bounds. If the factor has a triangular distribution, it returns
+a 2-elements tuple (without the peack value). Note that, for discrete factors, the actual
```
ConnectedSystems, line 479: "peack" -> "peak"

`(without the peak value)`

src/io/sampling.jl
```diff
@@ -503,73 +504,49 @@ Get factor default_bounds. Refer to `get_bounds` for more details of how the bou
...
 
 # Arguments
 - `dom` : Domain
 - `factor` : Parameter whose bounds will be change to a new value
-- `new_bounds` : Tuple of lower and upper bounds to be set as the new bounds of the
-respective parameter
+- `new_bounds` : Tuple bounds to be set as the new bounds of the respective factor. When
+factor has a triangular distribution, `new_bounds` must be is a 3-dimensional Tuple, with
+`(new_min, new_max, new_peak)` values; when it has a uniform distribution, `new_bounds` must be a
+2-dimensional Tuple, with `(new_lower, new_upper)` values.
```
ConnectedSystems, line 532: The values are the elements of the tuple.

`dimensional` -> `element`

src/io/sampling.jl
```diff
@@ -503,73 +504,49 @@ Get factor default_bounds. Refer to `get_bounds` for more details of how the bou
...
     set_factor_bounds!(dom::Domain, factor::Symbol, new_bounds::Tuple)::Nothing
     set_factor_bounds!(dom::Domain; factors...)::Nothing
 
-Set new lower and upper bounds for a parameter, `new_bounds = (lower_bound, upper_bound)`.
-All sampled values `s_vals` for this parameter will lie within the range
-`lower_bound ≤ s_vals ≤ upper_bound`. When used to set new bounds for a discrete parameter,
-the upper bound will be set to `upper_bound + 1` to guarantee that the sample values will
-always be less or equal `upper_bound`.
+Set new lower and upper bounds for a parameter. All sampled values `s_vals` for this
+parameter will lie within the range `lower_bound ≤ s_vals ≤ upper_bound`. When used to set
```
ConnectedSystems, line 518: "Sampled values for a parameter will lie within the range `lower_bound ≤ s_vals ≤ upper_bound`, where `s_vals` is the value of the factor"

... or something like the above.

Zapiano:
@ConnectedSystems @Rosejoycrocker  all tests passing now

ConnectedSystems (approved):
Thanks @Zapiano , as discussed I'll be making further changes to incorporate advances in the Distributions package on top of this.

PR #632, opened 2023-12-15
Consolidate methods that transform factor values into `sampling.jl`

ConnectedSystems:
Move `_check_discrete()` and `map_to_discrete()` methods into `sampling.jl` from `Ecosystem.jl` to maintain conceptual consistency.

This PR depends on #631 which should be merged first if possible. Currently this PR targets `[fix-set-factor-bounds](https://github.com/open-AIMS/ADRIA.jl/tree/fix-set-factor-bounds)` but could be rebased if/as necessary.

### DIFF
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -7,63 +7,6 @@ using Distributions
 
 abstract type EcoModel end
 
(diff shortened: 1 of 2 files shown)


Zapiano (approved):
LGTM

Zapiano:
@ConnectedSystems thanks for that ;)

PR #635, opened 2023-12-20
Replaces distance sorting with restrictions on the number of selected locations in the same reef

### LINKED ISSUE #618: Adjust `distribute_seeded_corals` to account for cases where the top 5 sites do not have enough space for seeded corals
Currently, after sites/reefs are chosen to seed at, the area of corals to seed is split between the 5 selected sites/reefs in proportion to the area available there. However, if the total area available is less than the area to be seeded, this is not accounted for and sites/reefs are seeded with more area than they can accomodate. This added area is then lost when `proportional_adjustment!` is applied and the site coral is adjusted to be no greater than capacity, meaning added area due to seeding may be unaccounted for.

This issue can be addressed by expanding the number of sites to seed at as needed for the area of coral to seed. If not enough space is available at the 5 seeding sites, seeding can occur at the next most highly ranked sites until sufficient space is found to accomodate the seeded coral. 

This may require outputting a larger ordered list of sites in `prefseedsites` so that the next most highly ranked sites are easily selected in `distribute_seeded_corals`.

Rosejoycrocker:
Currently, the `distance_sorting` algorithm (implemented if `use_dist=true`) evaluates the set of locations selected in guided location selection and replaces any which are "too close" with the next most highly ranked locations which are sufficiently far apart. This is problematic, however, as a minimum distance must be defined. This was defined as a proportion of the median distance, but was found to remove too many locations from the original set and difficult to understand the implications of.

This PR replaces `distance_sorting` with `constrain_reef_cluster` which checks how many of the selected locations are in the same reef. The maximum allowed number of locations in a given reef is set by `n_reefs`, a sim constant, and if the selected locations contain too many belonging to the same reef, the lowest ranked of these will be replaced by the next highest ranked location in a different reef.

Additionally, the algorithm first checks how much space is available in the selected location. If not enough space is available to fit seeded corals in `n_loc_iv` locations, the number of locations to intervene on is expanded until there is sufficient space.

Closes #618


test/constrain_reef_clusters.jl
```diff
@@ -0,0 +1,77 @@
...
+    s_order = Union{Float64,Int64}[Int64.(orig_site_order) rand(n_sites)]
+    rankings = Int64[site_ids zeros(Int64, n_sites) zeros(Int64, n_sites)]
+
+    new_prefsites, rankings = constrain_reef_cluster(
+        reef_locs, s_order, rankings, area_to_seed, available_space, n_site_iv, 3
+    )
+    @test length(new_prefsites) == (n_site_iv + 1) ||
+        "Not enough sites were selected to fit the corals to be seeded."
+
+end
```
ConnectedSystems, line 77:
```suggestion
end

```

Newline at end of file. Please turn on "Format on save" or format the file before committing.

test/runtests.jl
```diff
@@ -343,6 +343,7 @@ include("metrics.jl")
 include("sampling.jl")
 include("seeding.jl")
 include("site_selection.jl")
+include("constrain_reef_clusters.jl")
```
ConnectedSystems, line 346: How about a more generic name like "spatial_clustering.jl"?

src/ecosystem/const_params.jl
```diff
@@ -32,6 +32,7 @@ Struct of simulation constants for ADRIA
 """
 Base.@kwdef mutable struct SimConstants
     n_site_int::Int64 = 5  # max number of sites we intervene on in a given year.
+    n_reefs::Int64 = 3 # max number of locations selected in the same site/cluster (when using spatial group constraints)
```
ConnectedSystems, line 35:
```suggestion
    n_reefs::Int64 = 3  # max number of locations selected in the same site/cluster (when using spatial group constraints)
```

Double space before inline comments

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+            # Locations to replace
+            replace_locs = [
+                pref_locs[pref_reefs .== gp][(n_reefs + 1):end] for gp in reefs_swap
+            ]
+            replace_locs = replace_locs[1:end...]
+            reef_add = unique_reefs[(sum_pref_locs .+ 1) .<= n_reefs] # acceptable reefs to swtich out for
+
+            pref_locs = setdiff(pref_locs, replace_locs) # Remove locations to be replaced from preferred sites
+            loc_order = setdiff(loc_order, replace_locs) # Remove locations to be replaced from site order
+            add_locs = setdiff(loc_order, pref_locs) # Locations which can be added in place
```
ConnectedSystems, line 790:
```suggestion
            reef_add = unique_reefs[(sum_pref_locs .+ 1) .<= n_reefs]  # acceptable reefs to swtich out for

            pref_locs = setdiff(pref_locs, replace_locs)  # Remove locations to be replaced from preferred sites
            loc_order = setdiff(loc_order, replace_locs)  # Remove locations to be replaced from site order
            add_locs = setdiff(loc_order, pref_locs)  # Locations which can be added in place
```

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+    n_site_int::Int64,
+    n_reefs::Int64,
+)
+    # Get full ordering of locations
+    loc_order = s_order[:, 1]
+    # Get unique reefs in location set
+    unique_reefs = unique(reefs)
+    pref_locs = loc_order[1:n_site_int]
+
+    num_locs = n_site_int # No. of locations to consider
```
ConnectedSystems, line 757:
```suggestion
    # Get full ordering of locations
    loc_order = s_order[:, 1]

    # Get unique reefs in location set
    unique_reefs = unique(reefs)
    pref_locs = loc_order[1:n_site_int]

    num_locs = n_site_int  # Number of locations to consider
```

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+- `available_space` : absolute area available at each location (m²)
+- `n_site_int` : Minimum number of sites to intervene at
+- `n_reefs` : Number of selected locations to allow in the same reef.
 
 # Returns
-New set of selected sites for seeding or shading.
+Tuple :
+    - `pref_locs` : Vector, Indices of preferred intervention locations
+    - `rankings` : Matrix[n_sites ⋅ 3] where columns are site_id, seeding_rank, shading_rank
+        Values of 0 indicate sites that were not considered
```
ConnectedSystems, line 740: No indentation is necessary, and leads to incorrect formatting when the docs are generated

```suggestion
Tuple :
- `pref_locs` : Vector, Indices of preferred intervention locations
- `rankings` : Matrix[n_sites ⋅ 3] where columns are site_id, seeding_rank, shading_rank
    Values of 0 indicate sites that were not considered
```

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
 end
 
 """
-    distance_sorting(pref_locs::AbstractArray{Int}, s_order::Matrix{Union{Float64,Int64}}, dist::Matrix{Float64}, min_dist::Float64, rankings::Matrix{Int64}, rank_col::Int64)::Tuple{Vector{Union{Float64,Int64}},Matrix{Int64}}
-
-Find selected locations with distances between each other < median distance-dist_thresh*(median distance).
-Replaces these locations with those in the top ranks if the distance between these is greater.
+    constrain_reef_cluster(reefs::Union{Vector{String},Vector{Float64}}, s_order::Matrix{Union{Float64,Int64}}, rankings::Matrix{Int64},
+    area_to_seed::Float64, available_space::Vector{Float64}, n_site_int::Int64, n_reefs::Int64)
```
ConnectedSystems, line 725: No return type specified?

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+)
+    # Get full ordering of locations
+    loc_order = s_order[:, 1]
+    # Get unique reefs in location set
+    unique_reefs = unique(reefs)
+    pref_locs = loc_order[1:n_site_int]
+
+    num_locs = n_site_int # No. of locations to consider
+
+    for ll in 1:length(loc_order)
```
ConnectedSystems, line 759: Please use standard notation: `ll` -> `i` or at least make it readable: `ll` -> `loc`

src/decision/dMCDA.jl
ConnectedSystems: Additionally, as we discussed, could we replace `n_site_int` with `n_iv_locs` or similar.

src/decision/dMCDA.jl
ConnectedSystems: Please go through and standardize inline comments

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+        # If enough space for seeding corals, keep n_site_int, else expand as needed
+        cumulative_space = cumsum(available_space[loc_order])
+        num_locs = if cumulative_space[n_site_int] .< area_to_seed
+            findfirst(>=(area_to_seed), cumulative_space)
+        else
+            n_site_int
+        end
```
ConnectedSystems, line 766: Can't we just set `num_locs = max(findfirst(>=(area_to_seed), cumulative_space), n_site_int)` in the first instance?

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+            add_locs_ind = findall(
+                dropdims(
+                    any(
+                        reshape(reefs[add_locs], 1, length(reefs[add_locs])) .== reef_add;
+                        dims=1,
+                    );
+                    dims=1,
+                ),
+            )
+            # New preferred location set
```
ConnectedSystems, line 802: Add space for readability

```suggestion

            # New preferred location set
```

test/constrain_reef_clusters.jl
```diff
@@ -0,0 +1,77 @@
...
+    n_site_iv = 5
+
+    prefsites = site_order[1:5]
+    reef_locs = [fill("1", 10)..., fill("2", 10)..., fill("3", 10)...]
+
+    s_order = Union{Float64,Int64}[Int64.(site_order) rand(n_sites)]
+    # All selected sites are in the same reef, so 2 should be replaced
+    reef_locs[prefsites] .= "2"
+    # Empty ranking just for testing
+    rankings = Int64[site_ids zeros(Int64, n_sites) zeros(Int64, n_sites)]
```
ConnectedSystems, line 23: Please avoid wall-of-text blocks.

Add spaces between lines if you add comments.

```suggestion
    s_order = Union{Float64,Int64}[Int64.(site_order) rand(n_sites)]

    # All selected sites are in the same reef, so 2 should be replaced
    reef_locs[prefsites] .= "2"

    # Empty ranking just for testing
    rankings = Int64[site_ids zeros(Int64, n_sites) zeros(Int64, n_sites)]
```

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+
+            pref_locs = setdiff(pref_locs, replace_locs) # Remove locations to be replaced from preferred sites
+            loc_order = setdiff(loc_order, replace_locs) # Remove locations to be replaced from site order
+            add_locs = setdiff(loc_order, pref_locs) # Locations which can be added in place
+
+            # Indices of locations available that satisfy the reef constraint
+            add_locs_ind = findall(
+                dropdims(
+                    any(
+                        reshape(reefs[add_locs], 1, length(reefs[add_locs])) .== reef_add;
```
ConnectedSystems, line 796: I'm not understanding what this reshape is for, could you explain?

test/runtests.jl
Rosejoycrocker: Sure, sounds good. I made variable names in the function based on `reef_` but I can change this to be more generic also if that's best?

src/decision/dMCDA.jl
```diff
@@ -737,87 +721,111 @@ function guided_site_selection(
...
+    loc_order = s_order[:, 1]
+    # Get unique reefs in location set
+    unique_reefs = unique(reefs)
+    pref_locs = loc_order[1:n_site_int]
+
+    num_locs = n_site_int # No. of locations to consider
+
+    for ll in 1:length(loc_order)
+        # If enough space for seeding corals, keep n_site_int, else expand as needed
+        cumulative_space = cumsum(available_space[loc_order])
```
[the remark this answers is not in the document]
Rosejoycrocker: good point, `loc_order` changes but the areas will not, I'll move this outside

src/decision/dMCDA.jl
Rosejoycrocker: good point

src/decision/dMCDA.jl
Rosejoycrocker: I believe I did this because `reefs[add_locs]` and `reef_add` need to be orthogonal but `reefs` isn't a matrix so I can't just transpose, but I'll check

test/runtests.jl
ConnectedSystems: Inside the function is fine, the point is that the file is more generic so we can add other tests relevant to spatial clustering later.

src/decision/dMCDA.jl
Rosejoycrocker: There's a lot of instances of `n_site_int` in `dMCDA.jl` - should I replace all of these in this PR too?

ConnectedSystems: As we discussed earlier - we would ideally be making these changes/updates as necessary - i.e., when we come across old styles that should be updated.

Change the variable for this function, and any other place where you see it _within the context of this PR_.

test/spatial_clustering.jl
```diff
@@ -0,0 +1,76 @@
...
+		reef_locs, 
+		s_order, 
+		rankings,
+		area_to_seed,
+		available_space,
+		n_site_iv,
+		3,)
+
+	num_reefs = [sum(reef_locs[new_prefsites] .== rr) for rr in unique(reef_locs)]
+	l_diff_sites = length(setdiff(prefsites, new_prefsites))
```
ConnectedSystems, line 35: Use readable names please

src/decision/dMCDA.jl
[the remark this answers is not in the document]
Rosejoycrocker: actually this does need to be inside the loop, because `cumsum` is order dependent

src/decision/dMCDA.jl
Rosejoycrocker: Ok, for now I've just changed within the function and relevant tests

src/decision/dMCDA.jl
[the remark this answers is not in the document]
ConnectedSystems: I don't follow here, as `loc_order` does not change, neither does `available_space`, and is not dependent on `ll`.
So I'm not sure what ordering is changed? Could you elaborate?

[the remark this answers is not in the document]
Rosejoycrocker: `loc_order` will change, it's updated when `pref_locs` is updated at each iteration (this will in reality probably only occur once or twice). `loc_order` captures the full order of sites, so only the first 5 site ids or so will change at each iteration, but this will effect the cumulative sum vector. You probably only need to evaluate the first `n_iv_loc*2` entries though to work out how many sites to choose... `ll` is a dummy iterator, it doesn't index anything. I could use `while` instead but I usually avoid using `while`. Let me know if you have a preference.

src/decision/dMCDA.jl
Rosejoycrocker: So on checking, I did this as an alternative to transposition to allow for the possibility that the reef ids are strings (and so the vector cannot be transposed because the adjoint doesn't exist), which they are in the GBR domain (and also in tests although this can be changed)

src/decision/dMCDA.jl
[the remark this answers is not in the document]
ConnectedSystems: Sorry, I had a look at the full implementation.

This is an issue - you cannot update a variable while looping over it. Only the initial length will be considered.

Even if you could it would introduce other errors.
The only reason it seems to work in your case, I suspect, is that you're only testing situations where the location list can be resolved in a single loop.

```julia
x = rand(10)
for i in 1:length(x)
    x = rand(15)
    @info i
end
```

But specific to what you're trying to do here, I'm not sure it's even necessary.

The suggested change below:

`num_locs = max(findfirst(>=(area_to_seed), cumulative_space), n_site_int)`

Identifies `num_locs` that satisfy requirements already, no?
Meaning you only need to be looping to ensure the reefs are swapped out as necessary.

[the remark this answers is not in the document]
Rosejoycrocker: The suggested change works though only if `cumulative_space` is in the same order as the ranking order. If  `cumulative_space` is in the site id order, it doesn't tell you whether the sites you have selected have enough space, only that the first `num_locs` sites have enough space, regardless of their rank. What do you mean by `you cannot update a variable while looping over it'?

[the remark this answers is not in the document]
ConnectedSystems: > The suggested change works though only if cumulative_space is in the same order as the ranking order.

Then, why isn't `cumulative_space` in ranked order? Doesn't it make sense to take the space associated with each desirable location?

> What do you mean by "you cannot update a variable while looping over it"?

You expect the loop as defined by `for ll in 1:length(loc_order)` to be updated because you update `loc_order`.
As shown in my example above, updating `loc_order` will not affect the number of iterations.

src/decision/dMCDA.jl
ConnectedSystems: I still don't follow - looking at the type declarations, all the variables in use here (`unique_reefs`, `reefs`, `reef_add`) are all vectors so not sure why you need to transpose to a vector to a 1xN matrix to compare with a vector of size N.

Clearly I'm missing something, I'll have another closer look tomorrow.

src/decision/dMCDA.jl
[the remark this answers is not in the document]
Rosejoycrocker: `cumulative_space` is in ranked order when it is indexed by `loc_order`.`loc_order` will not affect the number of iterations but the loop will be broken out of is `reefs_swap` is empty : https://github.com/open-AIMS/ADRIA.jl/blob/39f20837d33c0067e8eea9af53a9d8b3b2a8ae9a/src/decision/dMCDA.jl#L776
`reefs_swap` will be empty only if the updated selected locations are within the reef constraint.

src/decision/dMCDA.jl
Rosejoycrocker: Let me know if you want to meet about it on Teams :)

src/decision/dMCDA.jl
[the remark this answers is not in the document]
ConnectedSystems: Let's chat about it

Rosejoycrocker:
I think this is ready for review again @ConnectedSystems. All tests passing on my end

ConnectedSystems:
Rose, could you please double check that your formatter is correctly configured to use Blue style, and that you are using 4-space indentation.

The recent change had formatted the file to use tabs instead of spaces for indentation.

Rosejoycrocker:
> Rose, could you please double check that your formatter is correctly configured to use Blue style, and that you are using 4-space indentation.
> 
> The recent change had formatted the file to use tabs instead of spaces for indentation.

These are the settings I have for the Julia formatter :
```style=BlueStyle(), indent=4, margin=256, always_for_in=true, for_in_replacement="∈", whitespace_typedefs=true, import_to_using=true, whitespace_in_kwargs=true, align_struct_field=true, align_assignment=true, align_conditional=true, align_pair_arrow=true, normalize_line_endings="unix", align_matrix=true, join_lines_based_on_source=true, indent_submodule=true, surround_whereop_typeparameters=false, yas_style_nesting=true```

I'm just looking to see if the indent measurement is in tabs instead of space but can't find anything yet.

ConnectedSystems (approved):
Looks good and tests pass for me as well.

ConnectedSystems:
> > Rose, could you please double check that your formatter is correctly configured to use Blue style, and that you are using 4-space indentation.
> > The recent change had formatted the file to use tabs instead of spaces for indentation.
> 
> These are the settings I have for the Julia formatter : `style=BlueStyle(), indent=4, margin=256, always_for_in=true, for_in_replacement="∈", whitespace_typedefs=true, import_to_using=true, whitespace_in_kwargs=true, align_struct_field=true, align_assignment=true, align_conditional=true, align_pair_arrow=true, normalize_line_endings="unix", align_matrix=true, join_lines_based_on_source=true, indent_submodule=true, surround_whereop_typeparameters=false, yas_style_nesting=true`
> 
> Seems like it should be set on spaces instead of tabs
> 
> Screenshot 2024-01-03 150807


https://code.visualstudio.com/docs/getstarted/settings#_language-specific-editor-settings

PR #636, opened 2023-12-22
Update factor types and their use

ConnectedSystems:
The `Distributions.jl` package now has support for discrete distributions.

There has also been: 

- a change in how distribution shape parameters are handled/interpreted.
- repeated use of "default" values when creating `Param()` types
- issues in how bounds should be handled/specified for discrete factors (e.g., #630)
- new evidence in the sampling literature that advices use of randomized Quasi-Monte Carlo approaches (#586)

This PR tracks a series of other planned changes to unify methods to be conceptually consistent, simplify use of ADRIA, and address associated bugs/annoyances.

This base PR defines a new constructor for Param types with ADRIA-specific metadata.

Principal changes include:

- `dists` -> `dist`, and now expects a Distributions.jl type, rather than a string indicating the distribution type
- `bounds` -> `dist_params`
- `ptype` indicating the parameter distribution type now expects a more informative name (see discussion in #512)
   one of: `["continuous", "ordered categorical", "unordered categorical", "discrete"]`

---

This PR defines custom convenience constructors for ADRIA-specific factors (this base PR - **MERGE LAST**).

PRs related to this set of changes (note, these should be accepted/merged in the indicated order):

- [x] #637
- [x] #638
- [x] #639
- [x] #653
- [x] #654

In the process of creating these PRs, the following issues are resolved:

- #512 
- #586

PR #637, opened 2023-12-22
Update factor definitions to use custom constructor

ConnectedSystems:
As in title - create parameter types using the custom `Factor` constructor.

Of the series of PRs as indicated in #636 , merge this one first.

### DIFF
--- a/src/decision/CriteriaWeights.jl
+++ b/src/decision/CriteriaWeights.jl
@@ -1,180 +1,172 @@
-Base.@kwdef struct CriteriaWeights{P,P2,P3} <: EcoModel
-    seed_wave_stress::P = Param(
+Base.@kwdef struct CriteriaWeights{U1,U2,U3,DU,DO} <: EcoModel
+    seed_wave_stress::U2 = Factor(
         1.0;
-        ptype="real",
-        bounds=(0.0, 1.0),
-        default_bounds=(0.0, 1.0),
-        dists="unif",
+        ptype="continuous",
+        dist=Uniform,
+        dist_params=(0.0, 1.0),
         criteria_keywords=(:seed, :weight),
         name="Seed Wave Stress",
         description="Importance of avoiding wave stress when seeding. Higher values places more weight on areas with low wave stress.",
     )
-    seed_heat_stress::P = Param(
+    seed_heat_stress::U2 = Factor(
         1.0;
-        ptype="real",
-        bounds=(0.0, 1.0),
-        default_bounds=(0.0, 1.0),
-        dists="unif",
+        ptype="continuous",
+        dist=Uniform,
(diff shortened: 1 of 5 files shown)


src/interventions/Interventions.jl
```diff
@@ -1,176 +1,140 @@
...
     )
-    N_seed_TA::N = Param(
+    N_seed_TA::DU = Factor(
         0;
-        ptype="integer",
-        bounds=(0.0, 1000000.0 + 1.0),
-        default_bounds=(0.0, 1000000.0 + 1.0),
-        dists="unif",
-        criteria_keywords=(""),
+        ptype="unordered categorical",
```
Rosejoycrocker, line 14: I would have thought variables like `N_seed_TA` etc. are ordered categorical because they translate to a scale of magnitude (e.g. the higher the category the larger the number of corals), whereas `guided` I would have thought of as unordered because the categories can go in any order and the meaning is not lost.

src/decision/CriteriaWeights.jl
```diff
@@ -1,210 +1,199 @@
-Base.@kwdef struct CriteriaWeights{P,P2,P3,N} <: EcoModel
-    seed_wave_stress::P = Param(
+Base.@kwdef struct CriteriaWeights{U1,U2,U3,DU,DO} <: EcoModel
+    seed_wave_stress::U2 = Factor(
         1.0;
-        ptype="real",
-        bounds=(0.0, 1.0),
-        default_bounds=(0.0, 1.0),
```
Rosejoycrocker, line 6: Will changes to sampling functions using `default_bounds` be done in a future PR?

ConnectedSystems: Yes. Please read the description above and for the linked PR.

src/interventions/Interventions.jl
ConnectedSystems: That's a good point - silly oversight on my part. I will correct.

Rosejoycrocker (approved):
Looks good to me now

src/decision/CriteriaWeights.jl
```diff
@@ -1,180 +1,172 @@
...
-        dists="unif",
+        ptype="continuous",
+        dist=Uniform,
+        dist_params=(0.0, 1.0),
         criteria_keywords=(:threshold, :seed, :fog),
         name="Risk Tolerance",
         description="Filters out sites with heat/wave stress above threshold.",
     )
-    depth_min::P3 = Param(
+    depth_min::DO = Param(
```
Zapiano, line 137: @ConnectedSystems Why haven't you changed this one to `Factor`?

ConnectedSystems: Ooops

PR #640, opened 2023-12-22
Address MCDA error for single locations

ConnectedSystems:
A bug was introduced in the previous fix #629, as it does not handle cases where only a single location is viable, resulting in an empty decision matrix.

### DIFF
--- a/src/decision/mcda_methods.jl
+++ b/src/decision/mcda_methods.jl
@@ -98,27 +98,28 @@ Details of this aggregation method in, for example [1]
     https://doi.org/10.3390/jrfm14060271
 
 # Arguments
-- `S` : Matrix
+- `S` : Decision matrix
 - `v` : Real
 
 # Returns
 - `Q` : aggregate score for ranking.
 """
-function adria_vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Float64}
-
-    # Remove columns with constant values. Zero-valued columns will return NaNs in the Q score.
-    S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
-    F_s = maximum(S)
+function adria_vikor(S::Matrix{Float64}; v::Float64=0.5)::Array{Float64}
+    # Remove columns with constant values.
+    # Zero-valued columns will return NaNs in the Q score.
+    if size(S, 1) > 1
+        S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
+    end
 
     # Compute utility of the majority Sr (Manhatten Distance)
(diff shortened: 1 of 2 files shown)


ConnectedSystems:
In the future, we should expand tests to cover the behaviour in these edge cases (n_locs >= 1).

src/decision/mcda_methods.jl
```diff
@@ -98,27 +98,28 @@ Details of this aggregation method in, for example [1]
...
 """
-function adria_vikor(S::Array{Float64,2}; v::Float64=0.5)::Array{Float64}
-
-    # Remove columns with constant values. Zero-valued columns will return NaNs in the Q score.
-    S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
-    F_s = maximum(S)
+function adria_vikor(S::Matrix{Float64}; v::Float64=0.5)::Array{Float64}
+    # Remove columns with constant values.
+    # Zero-valued columns will return NaNs in the Q score.
+    if size(S, 1) > 1
```
Rosejoycrocker, line 110: So there are cases where an empty matrix is not caught by the outer For loop? - https://github.com/open-AIMS/ADRIA.jl/blob/0667a6b6235bc2eeb7e7cc2d85bd96296c1b2b9e/src/decision/dMCDA.jl#L695

ConnectedSystems: This catches cases where the matrix has 1 row, not when it is empty.

If it has 1 row, then this:
```julia
S = S[:, .!vec((all(S .== maximum(S; dims=1); dims=1)))]
```

will remove it (because the maximum value is the only value), which then leads to an error.

Rosejoycrocker: Ah I see, thanks!

src/decision/mcda_methods.jl
```diff
@@ -98,27 +98,28 @@ Details of this aggregation method in, for example [1]
...
     # Compute individual regret R (Chebyshev distance)
-    sr_arg = (F_s .- S)
+    sr_arg = maximum(S) .- S
     Sr = sum(sr_arg, dims=2)
     R = maximum(sr_arg, dims=2)
 
     # Compute the VIKOR compromise Q
-    S_s, S_h = maximum(Sr), minimum(Sr)
-    R_s, R_h = maximum(R), minimum(R)
+    S_s, S_h = extrema(Sr)
```
ConnectedSystems, line 121: Incorrect order - should be `S_h`, `S_s` (minimum and maximum)

```suggestion
    S_h, S_s = extrema(Sr)
```

Rosejoycrocker (approved):
Tests all passing on my end

PR #647, opened 2024-01-11
Update use of Makie/GeoMakie to address deprecation errors

ConnectedSystems:
Addresses errors due to use of now fully deprecated keywords for Makie (v0.20.4) / GeoMakie (v0.6.0).

GeoMakie has updated (but the docs have not). The previous approach to setting lon/lat limits and associated tick labels does not work anymore.

The figure looks cleaner perhaps, but it would still be nice to see where we're looking at. In any case, I've opted to leave these off by default for later investigation.

image


ConnectedSystems:
These are minor changes so skipping the usual review. Sorry for the noise @Zapiano

PR #649, opened 2024-01-15
Add back commented lines for defining spatial plot limits

Rosejoycrocker:
Lines in `spatial.jl` which defined the x and y limits for spatial plotting had been left commented out, so that spatial plots are excessively zoomed out:

plot_cover_without_bounds

Adding the lines back in means plots can be seen without additional manual zooming in:
plot_cover_bounds

### DIFF
--- a/ext/AvizExt/viz/spatial.jl
+++ b/ext/AvizExt/viz/spatial.jl
@@ -44,14 +44,14 @@ function create_map!(
 
     spatial = GeoAxis(
         f[1, 1];
-        dest="+proj=latlong +datum=WGS84",
+        dest = "+proj=latlong +datum=WGS84",
         axis_opts...,
     )
-    # lon = first.(centroids)
-    # lat = last.(centroids)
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
Hi Rose, the reason I commented those lines out as a temporary fix is that those arguments no longer exist in the latest version of Makie. They have been deprecated for a while too.

In the latest version, the excessive zoom out does not happen, but we lose the axis tick labels.

The "real" fix would be to find out the new way of defining axis limits and use that instead

ConnectedSystems:
See https://github.com/open-AIMS/ADRIA.jl/pull/647

Rosejoycrocker:
> See #647

Ah ok, thanks, I'll put this in draft for now until I can find a new way of defining axis limits

ConnectedSystems:
Closing as this will be superseded by a later PR which updates figure creation to Makie's most recent API.

Issue #572, opened 2023-11-08
Compute mean of truncated distribution using a Burman series


Zapiano:
Now we are creating a `truncated(Normal(...))` distribution and then calling the `mean` function to find the mean (see`src/ecosystem/corals/growth.jl:311`). We can do this in a different way. Suppose we have a distribution `f(x)` and we want to find the mean within the interval `[a,b]`, then we could do integrate `x*f(x)` within the interval `[a,b]`. But also: instead of actually integrating, we could use the [Burman series of the error function](https://en.wikipedia.org/wiki/Error_function#B%C3%BCrmann_series) to approximate the value of this integral. This way we would save memory and time (I hope).

Zapiano:
@ConnectedSystems @Rosejoycrocker  Can you check if I missed something and if it makes sense?

Rosejoycrocker:
Just adding the link here for reference:

https://github.com/open-AIMS/ADRIA.jl/blob/a85588322bb4c6503092d91280907fe1c2b86ba5/src/ecosystem/corals/growth.jl#L310-L312

Rosejoycrocker:
> @ConnectedSystems @Rosejoycrocker Can you check if I missed something and if it makes sense?

Makes sense to me, is the intent not just to add the approximation but understand the order of truncation best suited for the application? For example by comparing the first, second and third order approximations, (k=1,2,3) and comparing their mean squared error to the Julia function equivalent.

Zapiano:
@ConnectedSystems @DanTanAtAims does it make sense to keep this issue opened?

ConnectedSystems:
Yes, the last set of work hasn't been merged yet.

PR #652, opened 2024-01-19
Speed up calculation of truncated normal mean and cdf

DanTanAtAims:
Instead of using built in mean, cdf, and truncation functions provided by DIstributions.jl, use the explicit formula of the truncated normal distributions.

Uses the approximated error function provided by SpecialFunctions.jl (added as dependency).

This does not resolve memory issues mentioned in issue #572.


ConnectedSystems:
General comments:

- There's a few other spots (at least two from memory) where the truncated normal mean is used. Please check the spec document I sent you earlier
- Could you add some high level tests? I think something that ensures the method produces values < some small error threshold compared to the original approach. This is just in case someone accidentally changes something in the future.
- There's some minor formatting issues we'll check when we chat again next.

ConnectedSystems (commented):
Some minor concerns, mostly to do with formatting.

ConnectedSystems:
Some performance notes:

Trial runs with the Moore domain (256 scenarios). 

Prior to the changes in this PR `bleaching_mortality()` takes ~38% of runtime with `adjust_DHW_distribution()` taking another ~13% (total of 51%).

Initial runtime was: ~1 min 40 secs (estimate - forgot to actually write it down)
Second run took: ~1 min 20 secs

Pasted image 20240119194729

With changes, `bleaching_mortality()` takes ~13% + 9% (22%), `adjusted_DHW_distribution()` taking another 12% (total of 34% of runtime).

Trajectories look as expected:

Pasted image 20240119210342


The image above is misleading as samples are biased towards guided scenarios, so here's one where I make sure there are equivalent number of samples for each scenario type.

Pasted image 20240119210951

DanTanAtAims:
I'll change interventions/seeding.jl to use the new calculation. 

Should I export the function truncated_normal_mean and truncated_normal_cdf so that in can be used in interventions/seeding.jl and tests?

ConnectedSystems:
Shouldn't need to export those, no - I think the functions need to be moved elsewhere (e.g., outside of `corals/growth.jl`) because it's not specific to growth.

ConnectedSystems:
Right now I'm thinking `Ecosystem.jl`, does that sound good to you?

DanTanAtAims:
I've come across a potential solution to the numerical issues we were having with the new implementation. I came across the [implementation here](https://github.com/cossio/TruncatedNormal.jl) that is currently the implementation Julia DIstribution.jl uses.

It turns out mean(Truncated(Normal(...)...)) uses this implementation but it appears that the calls to truncated() and Normal() are big slow downs from profiling. 

I implemented a nearly identical version of it excluding some error checks which we don't need. And benchmarked them and the result is nearly twice as fast as our previous attempt and agrees with mean(truncated(Normal())) which is unsurprising as its the same. The Benchmark results are attached. 1st benchmark is the original implementation and the second is the new implementation from Distributions.jl minus error checks.

tnm_bench

ConnectedSystems:
Hmm interesting - the potential maximum is about double your earlier implementation, but median is indeed half.
Either way, very happy with this as it's ~17% of runtime compared to the usual `mean(truncated())` approach.

Let me know when this is ready for a final review :)

DanTanAtAims:
The pull request is ready for final review. 

The Truncated Normal Mean has nearly equivalent accuracy to the in built implementation we were using but is much faster. 

The truncated normal cdf function has equivalent performance and I haven't found any speed ups that are stable for large deviations from the normal mean. However the new truncated_normal_cdf doesn't return NaN values for the large values unlike the inbuild cdf we were using.

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
         ),
     )
 end
+
+"""
+    truncated_standard_normal_mean(a::Float64, b::Float64)::Float64
+
+Calculates the mean of the truncated standard normal distribution. Implementation taken
+from Distribution.jl excluding unused error checks
```
ConnectedSystems, line 54: Add a reference section to the docstring pointing to exactly where the implementation was taken from.

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+
+# Arguements
```
ConnectedSystems, line 56: Fix typo

```suggestion
# Arguments
```

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+- 'a' : lower bound of the truncated distribution
+- 'b' : upper bound of the truncated distribution
```
ConnectedSystems, line 58: I prefer `lb` and `ub` if we're using shorthand for lower/upper bounds. This is general convention for bounds in scientific code.

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+
+# Returns
+- Float64, mean of the truncated standard normal distribution
```
ConnectedSystems, line 61: No need to make this a dot point if there's only a single return value.

General Julia style recommendation is to reduce repetition where possible/appropriate. As the return type is already documented in the function signature above, there's no need to repeat it here.

```suggestion
The mean of the truncated standard normal distribution
```

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+
```
ConnectedSystems, line 62: Delete stray new line at end of docstring

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+"""
+function truncated_standard_normal_mean(a::Float64, b::Float64)::Float64
+    if abs(a) > abs(b)
+        return - truncated_standard_normal_mean(-b, -a)
+    elseif (a == b) 
+        return a
+    end
+    mid = (a + b) / 2 
+    Δ = (b - a) * mid
+    a′ = a * invsqrt2
```
ConnectedSystems, line 72: Where is `invsqrt2` and `sqrthalfπ` from? 

They look like variables that aren't defined?

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+"""
+    truncated_normal_mean(
+        normal_mean::Float64, 
+        normal_stdev::Float64, 
+        lower_bound::Float64, 
+        upper_bound::Float64
+    )::Float64   
+
+Calculates the mean of the truncated normal distribution. Implementation is taken from
+Distributions.jl excluding unused error checks
```
ConnectedSystems, line 93: Same comment as above - add a references section and point to where the original implementation is

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+
+# Arguements
```
ConnectedSystems, line 95:
```suggestion
# Arguments
```

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+- 'normal_mean' : mean of the underlying (untruncated) normal distribution
+- 'normal_stdev' : standard deviation of the underlying (untruncated) normal distribution
+- 'lower_bound' : lower bound of the truncated normal distribution
+- 'upper_bound' : upper bound of the truncated normal distributionk
```
ConnectedSystems, line 99: Use backticks, not single quotations.
Just in case, it's the button directly above the tab key.

```suggestion
- `normal_mean` : mean of the underlying (untruncated) normal distribution
- `normal_stdev` : standard deviation of the underlying (untruncated) normal distribution
- `lower_bound` : lower bound of the truncated normal distribution
- `upper_bound` : upper bound of the truncated normal distributionk
```

src/ecosystem/Ecosystem.jl
ConnectedSystems: As mentioned elsewhere, use backticks, not single quotations.

src/ecosystem/Ecosystem.jl
ConnectedSystems: Go through and make the same changes for all the docstrings in this PR

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+
+# Returns
+- Float64, mean of the truncated normal distribution
+
```
ConnectedSystems, line 103: Delete stray new line here as well

src/ecosystem/Ecosystem.jl
DanTanAtAims: THey are precomputed constants from StatsFuns, I can replace them and remove the dependency?

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+"""
+function truncated_normal_mean(normal_mean::Float64, normal_stdev::Float64, lower_bound::Float64, upper_bound::Float64)::Float64
+
```
ConnectedSystems, line 106: I prefer no new line between function signature and its body.

In other words, do this:

```julia
function truncated_normal_mean(normal_mean::Float64, normal_stdev::Float64, lower_bound::Float64, upper_bound::Float64)::Float64
    alpha::Float64 = (lower_bound - normal_mean) / normal_stdev
    beta::Float64 = (upper_bound - normal_mean) / normal_stdev
```

Not this:

```julia
function truncated_normal_mean(normal_mean::Float64, normal_stdev::Float64, lower_bound::Float64, upper_bound::Float64)::Float64
    
    alpha::Float64 = (lower_bound - normal_mean) / normal_stdev
    beta::Float64 = (upper_bound - normal_mean) / normal_stdev
```

Do keep the return statement separate for clarity unless it's a very small function (e.g., 2 lines).

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+    alpha::Float64 = (lower_bound - normal_mean) / normal_stdev
+    beta::Float64 = (upper_bound - normal_mean) / normal_stdev
+
+    return normal_mean + truncated_standard_normal_mean(alpha, beta) * normal_stdev
+end
+
+"""
+    truncared_normal_cdf(x::Float64, normal_mean::Float64, normal_stdev::Float64,
```
ConnectedSystems, line 114: Typo in function signature?

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
...
+        for std in stdev
+            for width in bound_widths
+                calculated = ADRIA.truncated_normal_mean(
+                    mu, std, mu - width, mu + width
+                )
+                mean_diffs[ind] = abs(mu - calculated)
+                ind += 1
+            end
+        end
+    end
```
ConnectedSystems, line 27: Nesting like this is usually a code smell. Use iterators instead.

I think you want something like this:

```julia
for (mu, std, width) in Iterators.product(means, stdev, bound_widths)
    @info (mu, std, width)
end
```

FYI, `std` is a problematic choice of variable name as it's also a function name (`std()` from Statistics.jl).

I've used `stdev` instead for clarity. Suggest changing the current `stdev` to `stdevs` (note the `s` to pluralize).

Aside from that, I'm not sure what `ind` is for?

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
+using Test
+using ADRIA
+using ADRIA.Distributions
+
+@testset "Truncated normal mean" begin
+
+    means::Vector{Float64} = Vector{Float64}([0.0, 0.5, 1.5, 2.5, 4.0, 5.0])
+    stdev::Vector{Float64} = Vector{Float64}([0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0])
+    lower_bounds::Vector{Float64} = Vector{Float64}([0.0, 0.5, 1.5, 3.0, 7.0, 10.0])
+    bound_widths::Vector{Float64} = Vector{Float64}([0.25, 0.5, 1.0, 2.0, 5.0, 10.0])
```
ConnectedSystems, line 10: Is there reasoning behind these specific values?

Otherwise, I'd prefer we check a range of randomly generated combinations instead of (or in addition to) these.

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
...
+        end
+    end
+
+    @test all(cdf_diffs .< 1e-7) ||
+        "truncated normal cdf not equal to 0.0, 0.5, 1.0 at lower bound, mean, upper bound"
+
+    # evaluate cdf at proportion along domain of truncated normal
+    eval_cdf_at::Vector{Float64} = Vector{Float64}([
+        0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9
+    ])
```
ConnectedSystems, line 93:
```suggestion
    eval_cdf_at::Vector{Float64} = collect(0.1:0.1:0.9)
```

```julia
julia> collect(0.1:0.1:0.9)
# 9-element Vector{Float64}:
# 0.1
# 0.2
# 0.3
# 0.4
# 0.5
# 0.6
# 0.7
# 0.8
# 0.9
```

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
...
+    
+    n_checks = length(means) * length(stdevs) * length(n_stdevs) * 
+        length(n_stdevs) * 2 * length(eval_cdf_at)
+
+    cdf_diffs = Vector{Float64}(undef, n_checks)
```
ConnectedSystems, line 98: You might prefer

```suggestion
    cdf_diffs = zeros(n_checks)
```

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
...
+
+@testset "Truncated normal mean" begin
+
+    means::Vector{Float64} = Vector{Float64}([0.0, 0.5, 1.5, 2.5, 4.0, 5.0])
+    stdev::Vector{Float64} = Vector{Float64}([0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0])
+    lower_bounds::Vector{Float64} = Vector{Float64}([0.0, 0.5, 1.5, 3.0, 7.0, 10.0])
+    bound_widths::Vector{Float64} = Vector{Float64}([0.25, 0.5, 1.0, 2.0, 5.0, 10.0])
+
+    n_checks = length(means) * length(stdev) * length(bound_widths)
+    mean_diffs::Vector{Float64} = Vector{Float64}(undef, n_checks)
```
ConnectedSystems, line 13: You might prefer this

```suggestion
    mean_diffs::Vector{Float64} = zeros(n_checks)
```

Unless checking for `undef` is part of the test

test/Ecosystem.jl
```diff
@@ -0,0 +1,139 @@
...
+                        cdf_diffs[ind] = abs(custom_cdf - inbuilt)
+                        ind += 1
+                    end
+                end
+            end
+        end
+    end
+
+    @test all(cdf_diffs .< 1e-7) ||
+        "Implemented truncated normal cdf differs significantly from built in cdf"
```
ConnectedSystems, line 138:
```suggestion
        "Implemented truncated normal cdf differs significantly from built-in cdf"
```

test/runtests.jl
```diff
@@ -336,6 +336,7 @@ end
 include("clustering.jl")
 include("data_loading.jl")
 include("domain.jl")
+include("Ecosystem.jl")
```
ConnectedSystems, line 339: Not sure how I feel about this, but okay to leave as is.

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,119 @@ function EnvironmentalLayer(
...
+a = (lower_bound - μ)/σ
+b = (upper_bound - μ)/σ
+c = (x - μ)/σ
+
+# Arguments
+- x : value to evaluate cdf at
+- mean : mean of underlying normal distribution
+- stdev : standard deviation of underlying normal Distribution
+- lower_bound : lower bound of truncated distribution
+- upper_bound : upper bound of truncated distribution
```
ConnectedSystems, line 136: Add backticks to arguments please

ConnectedSystems, line 136: Use properly formed sentences

```suggestion
- lower_bound : lower bound of the truncated distribution
- upper_bound : upper bound of the truncated distribution
```

ConnectedSystems (changes requested):
Thanks @DanTanAtAims 

Some mostly style-related issues to fix then we're good to go.

src/ecosystem/Ecosystem.jl
ConnectedSystems: I suspected this, kind of surprised it's exported.

Options:

Either specify the package they're from for clarity:

```julia
a * StatsFun.invsqrt2
```

or add a note stating where they're from in the docstring (I'd be okay with this).

It's purely so future Dan/Takuya doesn't get confused.

ConnectedSystems:
> I can added a warning to the cdf function indicating possible loss of accuracy is we ever test these bounds?

Could you add it as a `@debug` level log please?

Same usage as `@info` and `@warn`

src/ecosystem/Ecosystem.jl
```diff
@@ -45,3 +46,108 @@ function EnvironmentalLayer(
         ),
     )
 end
+
+"""
+    truncated_standard_normal_mean(a::Float64, b::Float64)::Float64
```
ConnectedSystems, line 51: Forgot to update the function signature here

test/Ecosystem.jl
```diff
@@ -0,0 +1,95 @@
...
+            ub, mu, stdev, lb, ub
+        ))
+    end
+
+    @test all(cdf_diffs .< 1e-7) ||
+        "truncated normal cdf not equal to 0.0, 0.5, 1.0 at lower bound, mean, upper bound"
+
+    n_checks = 3000
+    
+    # check agreement with Distributions.jl implementation with bounds at most
```
ConnectedSystems, line 74: Just a note: I prefer comments that are on their own line be actual sentences (e.g., use capital letters at start of sentence, full stops, etc).

ConnectedSystems (approved):
Looking good, thanks!

ConnectedSystems:
Hmm, unfortunately we seem to be almost back at square 1.

image

Most of the time is spent in `erf()` and `logerf()`. Let's leave this for now and see if we can come up with anything else.

ConnectedSystems:
Just had a thought - it could be that that there is a small speed up, but the reason it's spending so much time in those functions is they're called a lot - it's just the nature of the use context. We can discuss more when Pedro gets back.

PR #660, opened 2024-01-29
Fix bleaching mortality 

ConnectedSystems:
Previous implementation was erroneously looping over all size classes for all functional groups, skipping only size classes 1 and 2 for the first functional group.

We now loop over all non-juvenile size classes for all functional groups as expected.

### DIFF
--- a/src/ecosystem/corals/growth.jl
+++ b/src/ecosystem/corals/growth.jl
@@ -637,7 +637,7 @@ end
 
 
 """
-    settler_cover(fec_scope::T, TP_data::AbstractMatrix{Float64}, leftover_space::T, α::V, β::V, basal_area_per_settler::V, potential_settlers::T)::T where {T<:Matrix{Float64},V<:Vector{Float64}}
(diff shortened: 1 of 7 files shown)


src/ecosystem/corals/growth.jl
```diff
@@ -667,18 +680,24 @@ function settler_cover(
 )::T where {T<:Matrix{Float64},V<:Vector{Float64}}
 
     # Determine active sources and sinks
-    valid_sources::BitVector = sum.(eachrow(TP_data)) .> 0.0
-    valid_sinks::BitVector = sum.(eachcol(TP_data)) .> 0.0
+    valid_sources::BitVector = vec(sum(conn, dims=2) .> 0.0)
+    valid_sinks::BitVector = vec(sum(conn, dims=1) .> 0.0)
 
     # Send larvae out into the world (reuse potential_settlers to reduce allocations)
+    @floop for src in findall(valid_sources)
```
Rosejoycrocker, line 687: Does this change need to be reverted now (following what we looked at)?

Rosejoycrocker (approved):
Looks good to me now

PR #661, opened 2024-01-29
Add note on properties of connectivity matrix to growth.jl

Rosejoycrocker:
Add comment on why the connectivity matrix does not sum to 1.0 along rows


ConnectedSystems:
I've cherry picked this change into #662

PR #665, opened 2024-01-29
Clean up scenario running methods

ConnectedSystems:
Clean up model run process for simplicity and conceptual validity.

Removes deprecated interface functions as well.

Addresses issue with NaN poisoning. 
RME datasets where empty wave datasets inadvertently resolves to NaN (due to max normalization with 0 values).
This has a flow-on effect to DHWs, resulting in DHW data being NaNs as well.

I've also corrected `set_factor_bounds!()` as it implies an in-place update will occur but this is only true if the function is called directly from the "main"/top-level module. If called from within another function, the updated `Domain` is not returned.


Rosejoycrocker (approved):
Looks good to me

PR #666, opened 2024-02-01
Return 0 in cases where lb > ub

ConnectedSystems:
As in title.

Currently causes an error to be raised as `m` never gets defined.

The "real" fix is to find out under what conditions lb becomes > ub (which will be a different PR).


ConnectedSystems:
Done, ready for final review.

DanTanAtAims (approved):
Looks good

PR #667, opened 2024-02-01
Clean up use of TP matrices to align with conceptualisation of connectivity

ConnectedSystems:
Connectivity matrices do not need to sum to 1, as it indicates the proportion of larvae that reach a sink location.

Transition probability matrices are now conceptualised as indicating the contribution of larvae from a source location to a sink location (i.e., the columns should sum to 1).

The changes in this PR clarifies the use of these terms and data.

### DIFF
--- a/src/ecosystem/connectivity.jl
+++ b/src/ecosystem/connectivity.jl
@@ -26,7 +26,7 @@ NOTE: Transposes transitional probability matrix if `swap == true`
 
 # Returns
 NamedTuple:
-- `TP_data` : Matrix, containing the transition probability for all sites
+- `conn` : Matrix, containing the connectivity for all sites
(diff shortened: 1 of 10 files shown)


src/ecosystem/connectivity.jl
```diff
@@ -26,7 +26,7 @@ NOTE: Transposes transitional probability matrix if `swap == true`
 
 # Returns
 NamedTuple:
-- `TP_data` : Matrix, containing the transition probability for all sites
+- `conn` : Matrix, containing the connectivity for all sites
```
Rosejoycrocker, line 29: Maybe add a note in brackets `(this accounts for larvae which do not settle and so rows may not sum to 1)` or something like that

Rosejoycrocker (approved):

PR #668, opened 2024-02-01
Adjustment to simulated events to address numerical errors

ConnectedSystems:
Fixes several issues to do with the simulation process.

ODE for growth was not accounting for mortality in corals moving in/out between size classes. At times the results could drive coral cover into negative values.

In addition, some mortality processes was inducing very small negative values.

We were also relying on a proportional readjustment to catch cases where the ODE were producing values > 1.0 (i.e., more coral cover than what was possible). This exceedance may happen more often and we rely on the proportional readjustment in lieu of constraining the ODE itself


Rosejoycrocker (approved):
Looks good to me, all tests passing

PR #669, opened 2024-02-02
Support updated domain structure

### LINKED ISSUE #580: Remove unused version check
This line checks if the datapackage version is bigger than "0.2.1" https://github.com/open-AIMS/ADRIA.jl/blob/916e947b9a7f22f1d69cd7137a1de2ba18bb530a/src/ExtInterface/ADRIA/Domain.jl#L274

BUT it's impossible to load a datapackage with that version, because it is not listed here: https://github.com/open-AIMS/ADRIA.jl/blob/916e947b9a7f22f1d69cd7137a1de2ba18bb530a/src/ADRIA.jl#L102

So we should just remove the check.

ConnectedSystems:
Changes to support a new Domain package structure to align with connectivity workflow.

Principally:

1. Connectivity files are now expected in a flat structure, with all files inside a single directory.
2. The geopackage file has an added field labelled "cluster_id" indicating cluster membership for each location. If this is not found, it defaults to `UNIQUE_ID`.

For point 1, it means restructuring the Domain from:

```bash
connectivity
|
|-- 2000
       |--conn_file1.csv
       |--conn_file2.csv
|-- 2001
       |--conn_file1.csv
       |--conn_file2.csv
|-- 2003
       |--conn_file1.csv
       |--conn_file2.csv
... etc ...
```

To:

```bash
connectivity
|--conn_2000.csv
|--conn_2000.csv
|--conn_2001.csv
|--conn_2001.csv
|--conn_2002.csv
|--conn_2002.csv
... etc ...
```

Closes #580 and addresses failing tests


src/ADRIA.jl
```diff
@@ -103,7 +103,7 @@ export ReefModDomain
 # export dims, ndims
 
 # List out compatible domain datapackages
-const COMPAT_DPKG = ["0.3.1", "0.4.0-rc", "0.4.0"]
+const COMPAT_DPKG = ["0.4.0-rc", "0.4.0", "0.5.0-rc", "0.5.0"]
```
Rosejoycrocker, line 106: Is a new Moore datapackage on the DS store?

ConnectedSystems: Not yet, I'll submit it later this afternoon. 

But if you make a copy and make the changes as outlined it'll work.

ConnectedSystems: Oh wait, I've got to push some changes for `ReefModDomain`

Rosejoycrocker: Thanks, I'm just running tests now but I'll run them again when you've pushed

ConnectedSystems: It won't matter, we don't have any tests covering the ReefModDomain because we don't want to commit a copy of that dataset into this repo.

We could mock up an example, but I don't have time for that today. Do you want to have a crack at it?

Rosejoycrocker: Yeah sure, I can do that. Just a test for loading the domain and running test scenarios?

ConnectedSystems: Yes, it's just the test domain restructured to match how RME's file structure

Rosejoycrocker: Tests are all passing, there's just a warning being thrown because the test domain still has `UNIQUE_ID` in the site data instead of `cluster_id`. I can change this when I add a test domain for ReefMod if that sounds good?

ConnectedSystems: Rose, that warning is there on purpose. It's because it can't find `cluster_id`, not because `UNIQUE_iD` is in the geopackage

Rosejoycrocker: Yes, so should I add a `cluster_id` to the test set?

Rosejoycrocker (approved):

src/ADRIA.jl
ConnectedSystems: > The reason I said "instead of" was that UNIQUE_ID was originally used for spatially constraining selected sites

What I'm reacting to is this:

> a warning being thrown *because* the test domain *still* has UNIQUE_ID in the site data *instead of* cluster_id.

This is expected behaviour. If `cluster_id` cannot be found, it will use `UNIQUE_ID`.
By saying "still" and "instead of", you seem to be suggesting that this is incorrect unexpected behaviour and that `UNIQUE_ID` should be removed.

> should I add a cluster_id to the test set?

Rather than adding `cluster_id`, perhaps it would be nicer if a test is added to ensure that the above warning is thrown?
I'll leave you to do it in a separate PR.

Rosejoycrocker: I understand this is expected behaviour, my point was that the test domain should have the preferred id column so that you don't get heaps of warnings when running tests

PR #671, opened 2024-02-05
Fix erroneous handling of location IDs

ConnectedSystems:
IDs are identifiers, not numeric values on which calculations can be performed.
Therefore they should be represented as strings, not numbers.


Zapiano (approved):

src/Domain.jl
```diff
@@ -31,7 +31,7 @@ function load_domain(path::String)::Domain
 end
 
 function unique_sites(d::Domain)::Vector{String}
-    return d.site_data[:, d.unique_site_id_col]
+    return d.site_data[:, d.cluster_id_col]
```
Zapiano, line 34: @ConnectedSystems Currently, the `:sites` dimension of the outcomes have cluster ids as labels instead of the sites/locations because of this change. Do you remember/know and can explain to me the reason for it?

PR #673, opened 2024-02-06
Fix error in domain loading when `cluster_id` is already in string format

Rosejoycrocker:
Previously tests were throwing an error due to the `cluster_id` column being already a string in the test domain. While loading the domain this column is converted to `Int` then `String` would throw an error due to trying to convert a string to integer. This has been fixed by adding a check to make sure `cluster_id` isn't already a string. `cluster_id` has also been added to the test set (currently a copy of `UNIQUE_ID`) to avoid throwing warnings whenever tests are run.


Zapiano:
@Rosejoycrocker Quick question: why does `cluster_id` needs to be converted to string? My first impression is that this should never happen, but I might be missing something here.

Zapiano (approved):
LGTM.

Only one question: why did you need to update those two test binary files?

Rosejoycrocker:
> LGTM.
> 
> Only one question: why did you need to update those two test binary files?

Previously the test site data files did not have a `cluster_id` column, so they threw a warning everytime they are used because `UNIQUE_ID` was used in place of `cluster_id`. I just made a column called `cluster_id` which is a copy of `UNIQUE_ID` to avoid this (the warning happens a lot so its just easier to read the read outcomes when it's not there).

PR #674, opened 2024-02-07
Switch to YAXArrays when loading ADRIA domain

Zapiano:
This is the first part of a series of PRs that, together, will consolidate the switch from `NamedDims` to `YAXArrays`. It comprises:

- [x] Loading ADRIA Domain (`load_domain()`)
- [x] Loading ReefMod Domain (`load_domain()`)

The tests will be handled in a separate PR.


src/io/inputs.jl
```diff
@@ -57,54 +57,50 @@ function load_scenarios(domain::Domain, filepath::String)::DataFrame
...
 """
-function load_nc_data(data_fn::String, attr::String, site_data::DataFrame)::NamedDimsArray
-    local loaded_data::NamedDimsArray
-
+function load_nc_data(
+    data_fn::String,
+    attr::String,
+    site_data::DataFrame;
+    dim_names::Vector=[],
+    dim_names_replace::Vector=[],
```
ConnectedSystems, line 70: No types specified here which doesn't align with the signature in the docstring?

src/io/inputs.jl
```diff
@@ -57,54 +57,50 @@ function load_scenarios(domain::Domain, filepath::String)::DataFrame
...
+
+        # Order sites
+        if :sites ∈ dim_names
+            sites_dim::Int64 = findfirst(x -> x == :sites, dim_names)
+            sites::Vector{String} = dim_labels[sites_dim]
+            sites_sort_idx = sortperm(sites)
+
+            dim_labels[sites_dim] = dim_labels[sites_dim][sites_sort_idx]
+
+            # Guarantees that even if sites dimension changes this sorting still works
```
ConnectedSystems, line 94: Minor change for clarity

```suggestion
            # Guarantees that sorting will work even if sites dimension changes
```

src/io/inputs.jl
```diff
@@ -57,54 +57,50 @@ function load_scenarios(domain::Domain, filepath::String)::DataFrame
...
+        end
+
+        axlist = Tuple(Dim{i[1]}(i[2]) for i in zip(dim_names, dim_labels))
 
         try
-            loaded_data = NamedDimsArray(data; zip(dim_names, dim_labels)...)
+            return YAXArray(axlist, data)
         catch err
             if isa(err, KeyError)
                 n_sites = size(data, 2)
```
ConnectedSystems, line 106: The reason for this to be wrapped in a `try/catch` isn't clear to me.

Additionally, there is no return in the case of a `KeyError` either so will lead to an issue.

src/io/inputs.jl
```diff
@@ -166,48 +160,47 @@ function _site_labels(nc_file::NetCDF.NcFile)::Vector{String}
...
+    _dim_names_replace = [:covers => :species, :reef_siteid => :sites]
+    data = load_nc_data(data_fn, attr, site_data; dim_names_replace=_dim_names_replace)
 
-    return data
+    return _convert_abs_to_k(data, site_data)
+end
+function load_cover(n_species::Int64, n_sites::Int64)::YAXArray
+    @warn "Using random initial coral cover"
+    axlist = (Dim{:species}(1:(n_species)), Dim{:sites}(1:n_sites))
+    return YAXArray(axlist, rand(Float32, n_species, n_sites))
```
ConnectedSystems, line 177: We spoke about having a convenience constructor already, just noting it down here.

ConnectedSystems (changes requested):
Looks okay to me overall, just a few minor things to address plus updating the auto-formatted style as we discussed.

src/io/inputs.jl
Zapiano: I removed the try/catch because it wasn't clear to me either (it was there before and I just kept it). Let me know if you have any problems with that.

ConnectedSystems: Thank you - it was there before for a reason, but you've removed the reason for it 😉

src/ExtInterface/ADRIA/Domain.jl
```diff
@@ -188,30 +188,19 @@ function Domain(
...
+    waves_params = ispath(wave_fn) ? (wave_fn, "Ub") : (timeframe, conn_ids)
+    waves = load_env_data(waves_params...)
 
-    cyclone_mortality = if ispath(cyclone_mortality_fn)
-        load_cyclone_mortality(cyclone_mortality_fn)
-    else
-        load_cyclone_mortality(timeframe, site_data)
-    end
+    cyc_params = ispath(cyclone_mortality_fn) ? (cyclone_mortality_fn,) : (timeframe, site_data)
+    cyclone_mortality = load_cyclone_mortality(cyc_params...)
```
ConnectedSystems, line 203: Very elegant!

src/io/inputs.jl
Zapiano: @ConnectedSystems 

I'm still thinking if it's necessary/useful to have this constructor. Let me give an example creating the same array using both `NamedDimsArrays` and `YAXArrays`:

```
nd = NamedDimsArray(rand(Int8, 2,10), row=[:a, :b], col=10:10:100)
ya = YAXArray((Dim{:row}([:a, :b]), Dim{:col}(10:10:100)), rand(Int8, 2, 10))
```

That's for every axes I need to write 7 characters more (`Dim{:}()` versus `=`) with YAXArray syntax. So, if we are creating an array with two axis maybe with `NamedDimsArrays` we wouldn't need more than one line and with `YAXArrays` we would need two (or more, depending on the formatting). But for less or for more than that, I think it's almost the same: if we have 4 dimensions, even with `NamedDimsArrays` it's better (for readability) to create the axlist, the data, and put it all together when creating the array.

ALSO 

1. `YAXArrays` allows us to also pass a metadata param (`props`), and with that even with the constructor it will be better to create these things (data, axlist and props) separately and use it all to create the array. 
2. We will need to give support to this constructor, test it, etc, so, unless it really helps, I tend to think we don't need it.


In any case, I would like to hear your thoughts on that before I do anything :)

ConnectedSystems: My thinking regarding suggesting an alternate constructor is about usability and convenience.

I don't suggest having a constructor for every single use case, just the common ones. If there are more specialised contexts where it makes sense to create YAXArrays directly, then we should by all means do so.

If there is a situation where you need to create a set of YAXArrays with named rows and columns as illustrated by the example above, then typing `Dim{:row}([:a, :b])` over and over again seems really annoying, repetitive, and redundant. I'd much prefer the below **for the specific use case you have**.

```julia
function RowColYAX(data, rows, cols; kwargs...)
    return YAXArray((Dim{:row}(rows), Dim{:col}(cols), data; kwargs...))
end

# ... in different locations around the codebase ...
ya = RowColYAX(rand(Int8, 2, 10), [:a, :b], 10:10:100)
ya2 = RowColYAX(rand(Int8, 2, 10), [:c, :d], 10:100:1000)
ya3 = RowColYAX(rand(Int8, 3, 10), [:x, :y, :z], 1:10)

# I'd prefer having to use the above than the below

ya = YAXArray((Dim{:row}([:a, :b]), Dim{:col}(10:10:100)), rand(Int8, 2, 10))
ya2 = YAXArray((Dim{:row}([:c, :d]), Dim{:col}(10:100:1000)), rand(Int8, 2, 10))
ya3 = YAXArray((Dim{:row}([:x, :y, :z]), Dim{:col}(10:10:100)), rand(Int8, 3, 10))
```

Now if you're saying there are no common use cases, then yes I agree, there's no point in having another generic constructor.

src/ExtInterface/ReefMod/Domain.jl
```diff
@@ -104,29 +103,24 @@ function load_domain(::Type{ReefModDomain}, fn_path::String, RCP::String)::ReefM
...
+    dim_species = Dim{:species}(1:6)
+    dim_scenarios = Dim{:scenarios}([1])
+
+    wave_axlist = (dim_timesteps, dim_locs, dim_scenarios)
+    wave_data = zeros(length(timeframe_range), nrow(site_data), 1)
+    wave_scens::YAXArray{Float64} = YAXArray(wave_axlist, wave_data)
+
+    cyc_axlist = (dim_timesteps, dim_locs, dim_species, dim_scenarios)
+    cyc_data = zeros(length(timeframe_range), nrow(site_data), 6, 1)
+    cyc_scens::YAXArray{Float64} = YAXArray(cyc_axlist, cyc_data)
```
ConnectedSystems, line 123: Feels like this could be a good case of a generic constructor for all Domain types to use

src/yaxarrays/YAXArrays.jl
```diff
@@ -0,0 +1,36 @@
...
+function axes_names(cube::YAXArray)
+    return name.(cube.axes)
+end
+
+"""
+    sort(cube::YAXArray, axis_name::Symbol)
+
+Sorts axis labels of a given YAXArray datacube.
+"""
+function sort(cube::YAXArray, axis_name::Symbol)::YAXArray
```
ConnectedSystems, line 12: Not a big fan of the module/filename.

We already have an `io` folder, can we not move it there?

Zapiano: Even with the YAXArray constructor?

src/yaxarrays/YAXArrays.jl
```diff
@@ -0,0 +1,36 @@
...
+    axis_idx::Int64 = findfirst(x -> x == axis_name, _axes_names)
+    selector[axis_idx] = labels_sort_idx
+
+    return cube[selector...]
+end
+
+"""
+    copy(cube::YAXArray)::YAXArray
+
+Use julia's Base.copy function to create a copy of an YAXArray
```
ConnectedSystems, line 30: This statement is technically incorrect - you're not using Base.copy.

Better to state what the method does / is for.

Zapiano: Hum, inside this function I'm using Base.copy to create a copy of the data inside the YAXArray. I can rewrite this to better explain what its being done there.

src/yaxarrays/YAXArrays.jl
```diff
@@ -0,0 +1,36 @@
...
+"""
+function copy(cube::YAXArray)::YAXArray
+    new_data = copy(cube.data)
+    new_axlist = Tuple(ax for ax in deepcopy(cube.axes))
+    return YAXArray(new_axlist, new_data)
+end
```
ConnectedSystems, line 36: Hmm, you're assigning a copy of the data to a variable then passing it into YAXArray.

If my understanding of Julia is correct, it'll temporarily allocate double the memory, and then at some point it will trigger the GC.

I think this is better (and more concise, without losing readability):

```julia
function copy(cube::YAXArray)::YAXArray
    new_axlist = Tuple(ax for ax in deepcopy(cube.axes))
    return YAXArray(new_axlist, copy(cube.data))
end
```

src/yaxarrays/YAXArrays.jl
ConnectedSystems: Oh I see - still, probably just something succinct like "copy a data cube" is fine

src/io/inputs.jl
Zapiano: I have created the DataCube and ZeroDataCube constructors for some common cases. Let me know if you have any suggestions (other than allowing passing a `props`, in the future) or if I've addressed your suggestion here. (:

PR #676, opened 2024-02-08
Additional setup instructions for code style and formatting

ConnectedSystems:
Additional detail on setting up VS Code to match project code style and formatting.

JuliaFormatter options may change or be adjusted later.

### DIFF
--- a/docs/src/development/development_setup.md
+++ b/docs/src/development/development_setup.md
@@ -14,6 +14,7 @@ In such cases, use the alternate MSIX App installer (https://install.julialang.o
 
 - It is recommended that the [juliaup](https://github.com/JuliaLang/juliaup) tool be used to ease managing Julia versions.
 - We recommend [VS Code](https://code.visualstudio.com/) with its Julia extension when developing ADRIA.
+- Install the VS Code Julia Formatter extension (note: **not** the JuliaFormatter.jl package).
 - We also recommend the built-in Julia REPL within VS Code be used (see the notes below).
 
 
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
@DanTanAtAims would you mind taking the lead on the review, given you set things up most recently?

Was there anything else you did that is missing here?

docs/src/development/development_setup.md
```diff
@@ -119,6 +120,31 @@ for further details.
...
+```bash
+style=BlueStyle(), indent=4, margin=256, always_for_in=true, for_in_replacement="∈", whitespace_typedefs=true, import_to_using=true, align_struct_field=true, align_assignment=false, align_conditional=true, align_pair_arrow=false, normalize_line_endings="unix", align_matrix=true, join_lines_based_on_source=true, indent_submodule=true, surround_whereop_typeparameters=false, yas_style_nesting=true
+```
+
+Next, search for `trim` and ensure the following options are all ticked/enabled:
+
+- Files: Trim Final Newlines
+- Files: Trim Trailing Whitespace
+- Editor: Trim Auto Whitespace
+
```
DanTanAtAims, line 147:
```Suggestion

or, as suggested by Blue Style Guide, append the following to your settings.json file

{
    "[julia]": {
        "editor.detectIndentation": false,
        "editor.insertSpaces": true,
        "editor.tabSize": 4,
        "files.insertFinalNewline": true,
        "files.trimFinalNewlines": true,
        "files.trimTrailingWhitespace": true,
        "editor.rulers": [92],
    },
}
```

Are we using margins of 256 instead of 92?

ConnectedSystems: The settings UI just edits the settings.json file.

We are using 92 character limit, and I do indeed have that set as a VS Code option, not a Julia Formatter option. I'll update.

ConnectedSystems: Done

ConnectedSystems: Note, I've opted not to include instructions to edit the settings.json file directly.

Of course, if people are comfortable doing so, that's fine, but given the Julia Formatter options are equivalent and we may adjust these in future, I think best to just give one set of canonical instructions.

DanTanAtAims:
Looks good, would you add an optional comment about a spell checking extension if it's a common problem for others? I assume correct spelling is almost definitely an assumption for all comments especially docstrings in all codebases, so most likely not required in the style guide.

Also my spell-checker configuration is quite poor right now and has a lot of false positives and missed negatives so I wouldn't add it here.

ConnectedSystems:
I think that can be informally communicated as needed. 

Probably best to assume people will be proactive about managing that for themselves if they feel it is an issue 😉

Zapiano (approved):
LGTM.

Rosejoycrocker (approved):
LGTM :)

ConnectedSystems:
Back to you Daniel

PR #679, opened 2024-02-11
Fix readjustment of number of slices bug in RSA

Rosejoycrocker:
Fixes readjustment of number of slices bug in RSA (partially addresses #615) and changes output of RSA to `YAXArrays.Dataset` type.

### DIFF
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -155,8 +155,8 @@ function ADRIA.viz.tsa(
 end
 
 """
-    ADRIA.viz.rsa(rs::ResultSet, si::YAXArrays.Dataset, factors::Vector{String}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
-    ADRIA.viz.rsa!(f::Union{GridLayout,GridPosition}, rs::ResultSet, si::YAXArrays.Dataset, factors::Vector{String}; opts::Dict=Dict(), axis_opts::Dict=Dict())
+    ADRIA.viz.rsa(rs::ResultSet, si::Dataset, factors::Vector{String}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
+    ADRIA.viz.rsa!(f::Union{GridLayout,GridPosition}, rs::ResultSet, si::Dataset, factors::Vector{String}; opts::Dict=Dict(), axis_opts::Dict=Dict())
 
 Plot regional sensitivities of up to 30 factors.
 
@@ -175,7 +175,7 @@ Makie figure
 function ADRIA.viz.rsa!(
     g::Union{GridLayout,GridPosition},
     rs::ResultSet,
-    si::YAXArrays.Dataset,
+    si::Dataset,
     factors::Vector{Symbol};
     opts::Dict=Dict(),
     axis_opts::Dict=Dict(),
@@ -273,7 +273,7 @@ function ADRIA.viz.rsa!(
 end
 function ADRIA.viz.rsa(
     rs::ResultSet,
-    si::YAXArrays.Dataset,
+    si::Dataset,
     factors::Vector{Symbol};
     opts::Dict=Dict(),
     fig_opts::Dict=Dict(),
--- a/ext/AvizExt/viz/sensitivity.jl
+++ b/ext/AvizExt/viz/sensitivity.jl
@@ -1,6 +1,6 @@
 using Statistics
 using Printf
-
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
Does the switch to YAXArray break the ability to plot results? 
I don't see any compatibility layer in place.

Rosejoycrocker:
> Does the switch to YAXArray break the ability to plot results? I don't see any compatibility layer in place.

There are also changes to `viz.rsa()` in this PR so plotting works

ext/AvizExt/viz/viz.jl
```diff
@@ -1,7 +1,7 @@
 using Makie, DataFrames
 
 using ADRIA: ResultSet, metrics.metric_label, analysis.col_normalize, model_spec
-using NamedDims, AxisKeys
+using NamedDims, AxisKeys, YAXArrays, DimensionalData
```
ConnectedSystems, line 4: Is the import of DimensionalData simply to have the `At` method?

If so, I'd prefer it be imported via YAXArrays:

```julia
import YAXArrays.DD: At
```

src/analysis/sensitivity.jl
```diff
@@ -5,7 +5,9 @@ using Logging
 using
     AxisKeys,
     NamedDims,
-    StaticArrays
+    StaticArrays,
+    YAXArrays,
+    DimensionalData
```
ConnectedSystems, line 10: As with previous comment - if this is for `At`, import it directly.

ConnectedSystems:
> > Does the switch to YAXArray break the ability to plot results? I don't see any compatibility layer in place.
> 
> There are also changes to `viz.rsa()` in this PR so plotting works

Thanks Rose, clearly I'm going blind.

ext/AvizExt/viz/sensitivity.jl
```diff
@@ -175,7 +175,7 @@ Makie figure
 function ADRIA.viz.rsa!(
     g::Union{GridLayout,GridPosition},
     rs::ResultSet,
-    si::NamedDimsArray,
+    si::YAXArrays.Dataset,
```
ConnectedSystems, line 178: Why not import `Dataset` directly so prefacing it with `YAXArrays` is not necessary?

Or is there a different `Dataset` type that is causing a conflict?

src/analysis/sensitivity.jl
```diff
@@ -451,78 +455,107 @@ ADRIA.sensitivity.rsa(X, y; S=10)
...
 function rsa(
-    X::DataFrame, y::AbstractVector{<:Real}, model_spec::DataFrame; S::Int64=10
-)::NamedDimsArray
+    X::DataFrame, y::AbstractVector{<:Real}, factors::Vector{Symbol}, model_spec::DataFrame;
+    S::Int64=10,
+)::YAXArrays.Dataset
     N, D = size(X)
 
-    X_di = zeros(N)
+    X_i = zeros(N)
```
ConnectedSystems, line 463: There was a reason for X_di ($X_{d, i}$) but that may now be lost to the sands of time...

Rosejoycrocker: Ah makes sense. Should I revert or leave as is?

ConnectedSystems: Leave as is

ext/AvizExt/viz/viz.jl
Rosejoycrocker: I think I just took this from the YAXArray docs and forgot to remove, I've removed it now

ext/AvizExt/viz/sensitivity.jl
Rosejoycrocker: I've replaced with `Dataset`

Rosejoycrocker:
> Some minor comments for consideration

I think I've addressed your comments, If you're happy with this I'll do another PR with similar changes for `outcome_map`

ConnectedSystems:
Busy with something else at the moment, would you mind attaching a figure here to demonstrate its working please?

Rosejoycrocker:
> Busy with something else at the moment, would you mind attaching a figure here to demonstrate its working please?

Here's an example 
sensitivity_rsa_ex

ConnectedSystems:
The peak is telling you that there is high variability in that factor space.

In other words, the outcome assessed is highly sensitive to that region of factor space, which corresponds with no intervention.

Rosejoycrocker:
> The peak is telling you that there is high variability in that factor space.
> 
> In other words, the outcome assessed is highly sensitive to that region of factor space, which corresponds with no intervention.

Makes sense

ConnectedSystems (approved):
LGTM

PR #449, opened 2023-09-04
Functions to plot spatial distributions of site selection frequencies

Rosejoycrocker:
Adds function to viz extension which plots the spatial distribution of location selection frequency. PR #431 must be merged to main first.

Example usage is added to the viz documentation.

Color map specification is also added to the inputs of `ADRIA.viz.map()` to allow different colour maps to be specified in spatial plots. Default is set to `:GrayC`.

PR #685, opened 2024-02-13
restructured documentation image directories

### LINKED ISSUE #567: Reorganize files/images for documentation
Image files for the documentation are placed in one folder with no organization (everything in the one folder). Better to reorganize these into sub-directories for each page where relevant.

> I dislike having `example` / `ex` / `eg` in the filename. A better organisation approach would be to have images in separate directories for each page.
> e.g., `assets/imgs/analysis/selected_locations_map.png`
> I will raise a separate issue for this to be tackled.

_Originally posted by @ConnectedSystems in https://github.com/open-AIMS/ADRIA.jl/pull/449#discussion_r1384217217_

DanTanAtAims:
In the issue it mentions updating the filename within the documentation here

```julia
save("ranks_plot_ex.png", tf)  # this filename has to be updated
```
was the idea to change it to the following

```julia
save("analysis/ranks_plot_ex.png", tf)
```
closes #567


ConnectedSystems:
I think it's okay to _not_ refer to the directory when saving plots, but *remove* any reference to "example": such as `ex`, `eg`, etc. 

The idea is the user should be able to run the examples directly with as little setup as possible (e.g., creating directories).

DanTanAtAims:
It looks like rose standardised the filenames in the original pull request (#449) as all example suffixes have been removed. 

If you are happy with the new file structure this should be ready for merge.

ConnectedSystems:
Awesome, thanks

ConnectedSystems (approved):
All is well!

PR #687, opened 2024-02-14
Truncated normal changes

### LINKED ISSUE #572: Compute mean of truncated distribution using a Burman series

DanTanAtAims:
Added truncated normal changes to speedup calculation of truncated normal mean and cdf.

The test cases are more relaxed to account for the loss of accuracy resulting from the approximation. 

The if statement warning about accuracy is now more specific in what it detects requiring both truncated bounds be on the same side of the mean as this was the source of cancellation errors. During the test runs done bellow, the warning was not encoutnered

The following plot is **without changes** contained in this pull request.
no_changes_plot

The following plot is **with changes** contained in this pull request.
with_changes_plot

closes #572

### DIFF
--- a/src/ecosystem/Ecosystem.jl
+++ b/src/ecosystem/Ecosystem.jl
@@ -48,6 +48,62 @@ function EnvironmentalLayer(
 end
 
 """
+    rational_erf(x::Float64)::Float64
+
+Rational approximation of the error function only using elementary functions [1].
+Maximum error of 1.5 × 10^{-7}.
+
+# References
+1. Abramowitz, Milton; Stegun, Irene Ann, eds. (1983) [June 1964]. "Chapter 7". 
+   Handbook of Mathematical Functions with Formulas, Graphs, and Mathematical Tables. 
+   Applied Mathematics Series. Vol. 55 (Ninth reprint with additional corrections of 
+   tenth original printing with corrections (December 1972); first ed.). Washington D.C.; 
+   New York: United States Department of Commerce, National Bureau of Standards; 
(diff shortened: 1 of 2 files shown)


src/ecosystem/Ecosystem.jl
```diff
@@ -81,9 +137,9 @@ function truncated_standard_normal_mean(lb::Float64, ub::Float64)::Float64
 
     m = ub
     if lb ≤ 0 ≤ ub
-        m = expm1(-Δ) * exp(-lb^2 / 2) / erf(ub′, lb′)
+        m = expm1(-Δ) * exp(-lb^2 / 2) / (rational_erf(lb′) - rational_erf(ub′))# erf(ub′, lb′)
```
ConnectedSystems, line 140:
```suggestion
        m = expm1(-Δ) * exp(-lb^2 / 2) / (rational_erf(lb′) - rational_erf(ub′))
```

ConnectedSystems (commented):
Thanks Dan, one minor change to remove defunct commented code.

PR #688, opened 2024-02-14
Allow user to set whatever bounds they wish

ConnectedSystems:
Recently, when trying to do some scenario runs to sweep through larger factor bounds, I ran into an issue where certain distributions cannot be set.

This PR allows the user to set whatever bounds they wish, so long as they specify the necessary distribution parameters.

Factor types and their expected parameters are:

- `DiscreteOrderedUniformDist` : `(min, max, step)`, 
- `TriangularDist` : `(min, max, peak)`

```julia
dom = ADRIA.load_domain(some_domain)
dom = ADRIA.set_factor_bounds(
    dom,
    N_seed_TA=(Int(1e6), Int(5e10), Int(1e6)),
    N_seed_CA=(Int(1e6), Int(5e10), Int(1e6)),
    N_seed_SM=(Int(1e6), Int(5e10), Int(1e6)),
    a_adapt=(0.0, 30.0, 0.0),
    seed_freq=(1, 5)
)
```

### DIFF
--- a/src/factors/Factors.jl
+++ b/src/factors/Factors.jl
@@ -57,7 +57,7 @@ function _check_has_required_info(kwargs::NT) where {NT <: NamedTuple}
     @assert haskey(kwargs, :description) "Missing factor field `description`"
 
     param_dist_types = [
-        "continuous", "ordered categorical", "unordered categorical", "discrete"
+        "continuous", "ordered categorical", "unordered categorical", "ordered discrete", "discrete"
     ]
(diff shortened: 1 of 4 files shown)


ConnectedSystems:
The test is currently a stub, but at least covers the setting of parameter values.
I can confirm the factors are indeed set within bounds (we have other tests that ensure that anyway) and scenarios do run.

Rosejoycrocker (approved):
Looks good to me

PR #689, opened 2024-02-14
Yax loaddomain

Zapiano:
This is the first part of a series of PRs that, together, will consolidate the switch from NamedDims to YAXArrays. It comprises:

- [x] Loading ADRIA Domain (load_domain())
- [x] Loading ReefMod Domain (load_domain())

The tests will be handled in a separate PR.

PR #690, opened 2024-02-14
Switch to YAXArrays when runnin scenarios

Zapiano:
As in title. This one should be merged AFTER [PR 689](https://github.com/open-AIMS/ADRIA.jl/pull/689).

### DIFF
--- a/src/decision/dMCDA.jl
+++ b/src/decision/dMCDA.jl
@@ -3,7 +3,6 @@ module decision
 
 using StatsBase
 using Distances, DataFrames
-using DimensionalData
 using Combinatorics
 using JMcDM
 using InteractiveUtils: subtypes
--- a/src/decision/dMCDA.jl
+++ b/src/decision/dMCDA.jl
@@ -3,9 +3,11 @@ module decision
 
 using StatsBase
 using Distances, DataFrames
+using DimensionalData
 using Combinatorics
 using JMcDM
 using InteractiveUtils: subtypes
+using YAXArrays
 using ADRIA: Domain, EcoModel, n_locations, site_area, site_k_area
 
 struct DMCDA_vars  # {V, I, F, M} where V <: Vector
@@ -88,7 +90,7 @@ Constuctors for DMCDA variables.
 """
 function DMCDA_vars(
     domain::Domain,
-    criteria::NamedDimsArray,
+    criteria::YAXArray,
     site_ids::AbstractArray,
     leftover_space::AbstractArray,
     area_to_seed::Float64,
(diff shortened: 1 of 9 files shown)


ConnectedSystems:
> As in title. This one should be merged AFTER [PR 689](https://github.com/open-AIMS/ADRIA.jl/pull/689).

FYI, you don't need to manually create the links, just tag the PR ID after a hash, e.g., `#689` -> #689

src/scenario.jl
```diff
@@ -89,7 +89,7 @@ function run_scenarios(
     setup()
 
     # Sort RCPs so the dataframe order match the output filepath
-    RCP = Base.sort(RCP)
+    RCP = sort(RCP)
```
ConnectedSystems, line 92: Was this intentional?

src/scenario.jl
```diff
@@ -329,19 +335,19 @@ function run_model(domain::Domain, param_set::DataFrameRow)::NamedTuple
...
-function run_model(domain::Domain, param_set::NamedDimsArray)::NamedTuple
+function run_model(domain::Domain, param_set::YAXArray)::NamedTuple
     p = domain.coral_growth.ode_p
     corals = to_coral_spec(param_set)
     cache = setup_cache(domain)
 
     # Set random seed using intervention values
     # TODO: More robust way of getting intervention/criteria values
-    rnd_seed_val::Int64 = floor(Int64, sum(param_set(!=("RCP"))))  # select everything except RCP
+    rnd_seed_val::Int64 = floor(Int64, sum(param_set[Where(x -> x != "RCP")]))  # select everything except RCP
```
ConnectedSystems, line 345: Not a big deal, but does this work?

```julia
not_RCP = param_set.factors .!= "RCP"
param_set[factors=not_RCP]
```

src/scenario.jl
```diff
@@ -329,19 +335,19 @@ function run_model(domain::Domain, param_set::DataFrameRow)::NamedTuple
...
     Random.seed!(rnd_seed_val)
 
-    dhw_idx::Int64 = Int64(param_set("dhw_scenario"))
-    wave_idx::Int64 = Int64(param_set("wave_scenario"))
-    cyclone_mortality_idx::Int64 = Int64(param_set("cyclone_mortality_scenario"))
+    dhw_idx::Int64 = Int64(param_set[At("dhw_scenario")])
```
ConnectedSystems, line 348: With these `param_set[At(...)]`, I think `param_set.factor_name` should work and save a few keystrokes.

src/scenario.jl
ConnectedSystems: The reason is I prefer to avoid lambdas (comparatively slow).

src/scenario.jl
Zapiano: Yes. This is how it is on the main branch currently. I had changed to `Base.sort(RCP)` in #689 but that is not necessary anymore.

src/scenario.jl
Zapiano: I think that doesn't work. What I *can* do, is `param_set.factors` , becausa `:factors` is the name of an axis - but that doesn't return what we want here. The names of the factors are *labels* of that axis. I can think of two ways of selecting one single axis element in this case:

```
param_set[param_set.factors .== "dhw_scenario"].data[1]
param_set[At("dhw_scenario")]
```

image

The former is useful when we have more axes (I'm not sure if/how the later works with more axes).

src/decision/dMCDA.jl
```diff
@@ -3,9 +3,11 @@ module decision
 
 using StatsBase
 using Distances, DataFrames
+using DimensionalData
```
ConnectedSystems, line 6: Is this import necessary?

If it's just for the `At` method, I prefer it's imported directly:

`import YAXArrays.DD: At`

ConnectedSystems (approved):
Thanks

src/decision/dMCDA.jl
Zapiano: Apparently it doesn't make any difference (I've removed it and everything worked fine).

PR #691, opened 2024-02-15
Adjustments to allow loading newest Domain datasets

ConnectedSystems:
Updated Domain for latest data including connectivity.

Includes changes to expected Domain directory layout.

This should be merged after `yax-rscens` (#690, rebase before reviewing)


src/io/inputs.jl
```diff
@@ -67,6 +68,21 @@ function load_nc_data(
     attr::String;
     dim_names::Vector{Symbol}=Symbol[],
     dim_names_replace::Vector{Pair{Symbol,Symbol}}=Pair{Symbol,Symbol}[],
+)::YAXArray
+    local data
+    try
+        data = sort_axis(Cube(data_fn), :locations)
+    catch
+        return fallback_nc_data(data_fn, attr; dim_names, dim_names_replace)
```
Zapiano, line 76: I didn't understand this. 

And, unless something's changed in the way DHW and Waves data is saved, I think this may not work. If you do `Cube(data_fn)` where `data_fn` is the path to the DHW file, you won't get and error, but the YAXArray will be formatted wrong.

Zapiano (commented):
Only one comment to address then we can merge.

src/io/inputs.jl
ConnectedSystems: There is indeed a change to the netCDFs.

`Cube` fails to load the old file, so I just fallback to the old approach in that case.

Not a huge issue as I'm not promising backward compatibility, but it's there during this awkward transition phase

ConnectedSystems: This is the process that creates the file being handled:
https://github.com/open-AIMS/rrap-dg/blob/main/rrap_dg/initial_coral_cover/icc.jl

You can see the old NetCDF.jl based approach commented out at the end.

Zapiano: But the problem is: this function is used to load `dhw` and `waves` data, but if I'm not mistaken, the call to `Cube` won't raise an error in these cases, it will only behave wrongly. So maybe we should have a separate function to load coral cover..?

ConnectedSystems: dhw and wave data hasn't changed formats, only initial coral cover.

Sorry if that wasn't clear

ConnectedSystems: > So maybe we should have a separate function to load coral cover..?

It's only here temporarily while we sort out/finalise our preprocessing steps. Loading the .`nc` data themselves should be all the same , the special cases are handled by the high level functions (`load_wave_data`, etc)

ConnectedSystems: > When we use Cube to load DHW and waves data, it doesn't throw an error but the data cube generated is wrong.

Oh no wait, I think I remembered it backwards - the wave/DHW data fails to load with YAXArrays so we fall back to the older approach the does load the data correctly.

ConnectedSystems: Let's discuss tomorrow

ConnectedSystems:
Confirming this is working as intended @Zapiano 

Coral cover:

image

DHW:

image

Cyclones don't hit this code branch (as you know) and waves are empty at the moment.

PR #693, opened 2024-02-15
Update rank metrics to use YAXArrays

ConnectedSystems:
Tests and use cases need to be checked to ensure full functionality is as expected.

To be merged after #690 (rebase before reviewing)


Zapiano (approved):
Thanks :)

PR #697, opened 2024-02-16
updated data loading test for migration to YAXArrays

DanTanAtAims:


test/data_loading.jl
```diff
@@ -1,11 +1,12 @@
 using ADRIA, ADRIA.DataFrames, ADRIA.CSV
-using ADRIA.NamedDims, ADRIA.AxisKeys
+using ADRIA.YAXArrays, ADRIA.AxisKeys
```
ConnectedSystems, line 2: Do we need AxisKeys imported with this update?

test/data_loading.jl
```diff
@@ -30,8 +31,8 @@ end
     conn_details = ADRIA.site_connectivity(conn_files, unique_site_ids)
 
     conn = conn_details.conn
-    @test all(axiskeys(conn, 1) .== axiskeys(conn, 2)) || "Site order does not match between rows/columns."
-    @test all(axiskeys(conn, 2) .== site_data.reef_siteid) || "Sites do not match expected order."
+    @test all(caxes(conn)[1].val.data .== caxes(conn)[2].val.data) || "Site order does not match between rows/columns."
```
ConnectedSystems, line 34: You want to axis data I think here.

Would this work?

```julia
d1, d2 = axes(conn)
d1.dim .== d2.dim
```

DanTanAtAims: Yes this works, and its clearer I'll commit

PR #700, opened 2024-02-19
Yax test update

DanTanAtAims:
Partial update for tests to use YAXArrays. Should be merged after #693.


Zapiano:
Hi @DanTanAtAims , I'm confused with the name of this PR. Is this supposed to only update tests? Because it seems to be doing more than this. Is that correct?

DanTanAtAims:
It contains a few relatively shallow type changes to so that I could continue altering some types. the `rank_to_frequencies` was more involved than I though but allowed me to progress further. But yes, I should start working on non-test conversions.

src/decision/location_selection.jl
```diff
@@ -202,17 +205,18 @@ Note: By default, ranks are aggregated over time where `ranks` is a 3-dimensiona
...
-        [1:size(ranks, n) for n in dn if n != :scenarios],
-        [1:size(ranks, :sites)],
+        [range(1, size(ranks, n), length=size(ranks, n)) for n in dn if name(n) != :scenarios],
+        [range(1, size(ranks, :sites), length=size(ranks, :sites))],
     )
+    axlist = Tuple(Dim{nm}(ran) for (nm, ran) in zip(dn_subset, freq_elements))
     mn = ([size(ranks, k) for k in freq_dims]..., size(ranks, :sites))
 
-    rank_frequencies = NamedDimsArray(zeros(mn...); zip(dn_subset, freq_elements)...)
+    rank_frequencies = YAXArray(axlist, zeros(mn...))
```
Zapiano, line 219: I believe you could use the new constructor `ZeroDataCube` for this (there are a few examples at `yax_main` branch)

test/seeding.jl
```diff
@@ -21,14 +21,22 @@ end
     # calculate available space
     available_space = vec((total_site_area .* k) .- current_cover)
 
+    axlst::Tuple = (
+        Dim{:taxa}(["N_seed_TA", "N_seed_CA", "N_seed_SM"]),
+    )
+    # Randomly generate seeded area
+    seeded_area = YAXArray(
```
Zapiano, line 28: You could use `DataCube(rand(...); axlist...)`

Zapiano (approved):
Only two minor comments, but other than that, lgtm

ConnectedSystems:
You can leave `rank_to_frequencies` and related to me

ext/AvizExt/viz/location_selection.jl
```diff
@@ -1,12 +1,12 @@
-using AxisKeys, NamedDims
+using AxisKeys, YAXArrays
```
ConnectedSystems, line 1: Remove AxisKeys too? I don't think this is used anywhere in this file now

ConnectedSystems (approved):

PR #701, opened 2024-02-19
added nameddimsarray2yaxarray

DanTanAtAims:
adding temporary utility function

### DIFF
--- a/src/io/inputs.jl
+++ b/src/io/inputs.jl
@@ -236,3 +236,8 @@ function yaxarray2nameddimsarray(cube::YAXArray)::NamedDimsArray
     axis_labels = collect.(lookup(cube, axis_names))
     return NamedDimsArray(cube.data; NamedTuple{axis_names}(axis_labels)...)
 end
+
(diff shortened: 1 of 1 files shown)


src/io/inputs.jl
```diff
@@ -236,3 +236,8 @@ function yaxarray2nameddimsarray(cube::YAXArray)::NamedDimsArray
     axis_labels = collect.(lookup(cube, axis_names))
     return NamedDimsArray(cube.data; NamedTuple{axis_names}(axis_labels)...)
 end
+
+function nameddimsarray2yaxarray(named_arr::NamedDimsArray)::YAXArray
+    new_axes = caxes(named_arr)
+    return YAXArray(new_axes, named_arr)
```
Zapiano, line 242: I did a quick test here and it seems that if you do that, when you `.data` on the resulting YAXArray, this data is a `NamedDimsArray`. Since this is just a temporary method, you could do `YAXArray(new_axes, collect(named_arr))` instead.

Zapiano (approved):
Thanks

PR #704, opened 2024-02-21
Yax cube tests

DanTanAtAims:
Tests for ADRIA.DataCube and ADRIA.ZeroDataCube.

### DIFF
--- a/test/io/inputs.jl
+++ b/test/io/inputs.jl
@@ -58,3 +58,61 @@ end
 
     @test match == true
 end
+
+@testset "ZeroDataCube" begin
+
+    ZeroDataCube = ADRIA.ZeroDataCube
+    
+    dim_1_name = :timesteps
+    dim_1_vals = rand(10)
+    dim_1 = Dim{dim_1_name}(dim_1_vals)
+    
+    dim_2_name = :location
+    dim_2_vals = ["loc 1", "loc 2", "loc 3", "loc 4", "loc 5"]
+    dim_2 = Dim{dim_2_name}(dim_2_vals)
+
+    yax_res = ZeroDataCube(Int, timesteps=dim_1_vals)
+
+    @test typeof(yax_res) <: YAXArray{Int, 1} || 
+        "Incorrect return type. Expected a subtype of YAXArray{Int, 1} \
+         but received $(typeof(yax_res))"
+
+    @test all(0 .== yax_res) || 
+        "Incorrect data contained in YAXArray. Expected all 0."
+
+    @test size(yax_res) == (10,) ||
+        "Incorrect YAXArray shape. Expected (10,) but received $(size(yax_res))"
+
+    @test name(dim_1) == name(yax_res.axes[1]) ||
+        "Incorrect dimension name. Expected $(name(dim_1)) \
+         but received $(name(yax_res.axes[1]))"
+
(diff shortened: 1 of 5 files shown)


test/io/inputs.jl
```diff
@@ -58,3 +58,94 @@ end
...
+    @test dim_1_vals == collect(yax_res.axes[1]) || 
+        "Incorrect axis indices. Expected $(dim_1_vals) but received $(collect(yax_res.axes[1]))"
+
+    @test dim_2_vals == collect(yax_res.axes[2]) || 
+        "Incorrect axis indices. Expected $(dim_2_vals) but received $(collect(yax_res.axes[2]))"
+end
+
+function test_DataCube(data; kwargs...)
+    # Check the true values are valid
+    @assert (size(data) == Tuple(length(dm_vls) for (_, dm_vls) in kwargs)) "Data shape should agree with dimensions shape"
```
Zapiano, line 121: I am not sure we need this `@assert`. If the dimensions don't match, this will raise and error so this way we are just adding a line to change where the error is being raised. 

Also, in Julia's doc there's a warning saying "Assert should therefore only be used as a debugging tool and not used for authentication verification (e.g., verifying passwords), nor should side effects needed for the function to work correctly be used inside of asserts." (https://docs.julialang.org/en/v1/base/base/#Base.@assert) which, to me, suggests that we shouldn't use it here (but please, let me know if you have a different interpretation, I might be misunderstanding something here).

test/io/inputs.jl
```diff
@@ -58,3 +58,94 @@ end
...
+    for (axs, (nme, val)) in zip(yax_res.axes, kwargs)
+        @test name(axs) == nme ||
+            "Incorrect axis name. Expected $(nme) but received $(name(axs))"
+        @test collect(axs) == val ||
+            "Incorrect axis values. Expected $(val) but received $(collect(axs))"
+    end
+end
+
+@testset "DataCube" begin
+    dim_1_name = :timesteps
```
Zapiano, line 137: Are the variables `dim_1_name`, `dim_2_name` and `dim_3_name` being used?

test/io/inputs.jl
```diff
@@ -58,3 +58,94 @@ end
...
+        "Incorrect return type. Expected a subtype of YAXArray{Int, 1} \
+         but received $(typeof(yax_res))"
+
+    @test all(0 .== yax_res) || 
+        "Incorrect data contained in YAXArray. Expected all 0."
+
+    @test size(yax_res) == (10,) ||
+        "Incorrect YAXArray shape. Expected (10,) but received $(size(yax_res))"
+
+    @test name(dim_1) == name(yax_res.axes[1]) ||
```
Zapiano, line 85: You don't need to build a `Dim` just for this.  `name(yax_res.axes[1]) == dim_1_name` does the same thing and is more economic. 

Also, you could have used this variable `dim_1_name` to build the `ZeroDataCube` - this will make it easier to update this test in the future, if needed:

```
dim_1_name = (:timesteps,)
dim_1_vals = ((rand(10)),)
ADRIA.ZeroDataCube(Int; NamedTuple{dim_1_name}(dim_1_vals)...)
```

Zapiano (commented):
Thanks for writing this tests. Just a few comments and we'll be good to go.

PR #695, opened 2024-02-15
Migrate from NamedDimsArrays to YAXArrays

Zapiano:
This involves a series of PRs. When everything is done, this branch should be merged with the `main`.

Tasks involved:

- [x] ADRIA Domain and ReefMod Domain loading #689 
- [x] Scenario Running #690 
- [x] Metrics module #709 
- [x] Analysis module #707 
- [x] Sensitivity module #710 
- [x] Viz module (may need multiple PRs) #703 
- [x] Write tests for DataCube and ZeroDataCube constructors #704 
- [x] Update tests (may need multiple PRs) #718

PR #709, opened 2024-02-26
Migrate metrics.jl to YAXArrays

Zapiano:
Migrate the entire metrics module. All the data inside result sets that previously were saved in NamedDimsArrays is now saved in YAXArrays.

Part of #695.

### DIFF
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -61,110 +61,6 @@ function (f::Metric)(rs::ResultSet, args...; kwargs...)::NamedDimsArray
     end
 end
 
-"""
-    to_string(m::Metric)::String
-
-Get name of metric as a string.
-"""
-function to_string(m::Metric)::String
-    return to_string(m.func)
-end
-function to_string(f::Function)::String
-    return join(split(String(Symbol(f))[2:end], "_"), " ")
-end
-
-"""
-    to_symbol(m::Metric)::String
-
-Get name of metric as a symbol.
-"""
-function to_symbol(m::Metric)::Symbol
-    return Symbol(replace(to_string(m), ' ' => '_'))
-end
-
-"""
-    metric_label(m::Metric)::String
-    metric_label(f::Function, unit::String)
-
-Return name of metric in the format: "Title Case [Unit]", suitable for use as a label.
-
-# Example
-```julia
-m_label = metric_label(scenario_total_cover)
-# "Scenario Total Cover [m²]"
-```
-"""
-function metric_label(m::Metric)::String
-    return metric_label(m.func, m.unit)
-end
-function metric_label(f::Function, unit::String)::String
-    n = titlecase(to_string(f))
-    if length(unit) > 0
-        n *= " [$unit]"
-    end
-
-    return n
-end
-
-"""
-    dims(m::Metric)::Tuple
-
-Get dimension names for a given outcome/metric.
-"""
-function dims(m::Metric)::Tuple
-    return m.dims
-end
-
-
-"""
-    ndims(m::Metric)::Int64
-
-Infer the number of dimensions for a given outcome/metric.
-"""
-function Base.ndims(m::Metric)::Int64
-    return length(dims(m))
-end
-
-
-"""
-    call_metric(metric, data, args...; timesteps=(:), species=(:), sites=(:), scens=(:))
-
-Convenience method that slices the data in the specified manner.
-
-# Arguments
-- `metric` : Function, the metric function to apply to "raw" data.
-- `data` : NamedDimsArray, data to pass into `metric`
-- `args` : Additional positional arguments to pass into `metric`
-- `dims` : dummy keyword argument, not used but defined to allow use with other methods
-"""
-function call_metric(metric::Union{Function,Metric}, data::NamedDimsArray, args...; kwargs...)
-    dims = haskey(kwargs, :dims) ? kwargs[:dims] : nothing
-    if isnothing(dims)
-        return metric(slice_results(data; kwargs...), args...)
-    else
-        return metric(slice_results(data; kwargs...), args...; dims=dims)
-    end
-end
-
-
-"""
-    slice_results(data::NamedDimsArray; timesteps=(:), species=(:), sites=(:), scenarios=(:))
-
-Slice data as indicated.
-Dimensions not found in target data are ignored.
-"""
-function slice_results(data::NamedDimsArray; timesteps=(:), species=(:), sites=(:), scenarios=(:))
-    f_dims = (timesteps=timesteps, species=species, sites=sites, scenarios=scenarios)
-
-    s_names = keys(f_dims)
-    d_names = NamedDims.dimnames(data)
-    common_dims = intersect(s_names, d_names)
-
-    selected_slice = (; zip(common_dims, [getfield(f_dims, k) for k in common_dims])...)
-    return data[selected_slice...]
-end
-
-
 function _relative_cover(X::AbstractArray{<:Real})::AbstractArray{<:Real}
     # Sum over all species and size classes
     return dropdims(sum(X, dims=2), dims=2)
@@ -695,12 +591,13 @@ Shelter volume relative to a theoretical maximum volume for the available \$k\$ 
 relative_shelter_volume = Metric(_relative_shelter_volume, (:timesteps, :sites, :scenarios))
 
 
-include("reef_indices.jl")
-include("temporal.jl")
-include("site_level.jl")
-include("scenario.jl")
-include("ranks.jl")
 include("pareto.jl")
+include("ranks.jl")
+include("reef_indices.jl")
+include("scenario.jl")
+include("site_level.jl")
+include("temporal.jl")
+include("utils.jl")
 
 
 if ccall(:jl_generating_output, Cint, ()) == 1
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -7,9 +7,8 @@ using
     OnlineStats,
     Statistics
 
-using
-    AxisKeys,
-    NamedDims
+using YAXArrays
+using ADRIA: DataCube, ZeroDataCube, axes_names, axis_labels, axis_index
 
 using FLoops
 using DataFrames
@@ -22,7 +21,7 @@ abstract type Outcome end
 
 struct Metric{F<:Function,T<:Tuple,S<:String} <: Outcome
     func::F
-    dims::T
+    dims::T     # output dimension axes ?
     unit::S
 end
 Metric(f, d) = Metric(f, d, "")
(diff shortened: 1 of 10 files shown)


Zapiano:
@DanTanAtAims  @ConnectedSystems  I think this one should be merged next to `yax_main` and all other PRs related to this migration to YAXArrays should then be rebased (and refactored if necessary) before merging them. Opinions?

Zapiano:
@Rosejoycrocker I know you probably have a lot on your plate but if you can just take a quick look, since you use these metrics a lot, it would be great. :)

DanTanAtAims:
> @DanTanAtAims @ConnectedSystems I think this one should be merged next to `yax_main` and all other PRs related to this migration to YAXArrays should then be rebased (and refactored if necessary) before merging them. Opinions?

Sounds good as most analysis/sensitivity/viz calls succeed metric calls.

Rosejoycrocker:
> @Rosejoycrocker I know you probably have a lot on your plate but if you can just take a quick look, since you use these metrics a lot, it would be great. :)

No problem :)

src/metrics/metrics.jl
```diff
@@ -279,44 +158,39 @@ end
 #     return rs.outcomes[:relative_loc_taxa_cover]
 # end
 
-relative_loc_taxa_cover = Metric(_relative_loc_taxa_cover, (:timesteps, :taxa, :location, :scenarios))
+relative_loc_taxa_cover = Metric(_relative_loc_taxa_cover, (:timesteps, :taxa, :sites, :scenarios))
```
Rosejoycrocker, line 161: I would have thought the dimension name `:location` or `:locations` was more general

Zapiano: Yes, we are going to migrate from `:sites` to `:locations`, but that will be on a separate PR and I thought it would be better if all dimensions follow the same standard. Do you think this change (this one line in this PR, I mean) can cause any problem? Probably the migration to `:locations` will happen next (after migrating to YAXArrays).

src/metrics/metrics.jl
```diff
@@ -279,44 +158,39 @@ end
...
+"""
+    relative_juveniles(X::AbstractArray{T}, coral_spec::DataFrame)::AbstractArray{T} where {T<:Real}
+    relative_juveniles(rs::ResultSet)::AbstractArray
 
+Juvenile coral cover relative to total site area.
+"""
 function _relative_juveniles(X::AbstractArray{T}, coral_spec::DataFrame)::AbstractArray{T} where {T<:Real}
     # Cover of juvenile corals (< 5cm diameter)
-    juv_groups::AbstractArray{<:Real} = X[species=coral_spec.class_id .== 1] .+ X[species=coral_spec.class_id .== 2]
+    juv_groups = X[species=(coral_spec.class_id .== 1)] .+ X[species=(coral_spec.class_id .== 2)]
```
DanTanAtAims, line 171: If `X` is a YAXArray then it is possible that indexing like this can error if the values `coral_spec.class_id .== 1` are all false, you would receive a very uninformative error message `ArgumentError: Chunk sizes must be strictly positive`.

My solution was to cast to a matrix/vector however this may be undersirable

src/metrics/metrics.jl
```diff
@@ -462,15 +337,19 @@ e.g., X[species=1:6] is Taxa 1, size classes 1-6; X[species=7:12] is Taxa 2, siz
...
+function _shelter_species_loop(
+    X::AbstractArray{T1,3},
+    n_species::Int64,
+    colony_vol_m3_per_m2::Array{F},
+    max_colony_vol_m3_per_m2::Array{F},
+    k_area::Array{F}
+)::YAXArray where {T1<:Real,F<:Float64}
     # Calculate absolute shelter volumes first
-    ASV = NamedDimsArray{(:timesteps, :species, :sites)}(zeros(size(X)...))
+    ASV::YAXArray = DataCube(zeros(size(X)...), (:timesteps, :species, :sites))
```
DanTanAtAims, line 352: Should we add an equivalent ZeroDataCube function with axes lists like DataCube?

src/metrics/metrics.jl
```diff
@@ -587,16 +430,151 @@ shelter volume (a 3D metric).
...
+    The metric applies log-log linear models developed by Urbina-Barreto et al., [1]
+    which uses colony diameter and planar area (2D metrics) to estimate
+    shelter volume (a 3D metric).
+
+    ```math
+    RSV = \\begin{cases}
+    TASV / MSV & TASV > 0, \\\\
+0 & \\text{otherwise}
+\\end{cases}
+```
```
DanTanAtAims, line 509:
```suggestion
    ```math
    RSV = \\begin{cases}
    TASV / MSV & TASV > 0, \\\\
    0 & \\text{otherwise}
    \\end{cases}
    ```
```
Possible indent?

DanTanAtAims (commented):
Aside from a comment about a possible crash during an edge case, everything looks good to me.

src/metrics/metrics.jl
Rosejoycrocker: Ah ok, sounds good, I don't think it effects anything else

src/metrics/metrics.jl
Zapiano: Two things:

1. This should never be an "all false" vector in this context because we are selecting one of the size classes (if I'm not mistaken).
2. In any case, what do you think of the following? I'm not sure we should use this HERE, because of 1., but in other places where an all false vector might happen this won't raise an error. 

```
species_idx::Vector{Int64} = collect(X.species)
juv_groups = X[species=At(species_idx[coral_spec.class_id .== 1])] .+ X[species=At(species_idx[coral_spec.class_id .== 2])]
```

src/metrics/metrics.jl
```diff
@@ -587,16 +430,151 @@ shelter volume (a 3D metric).
...
+Provide indication of shelter volume relative to theoretical maximum volume for
+    the area covered by coral.
+
+    The metric applies log-log linear models developed by Urbina-Barreto et al., [1]
+    which uses colony diameter and planar area (2D metrics) to estimate
+    shelter volume (a 3D metric).
+
+    ```math
+    RSV = \\begin{cases}
+    TASV / MSV & TASV > 0, \\\\
```
Rosejoycrocker, line 506: Should this be `if MSV>0` ?

src/metrics/metrics.jl
DanTanAtAims: My attempts at using the suggestion all resulted in an error. At this time the only error free solution I can think of is
```julia
bitVec = coral_spec.class_id .== 1
if sum(bitVec) == 0
    juv_groups = X[species=At([])] # results in 0 dimensional array
else 
    # normal behaviour
end
```
However, this doesn't apply to the above code as I assume both terms on either side of the addition would be of equal shape.

The reason I was casting to Matrix/Vector is because they support indexing with a bitVector of all 0's. Either way the occurences I found of it handle the zero case immediately after.

src/metrics/metrics.jl
Zapiano: I've just added a new ZeroDataCube constructor and updated the old one. You might need to update some code (sorry for that).

src/metrics/metrics.jl
Zapiano: I think you're right. 

@ConnectedSystems FYI I'm changing this.

src/metrics/metrics.jl
Zapiano: Hum, I don't understand why your attempts to use my suggestions resulted in error (what was the error?), I tested here and it worked. In any case, do you think we can leave this this way, since there shouldn't be an all false vector? If, by some mistake, there is an all false vector here, I think it **should** raise an error. Let me know what you think.

DanTanAtAims: I am happy for this to go through then. 

The error I received was the same as usual `ArgumentError: Chunk sizes must be strictly positive`. Reproduced with

```julia 
nmy = YAXArray((Dim{:dim1}(1:20), ), rand(20))
idx = collect(nmy.dim1)
nmy[dim1=At(idx[rand(20) .< 0])] # guarantee bit vector is all zeros
```

DanTanAtAims:
I'm happy with its current state.

src/metrics/metrics.jl
Zapiano: Hum, I see what the problem is. I tested `nmy[dim1=At([])]` and that works (for some reason). But `nmy[dim1=At(Int64[])]` doesn't.

Zapiano:
@DanTanAtAims @Rosejoycrocker @ConnectedSystems  This one is ready for another review. Now the entire metrics module has been migrated. Have fun! o/

src/metrics/metrics.jl
```diff
@@ -684,35 +520,116 @@ maximum shelter volume possible.
...
+include("temporal.jl")
+include("utils.jl")
+
+
+#if ccall(:jl_generating_output, Cint, ()) == 1
+#    Base.precompile(Tuple{Metric{typeof(_relative_shelter_volume),NTuple{4,Symbol},String},Array{Float64,3},Vector{Float64},Vararg{Any}})   # time: 1.6421927
+#    Base.precompile(Tuple{Metric{typeof(_relative_juveniles),Tuple{Symbol,Symbol,Symbol},String},Array{Float64,3}})   # time: 0.4322211
+#    Base.precompile(Tuple{Metric{typeof(_absolute_shelter_volume),Tuple{Symbol,Symbol,Symbol},String},Array{Float64,3},Vector{Float64},Vararg{Any}})   # time: 0.3056465
+#    Base.precompile(Tuple{Metric{typeof(_total_absolute_cover),Tuple{Symbol,Symbol,Symbol},String},Array{Float64,3},Vector{Float64}})   # time: 0.2397704
+#    Base.precompile(Tuple{Metric{typeof(_relative_taxa_cover),Tuple{Symbol,Symbol,Symbol},String},Array{Float64,3}})   # time: 0.1171238
```
DanTanAtAims, line 628: Should this be uncommented or removed?

src/metrics/scenario.jl
```diff
@@ -203,12 +191,12 @@ metrics::Vector{ADRIA.metrics.Metric} = [
...
 """
-function scenario_outcomes(rs::ResultSet, metrics::Vector{<:Metric})::NamedDimsArray
+function scenario_outcomes(rs::ResultSet, metrics::Vector{<:Metric})::YAXArray
     n_timesteps = length(timesteps(rs))
     n_scenarios = size(rs.inputs, 1)
     n_metrics = length(metrics)
 
-    scen_outcomes = NamedDimsArray(
+    scen_outcomes = DataCube(
         zeros(n_timesteps, n_scenarios, n_metrics);
```
DanTanAtAims, line 200:
```suggestion
    scen_outcomes = ZeroDataCube(;
        T=Float,
```
Is this a place to use the ZeroDataCube constructor?

DanTanAtAims (commented):
Aside from a few minor comments, looks good to me.

src/metrics/metrics.jl
ConnectedSystems: I think the reason for this was I had a lot of trouble getting the equation to be properly formatted when rendered in the documentation. The "if" clause would push everything out of alignment.

In the end I gave up. Maybe the issue was a quirk of the latex engine used to generate the doc page and is fixed now but I won't be surprised if this renders incorrectly.

src/metrics/metrics.jl
ConnectedSystems: You know what? Go ahead and delete. We can regenerate those statements once everything settles down anyway.

src/metrics/metrics.jl
Rosejoycrocker: If formatting is being annoying that's fine to leave the $\text{if}$, it was more that the condition should be $MSV>0$ not $TASV>0$

Rosejoycrocker (approved):
Loks good to me, although I think it needs a rebase before merging

Zapiano:
@ConnectedSystems  @Rosejoycrocker  @DanTanAtAims  I've fixed the commit mess within this branch. Apparently some things went wrong when rebasing `yax_main` yesterday that's why there are two extra commits fixing things.

src/metrics/metrics.jl
ConnectedSystems: Ah I see, yes.

DanTanAtAims (approved):
Looks good to me

ConnectedSystems (approved):
Haven't had the chance to fully test but changes look okay to me.

PR #702, opened 2024-02-19
Fix error in bin size for `sensitivity.outcome_map`

Rosejoycrocker:
Fixes bug where the number of bins was being set by the maximum number of categories where a categorical variable was used. Also moves similar code used in `sensitivity.rsa` into functions to be used by both methods.

Closes #615

PR #710, opened 2024-02-26
Yax sensitivity

DanTanAtAims:
This pull request would override changes on the (unmerged) branch [yax-sensitivity-analysis](https://github.com/open-AIMS/ADRIA.jl/tree/yax-sensitivity-analysis). The older branch seemed only to find and replace `NamedDimsArray` and was most likely unfinished.

This branch does not migrate `outcome_maps` as this function is migrated in #702. 

Part of #695

### DIFF
--- a/src/analysis/sensitivity.jl
+++ b/src/analysis/sensitivity.jl
@@ -3,8 +3,8 @@ module sensitivity
 using Logging
 
 using
-    AxisKeys,
     NamedDims,
+    AxisKeys,
     StaticArrays,
     YAXArrays
 using DataFrames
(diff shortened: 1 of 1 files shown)


src/analysis/sensitivity.jl
```diff
@@ -3,7 +3,6 @@ module sensitivity
 using Logging
 
 using
-    AxisKeys,
     NamedDims,
```
Zapiano, line 6: Do you still need to import `NamedDims`?

Zapiano (approved):
One small thing but, other than that, LGTM.

src/analysis/sensitivity.jl
DanTanAtAims: NamedDims is still required for the version of `outcome_map` contained in this pull request. `outcome_map` is migrates in a seperate pull request #702.

Yes it should eventually be removed however I didn't want to break precompilation.

PR #711, opened 2024-03-10
Remove GC workaround

ConnectedSystems:
Workaround added to resolve #416 is now no longer needed as the memory leak has been resolved upstream in Julia v1.10.
The issue was likely caused by something to do with the (auto-)garbage collection process.

See other details noted in #405 .

Might require bumping the minimum supported Julia version to v1.10

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -304,10 +304,6 @@ function run_scenario(
         end
     end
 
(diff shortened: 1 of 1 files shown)


Zapiano (approved):
🚀

PR #718, opened 2024-03-15
Update all tests and remove NamedDims and AxisKeys from ADRIA.jl

Zapiano:
Tests `site_selection.jl` and `spatial_clustering.jl` were commented and should be updated in a separate PR.

Part of #695 .


src/io/sampling.jl
```diff
@@ -404,66 +377,63 @@ function fix_factor!(d::Domain; factors...)::Nothing
     return nothing
 end
 
+const DISCRETE_FACTOR_TYPES = ["ordered categorical", "unordered categorical", "ordered discrete"]
+
+"""
+    is_discrete(dom, factor::Symbol)::Bool
```
ConnectedSystems, line 383:
```suggestion
    is_discrete(dom::Domain, factor::Symbol)::Bool
```

src/io/sampling.jl
```diff
@@ -404,66 +377,63 @@ function fix_factor!(d::Domain; factors...)::Nothing
...
+    is_discrete(p_type::String)::Bool
+
+Check ptype for discrete variable types. Returns true if discrete, false otherwise.
```
ConnectedSystems, line 386:
```suggestion
Check `ptype` attribute whether the factor is a discrete variable type or not. 
Returns `true` if discrete, `false` otherwise.
```

test/sampling.jl
```diff
@@ -1,4 +1,5 @@
 using ADRIA
+using Distributions
```
ConnectedSystems, line 2: I remembered that if you don't want to specify `Distributions` as a direct dependency, you could pull it in via ADRIA, like so:

```julia
using ADRIA.Distributions
```

src/io/sampling.jl
```diff
@@ -404,66 +377,64 @@ function fix_factor!(d::Domain; factors...)::Nothing
...
 """
-function get_default_dist_params(dom::Domain, factor::Symbol)::Tuple
-    model::Model = dom.model
-    factor_filter::BitVector = collect(model[:fieldname]) .== factor
-    default_params::Tuple = model[:default_dist_params][factor_filter][1]
-
-    return default_params
+function get_attr(dom::Domain, factor::Symbol, attr::Symbol)
+    ms = model_spec(dom)
+    ms[ms.fieldname.==factor, attr][1]
```
ConnectedSystems, line 437: Add explicit return as per style guide please

test/sampling.jl
```diff
@@ -9,35 +10,44 @@ end
...
+
+            if scens[i, :guided] > 0
+                cond = not_cw_lb .<= not_cw_scen_vals .<= not_cw_ub
+                @test all(cond) || "$msg | $(ms[.!(cond), :]) | $(not_cw_scen_vals[.!(cond)])"
+                continue
+            end
+            # When no interventions are used, e.g., for counterfactual or unguided scenarios
+            # (guided ∈ [-1, 0]) intervention parameters are set to 0 so only check ecological values
+            cond = lb[eco] .<= scen_vals[eco] .<= ub[eco]
+            @test all(cond) || "$coral_msg | $(ms[.!(cond), :]) | $(scen_vals[eco][.!(cond)])"
```
ConnectedSystems, line 46: Spacing


```suggestion
            end

            # When no interventions are used, e.g., for counterfactual or unguided scenarios
            # (guided ∈ [-1, 0]) intervention parameters are set to 0 so only check ecological values
            cond = lb[eco] .<= scen_vals[eco] .<= ub[eco]
            @test all(cond) || "$coral_msg | $(ms[.!(cond), :]) | $(scen_vals[eco][.!(cond)])"
```

test/sampling.jl
```diff
@@ -9,35 +10,44 @@ end
...
-        cond = min_x[eco] .<= x[eco] .<= max_x[eco]
-        @test all(cond) || "$coral_msg | $(ms[.!(cond), :]) | $(x[eco][.!(cond)])"
+        eco = (ms.component .== "Coral") .& .!(constant_params)
 
-        # Note: Test to ensure all intervention factors are set to 0 is covered by the guided
-        # sampling test below
+        msg = "Sampled values were not in expected bounds!"
+        coral_msg = "Sampled coral values were not in expected bounds!"
+        for i in 1:num_samples
+            # Filter off CriteriaWeights factors
```
ConnectedSystems, line 34:
```suggestion
            # Filter CriteriaWeights factors
```

src/io/sampling.jl
```diff
@@ -404,66 +377,64 @@ function fix_factor!(d::Domain; factors...)::Nothing
     return nothing
 end
 
+const DISCRETE_FACTOR_TYPES = ["ordered categorical", "unordered categorical", "ordered discrete"]
```
ConnectedSystems, line 380: I'd prefer constants be defined at the top of the file as per convention.

ConnectedSystems (changes requested):
Some minor comments but in all looks good.

Thanks for all the effort involved @DanTanAtAims @Zapiano @Rosejoycrocker

PR #722, opened 2024-03-20
Add support to cyclone mortality to ReefModDomain

Zapiano:
As in title.


ConnectedSystems (approved):
Thanks for doing this so quickly

Issue #452, opened 2023-09-05
Add option to viz.scenario to plot mean difference to cf ensemble for each scenario


Rosejoycrocker:
Currently viz.scenario differentiates counterfactual, guided and unguided scenarios through color coding. Adding the option to plot the  intervention scenarios as mean difference to the counterfactual ensemble would give a useful alternative visualisation for communicating expected improvements.

Rosejoycrocker:
The function `scenarios_series()` in `viz/scenarios' already plots the means and upper and lower quantiles (as bands around a line), given an array containing the mean and upper/lower quantiles for a series of x-values (e.g. time steps).

An additional function should be written which calculates the mean difference between the conterfactual and guided or unguided scenario set, as well as the lower and upper quantiles. These will be stored in an array with first dimension timesteps and second dimension of size 3.

ConnectedSystems:
Here's an example that will get you started.

I suggest adding a red line that spans across the time frame at 0 to indicate the counterfactual/benchmark value.

```julia
# load domain etc

# Only sweep through one DHW scenario
ADRIA.fix_factor!(dom, :dhw_scenario, 1)

n_scens = 64

# Run counterfactuals
cf_scens = ADRIA.sample_cf(dom, n_scens)
rs_cf = ADRIA.run_scenarios(dom, cf_scens, "45")
rs_stac = ADRIA.metrics.scenario_total_cover(rs_cf)

# Run with max deployments (unguided)
ug_scens = ADRIA.sample_unguided(dom, n_scens)
max_unguided = ADRIA.run_scenarios(dom, ug_scens, "45")
max_ung_stac = ADRIA.metrics.scenario_total_cover(max_unguided)

# Run with max deployments (guided)
g_scens = ADRIA.sample(dom, n_scens)
max_guided = ADRIA.run_scenarios(dom, g_scens, "45")
max_g_stac = ADRIA.metrics.scenario_total_cover(max_guided)

unguide_res = zeros(size(max_ung_stac, 1), 3)
for n in axes(max_ung_stac, 1)
    shuf_set = shuffle(1:n_scens)
    cf_shuf_set = shuffle(1:size(rs_stac, 2))
    unguide_res[n, :] .= collect(Iterators.flatten(confint(bootstrap(median, collect(max_ung_stac[n, shuf_set]) .- collect(rs_stac[n, cf_shuf_set]), BalancedSampling(100)), PercentileConfInt(0.95))))
end

[...]

PR #726, opened 2024-03-22
Add counterfactual minus guided/unguided diff at location level

Zapiano:
Followed an approach similar to https://github.com/open-AIMS/ADRIA.jl/issues/452#issuecomment-1942888428 but for a spatial level.

Note: Depends on #847 

## Example of code

```julia
using ADRIA
using WGLMakie, GeoMakie, GraphMakie

Makie.inline!(false)

# Load domain
dom = ADRIA.load_domain(...)

# Create scenarios
num_scens = 2^10
scens = ADRIA.sample(dom, num_scens)

# Run model / Load ResultSet

# Extract cf_difference_loc for guided and unguided (in that order)
rc = ADRIA.metrics.relative_cover(rs)

# Compute difference between guided and counterfactual using the 0.6-th quantile
gd_res = metrics.ensemble_loc_difference(rc, scens; agg_metric=0.6)

# Compute difference between unguided and counterfactual using the median
ug_res = metrics.ensemble_loc_difference(r_cover, scens; diff_target=:unguided)

# Plot maps of difference to the counterfactual
ADRIA.viz.map(rs, gd_res[summary=At(:agg_value)]; diverging=true)
ADRIA.viz.map(rs, ug_res[summary=At(:agg_value)]; diverging=true)
```

## Plots
- 1024 scenarios

| Guided - Counterfactual (0.6-th quantile) | Unguided - Counterfactual (median)|
| ----------------------- | ------------------------- |
| gd_cf_diff_map | ug_cf_diff_map |


src/metrics/spatial.jl
```diff
@@ -1,3 +1,6 @@
+using Bootstrap
+using Random
+
 """Functions and methods to produce location-level summaries."""
```
ConnectedSystems, line 4: High-level docstrings should be at the very top of file.

src/metrics/spatial.jl
```diff
@@ -106,3 +109,40 @@ function summarize(
...
+    n_scens = size(cf_outcomes, :scenarios)
+    for loc in 1:n_locs
+        cf_shuf_set1 = shuffle(1:n_scens)
+        ug_shuf_set = shuffle(1:n_scens)
+        ug_diff = collect(ug_outcomes[loc, ug_shuf_set] .- cf_outcomes[loc, cf_shuf_set1])
+        ug_result[loc] = bootstrap(median, ug_diff, BalancedSampling(100)).t0[1]
+
+        cf_shuf_set2 = shuffle(1:n_scens)
+        gd_shuf_set = shuffle(1:n_scens)
+        gd_diff = collect(gd_outcomes[loc, gd_shuf_set] .- cf_outcomes[loc, cf_shuf_set2])
```
ConnectedSystems, line 143: Could reuse the first `cf_shuf` here

src/metrics/spatial.jl
```diff
@@ -106,3 +109,40 @@ function summarize(
 )::YAXArray where {D,T,N,A}
     return summarize(data[timesteps=timesteps], alongs_axis, metric)
 end
+
+"""
+    cf_difference_map(rs, scens, metric)
+
+Give the mean bootstraped difference from the counterfactual for a given metric.
```
ConnectedSystems, line 116:
```suggestion
Give the mean bootstrapped difference from the counterfactual for a given metric.
```

ConnectedSystems: Document what is returned?

src/metrics/spatial.jl
```diff
@@ -106,3 +109,40 @@ function summarize(
...
+
+    cf_outcomes = outcomes_mean[scenarios=scens.guided .== -1]
+    ug_outcomes = outcomes_mean[scenarios=scens.guided .== 0]
+    gd_outcomes = outcomes_mean[scenarios=scens.guided .> 0]
+
+    n_locs = length(outcomes.sites)
+
+    ug_result = ZeroDataCube(; T=Float64, sites=locations.reef_siteid)
+    gd_result = ZeroDataCube(; T=Float64, sites=locations.reef_siteid)
+    n_scens = size(cf_outcomes, :scenarios)
```
ConnectedSystems, line 134: Not interested in the confidence bounds?

ConnectedSystems:
Bump @Zapiano

ConnectedSystems:
How close are we to merging this?

Zapiano:
> How close are we to merging this?

I'll start looking into this now and let you know what's missing or of it's ready soon.

Zapiano:
@ConnectedSystems This one is ready for another review.

ConnectedSystems:
An example usage would be appreciated too @Zapiano

Zapiano:
> Thanks.
> 
> An example usage would be appreciated too @Zapiano

I've added an example in this PR. Once this PR is approved I can add this plot to the documentation on a separate PR.

ConnectedSystems:
I'm not sure the indicated use it the best approach.

It forces only one set of bootstrapped results (only the median?), and forces additional computations to be done (e.g., what if I only want `gd_res`?)

```julia
# Extract cf_difference_loc for guided and unguided (in that order)
rc = ADRIA.metrics.relative_cover(rs)
gd_res, ug_res = ADRIA.metrics.cf_difference_loc(rc, scens)  # <- this method name feels awkward

# Plot cf_difference_loc for unguided and guided
gd_fig = ADRIA.viz.diff_map(rs, gd_res[2, :])  # <- why `[2, :]`? (I know why, but feels awkward)
ug_fig = ADRIA.viz.diff_map(rs, ug_res[2, :])
```

Instead, I suggest something like the below which, to me, is a cleaner UX.

```julia
rc = ADRIA.metrics.relative_cover(rs)

# outcome, scenario spec, percentile value of interest
gd_res = ADRIA.metrics.ensemble_difference(rc, scens, 0.5; timestep=2, scenario_type=:guided)

# outcome, scenario spec, statistical function of interest
ug_res = ADRIA.metrics.ensemble_difference(rc, scens, mean; timestep=2, scenario_type=:unguided)
```

On the plotting, it kind of feels like we're getting to the point where we want our own YAXArray subtype. 

```julia
ADRIA.viz.map(rs, gd_res::SomeCustomYAXArray?)  #<- uses different defaults to the "usual" `viz.map()`
```

Zapiano:
@ConnectedSystems this was updated to address your comments. The description was updated with new example of usage and plots.

Zapiano:
> On the plotting, it kind of feels like we're getting to the point where we want our own YAXArray subtype.
> 
> ```julia
> ADRIA.viz.map(rs, gd_res::SomeCustomYAXArray?)  #<- uses different defaults to the "usual" `viz.map()`
> ```

I'm not sure I understand the purpose of having our own YAXArray subtype.

ConnectedSystems:
> I'm not sure I understand the purpose of having our own YAXArray subtype.

To be able to define our own custom styling for specific use cases and be able to make different styles when plotting some arbitrary results (e.g., `some_plot(a_matrix)`) compared to `some_plot(a_general_yaxarray)` and `some_plot(a_yaxarray_with_context_specific_metadata)`

ConnectedSystems:
@Zapiano fyi there are conflicts now because of the merge of #863. Could you rebase or merge main as appropriate?

Zapiano:
> @Zapiano fyi there are conflicts now because of the merge of #863. Could you rebase or merge main as appropriate?

I rebased but the tests are not passing. I'll let you know when this is ready again.

Zapiano:
> > @Zapiano fyi there are conflicts now because of the merge of #863. Could you rebase or merge main as appropriate?
> 
> I rebased but the tests are not passing. I'll let you know when this is ready again.

@ConnectedSystems  the tests that were not passing is a problem on the main branch apparently (https://github.com/open-AIMS/ADRIA.jl/issues/890). If you prefer waiting for that to be fixed so we are sure the tests are passing for this PR as well it's fine, otherwise this one is ready to merge.

ConnectedSystems:
Have you updated to the latest dependencies?

Zapiano:
> Have you updated to the latest dependencies?

Yes.

ConnectedSystems:
> I'm not sure I understand the purpose of having our own YAXArray subtype.

FYI, Makie's theme recipes seem more stable now, and better documented:

https://docs.makie.org/stable/tutorials/wrap-existing-recipe

ConnectedSystems:
This PR can be merged once #890 is confirmed to be addressed by #891

Zapiano:
@ConnectedSystems tests are passing and I've updated testset names for the tests that are related to this PR as we discussed.

Zapiano:
@ConnectedSystems just reminding you that this is ready for review.

src/metrics/spatial.jl
```diff
@@ -122,3 +125,105 @@ function summarize(
...
+rs = ADRIA.run_scenarios(dom, scens, "45")
+
+# Calculate difference to the counterfactual for given metric
+_relative_cover = metrics.relative_cover(rs)
+
+# Compute difference between guided and counterfactual using the 0.6-th quantile
+gd_res = metrics.ensemble_loc_difference(r_cover, scens; agg_metric=0.6)
+
+# Compute difference between unguided and counterfactual using the median
+ug_res = metrics.cf_difference_loc(r_cover, scens; diff_target=:unguided)
```
ConnectedSystems, line 165: Incorrect example here?

src/metrics/spatial.jl
```diff
@@ -122,3 +125,105 @@ function summarize(
...
+
+# Plot maps of difference to the counterfactual
```
ConnectedSystems, line 167:
```suggestion
# Plot maps of difference to the counterfactual at timestep 2
```

Is the above correct?

src/metrics/spatial.jl
```diff
@@ -122,3 +125,105 @@ function summarize(
...
+# Returns
+Vector with bootstrapped difference (counterfactual - guided) for each location.
+"""
+function ensemble_loc_difference(
+    outcome::YAXArray{T,3},
+    scens::DataFrame;
+    agg_metric::Union{Function,AbstractFloat}=median,
+    diff_target=:guided,
+    conf::Float64=0.95,
+    rand_seed=1234
```
ConnectedSystems, line 181: Minor point - could we use a global random seed instead?

```suggestion
    rng::AbstractRNG=Random.GLOBAL_RNG
```

Later on...

`shuffle(rng, 1:n_cf_outcomes)[1:min_n_outcomes]`

src/metrics/spatial.jl
Zapiano: Yes, this `2` is not timestep, `gd_res[1,:]` and `gd_res[3,:]` are the lower and upper bounds of the confidence interval and `gd_res[2,:]` is the metric used (median or whatever).

src/metrics/spatial.jl
Zapiano: Fixed.

src/metrics/spatial.jl
Zapiano: Does that mean that the user shouldn't be allowed to set the seed and I should also remove this `Random.seed!(rand_seed)` ?

ConnectedSystems (changes requested):
Two minor comments, but this is ready to be merged once those are addressed.

src/metrics/spatial.jl
ConnectedSystems: Yes, you should remove the `Random.seed!` inside the function.

It means the user can set it outside the function.

This way the state of the random number generator does not "leak" outside the scope of the function.
Currently, the function "resets" the seed state, which may affect other functions.

Best practice is to set a RNG and pass that to each function if necessary

```julia
rng = Random.Xoshiro(123)  # use a different RNG algorithm
some_function(args...; rng=rng)
another_function(args...; rng=rng)
```

As an aside: the current exception in ADRIA is the scenario running, where I made the decision that we want the outcome to be tied to the scenario factors. But it's implemented in a suboptimal way... another thing to revisit.

Zapiano: Makes sense, thanks for the explanation.

Zapiano:
> Two minor comments, but this is ready to be merged once those are addressed.

Done, and all tests passing.

src/metrics/spatial.jl
ConnectedSystems: Hmm, is `gd_res` a YAXArray? It would be clearer what the "2" is for the example if the axis name is used as the selector instead

ConnectedSystems: I'd ask that the comment be updated to clarify what the "2" is.

Zapiano:
@ConnectedSystems I made two changes:

1. I updated the example (and tests) to use the YAXArray dim/label names instead of the index `2` directly when assessing the aggregate metric 
2. I have deleted the function `diff_map` and I'm using `map` instead, with an option for `diverging` color palette. That seems more flexible if we want to plot other maps that also need this palette in the future but are not "diff maps".

PR #732, opened 2024-04-07
Add support for specifying locations

ConnectedSystems:
Allow user to specify a set of locations (by their IDs) to restrict assessment to.

### DIFF
--- a/src/decision/location_selection.jl
+++ b/src/decision/location_selection.jl
@@ -40,8 +40,8 @@ fogging locations.
 - `rcp` : RCP conditions to assess
 - `min_iv_locs` : Minimum number of locations to intervene
 - `max_members` : Maximum number of locations per cluster
-- `target_seed_locs` : Locations to prioritize for seeding. Currently does nothing.
-- `target_fog_locs` : Locations to prioritize for fogging. Currently does nothing.
+- `target_seed_locs` : Specify locations by their IDs to consider for seeding.
+- `target_fog_locs` : Specify locations by their IDs to consider for fogging.
 
 # Returns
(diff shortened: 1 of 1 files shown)


src/decision/location_selection.jl
```diff
@@ -154,15 +156,15 @@ function rank_locations(
         # Create shared decision matrix
         # Ignore locations that cannot support corals or are out of depth bounds
         # from consideration
-        decision_mat = decision_matrix(
-            dom.site_ids[valid_locs],
+        seed_decision_mat = decision_matrix(
```
DanTanAtAims, line 159: If `valid_seed_locs` is all false this will crash. Though I assume that something is wrong with the user location inputs anyway.

src/decision/location_selection.jl
```diff
@@ -185,8 +187,18 @@ function rank_locations(
             ranks_store[locations=At(selected_seed_ranks), intervention=At(:seed), scenarios=scen_idx] .= 1:length(selected_seed_ranks)
         end
 
+        fog_decision_mat = decision_matrix(
```
DanTanAtAims, line 190: Same comment as above. This will crash if the location inputs don't contain viable seeding locations.

DanTanAtAims (commented):
Works as expected of me. 

I'm happy to leave the crash in for invalid location inputs as it is probably better than returning an empty array that someone might not notice.

ConnectedSystems:
> Works as expected of me.
> 
> I'm happy to leave the crash in for invalid location inputs as it is probably better than returning an empty array that someone might not notice.

It's a good point. If you hang on a tick I'll put in a check and error out with a meaningful message

ConnectedSystems:
Decided to emit a warning instead of erroring out instead.

PR #733, opened 2024-04-09
improve load_domain docs

arlowhite:
Clarify that path is to the directory, not the datapackage.json file.

Note: could make ADRIA support pointing to either, but just updating the docs for now.

### DIFF
--- a/docs/src/usage/domain.md
+++ b/docs/src/usage/domain.md
@@ -1,16 +1,21 @@
 # Loading a Domain
 
 ADRIA is designed to work with `Domain` data packages.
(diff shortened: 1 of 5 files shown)


ConnectedSystems:
Hi @arlowhite 

I understand this is a bit opaque in the documentation but note that the directory is the data package.

> A minimal data package on disk would be a directory containing a single file
>https://specs.frictionlessdata.io/data-package/#illustrative-structure

where the single file in the example is the `datapackage.json`.

I would prefer that the docs clarify the above rather than adding flexibility as to what is provided - though in principle I don't mind that either.

ConnectedSystems (approved):

PR #734, opened 2024-04-11
Only rank locations with available area > 0

ConnectedSystems:
Fix bug where locations == 0 were included when they should be excluded.


ConnectedSystems:
Bump @Zapiano

Zapiano (approved):

PR #736, opened 2024-04-14
Support more vector types for maps

ConnectedSystems:
Previous implementation only supported YAXArrays, so if you could not easily plot an arbitrary dataset

```julia
# This could not be done
some_values = [...]
ADRIA.viz.map(dom, some_values )
```

This PR expands the supported datatypes to include `AbstractVector{<:Real}`.


DanTanAtAims (approved):
Works for me.

PR #737, opened 2024-04-14
Clean up connectivity and centrality methods

ConnectedSystems:
This PR:

- Separates out determination of the strongest source locations
- Allows flexible use of alternate centrality measures in determining centrality
- Removes unused connectivity data from Domain types, making them more memory efficient

Previous implementation of the strongest source location violated the Single Responsibility Principle, as it was lumped in with all other connectivity-related calculations. In addition, it was unused but was being recalculated every decision point. As it was naively iterating through all nodes in the graph (rather than just target nodes of interest), it would take upwards of 10 seconds per decision point.

In the worst case for a large network (such as one representing the entire GBR), this would explode run times from ~2 hours to ~12 hours for a small set of 768 scenarios.

This set of changes makes `connectivity_strength()` essentially a wrapper around two lines of code. Really, it could be removed, but I am leaving it as is for now.

As an aside, I am unsure of the area-weighted implementation of connectivity is worthwhile given MCDA considers available space as a separate criteria already, but again, leaving it as is for now.

### DIFF
--- a/src/ExtInterface/ADRIA/Domain.jl
+++ b/src/ExtInterface/ADRIA/Domain.jl
@@ -18,9 +18,6 @@ mutable struct ADRIADomain <: Domain
     env_layer_md::EnvLayer  # Layers used
     scenario_invoke_time::String  # time latest set of scenarios were run
     const conn::YAXArray  # connectivity data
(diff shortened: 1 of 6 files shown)


DanTanAtAims (approved):
Works for me.

PR #738, opened 2024-04-14
Additional rank methods to determine number of locations intervened on per timestep

ConnectedSystems:
As in title.

Adds:

```julia
ADRIA.metrics.n_seed_locations(rs)
ADRIA.metrics.n_fog_locations(rs)
```


DanTanAtAims (approved):
Works for me.

PR #741, opened 2024-04-17
Fix bug, min. locations to select > available locations

ConnectedSystems:
Addresses error raised when specifying a minimum number of locations to select, when that minimum is greater than the number of locations that are viable/considered.


DanTanAtAims (approved):
Works for me.

PR #747, opened 2024-05-02
Working example of CUDA-fied `settler_cover()`

ConnectedSystems:
See changes.

I've changed the benchmark approach to simulate 75 time steps to better reflect use case (and align with the mock example shown previously).

Timings:

image

~10.9x faster compared to CPU but orders of magnitude slower than the mock example in #739.

Note the number of allocations explode when using the GPU approach. There's some memory allocations that could be avoided by either making a GPU-specific `recruitment_rate()` or pre-allocating arrays as necessary.

Note: In the code I comment that "`recruitment_rate()` takes < 1ms" but I'm not sure about that any more. Suspect there are additional memory allocations occurring.

Incidentally, in `recruitment_rate()` this line (`growth.jl:658`):

https://github.com/open-AIMS/ADRIA.jl/blob/aeeb749b9c1ea928504de9837af90e0331e59dde/src/ecosystem/corals/growth.jl#L658

I suspect the use of `replace()` is no longer necessary as we should always be passing in values for locations that are > 0.0

PS: Please follow style guide - space indentation instead of tab indentation for one.

### DIFF
--- a/test_performance/Manifest.toml
+++ b/test_performance/Manifest.toml
@@ -96,9 +96,9 @@ weakdeps = ["StaticArrays"]
 
 [[deps.AliasTables]]
 deps = ["Random"]
-git-tree-sha1 = "ca95b2220ef440817963baa71525a8f2f4ae7f8f"
+git-tree-sha1 = "07591db28451b3e45f4c0088a2d5e986ae5aa92d"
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
@arlowhite we have our dedicated remote desktop with a GPU. You just didn't have access to it because someone else was using it at the time.

PR #748, opened 2024-05-05
Repeat mention that the VS Code extension is being setup

ConnectedSystems:
Avoid confusion that may occur if the reader skips straight to the sub-section.


arlowhite:
Also, VSCode does automatically popup and suggest installing Julia Formatter. I'm assuming this is the same extension.

BG-AIMS (approved):

PR #750, opened 2024-05-05
Improved support for spatial maps

ConnectedSystems:
```julia
using ADRIA
using GeoMakie, GraphMakie, GeoMakie

moore_dom = ADRIA.load_domain("<path to cluster-scale domain>")
gbr_dom = ADRIA.load_domain(RMEDomain, "<path to RME dataset>", "45")

ADRIA.viz.map(moore_dom)  # See figure 1
ADRIA.viz.map(gbr_dom)  # See figure 2

```

Figure 1:
image

Figure 2:
image

I've yet to work out why the gridlines (with lat/longs) disappear for the smaller spatial scale, and why there is a relatively large space between the map and the colorbar.

Any thoughts @DanTanAtAims ?

FYI I've made the figures with the following Domain versions:

`Moore_2024-02-05b_v060_rc1` and `rme_ml_2024_01_08`

### DIFF
--- a/ext/AvizExt/AvizExt.jl
+++ b/ext/AvizExt/AvizExt.jl
@@ -38,11 +38,6 @@ include("./analysis.jl")
 include("./viz/viz.jl")
 
 
-function __init__()
-    ADRIA.viz.tmpdir = mktempdir()
-end
-
-
 """Main entry point for app."""
 function julia_main()::Cint
     if "explore" in ARGS
@@ -204,9 +199,6 @@ function ADRIA.viz.explore(rs::ResultSet)
     Label(traj_time_sld[1, 1], left_year_val)
     Label(traj_time_sld[1, 4], right_year_val)
 
-    # Get bounds to display
-    centroids = rs.site_centroids
-
     # Placeholder store to control which trajectories are visible
     X = rs.inputs
     min_color_step = (1.0 / 0.05)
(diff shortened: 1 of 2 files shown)


DanTanAtAims:
The grid lines for small domains are so coarse/spread out and the reef so small that the reef is displayed inbetween gridlines with the grid lines being off screen. If you zoom out enough for small plots the grid lines will appear.

DanTanAtAims:
The gap between the map and the colourbar is caused by the size of the map. Other people who have run into this issue seem to find it to be a result of the aspect ratio, but this doesn't seem to be our problem.
image

ConnectedSystems:
Noting some things for me to try later:

- [Aspect settings](https://docs.makie.org/stable/reference/blocks/axis/#aspect)
- Check what `datalims!()` does and see if it recalculates where grid lines appear
- Could also investigated x/y auto tick settings

DanTanAtAims (approved):
Works for me.

ConnectedSystems:
Okay, the "fix" is to define a figure size so that the auto-resizing has something to work off:

image

image

The downside is now there's extra space along the top and bottom, but it looks nicer this way.

Grid lines and missing x/y axis labels seems to be an issue with GeoMakie, so I'm leaving that as is for now.

ConnectedSystems:
@DanTanAtAims see very last commit, but otherwise I think good to go if you can't see any problems.

PR #751, opened 2024-05-06
use provided centrality measures instead of defaults

### LINKED ISSUE #749: Example given in #742 no longer works
The example provided in #742 (and attached in docstrings) no longer work:

Specifically this method appears to not be implemented?
`in_conn, out_conn, network = ADRIA.connectivity_strength(dom; out_method=eigenvector_centrality)`

DanTanAtAims:
I should have found this when reviewing pull request #737. The connectivity strength was always using defaults instead of provided centrality measures. 

It also appears that the default connectivity visualisation is quite poor for small reefs.

```julia
in_conn, out_conn, network = ADRIA.connectivity_strength(dom.conn, out_method=eigenvector_centrality)
ADRIA.viz.connectivity(dom, network, out_conn)
```

image

closes #749.

### DIFF
--- a/src/ecosystem/connectivity.jl
+++ b/src/ecosystem/connectivity.jl
@@ -182,8 +182,8 @@ function connectivity_strength(
     g = SimpleWeightedDiGraph(conn)
 
     # Measure centrality based on number of incoming connections
-    C1 = indegree_centrality(g)
-    C2 = outdegree_centrality(g)
+    C1 = in_method(g)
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
Oh I thought we called the transparency based on connectivity strengths? 

I should have caught this in my own review too

ConnectedSystems (approved):
Thank you

PR #753, opened 2024-05-08
Misc cleanup (coral dynamics)

ConnectedSystems:
Removing ODE package for slightly faster compilation times.

I'm also fixing factors for dev purposes and found that the keyword argument version of `fix_factor!()` is very slow.
The previous implementation naively looped over each user-specified factor. I've adjusted so fixes all specified factors to their given values. A tiny speed up in the end, but it avoids repeated casting of the ModelParameters.jl type to a DataFrame.

Also ran into a small bug where a worker would be attempted to be spun up, despite only having a single core configured.


DanTanAtAims (approved):
Thanks

PR #754, opened 2024-05-09
Changes to make use of expected multi-dimensional data structure

ConnectedSystems:
Now gets all the way to the collating the metrics before erroring out.

Many methods still expect a matrix, but have added helper methods to correctly transform/reshape the data.


DanTanAtAims (approved):
Thanks

PR #757, opened 2024-05-12
fix initial cover block constructor

DanTanAtAims:
pass available area (k_area m2) instead of total area at location.

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -605,7 +605,7 @@ function run_model(domain::Domain, param_set::YAXArray)::NamedTuple
 
     cover_blocks::Vector{Matrix{CoverBlock}} = [
         DynamicCoralCoverModel.blocks_model.CoverBlock.(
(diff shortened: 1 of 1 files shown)


src/scenario.jl
```diff
@@ -605,7 +605,7 @@ function run_model(domain::Domain, param_set::YAXArray)::NamedTuple
 
     cover_blocks::Vector{Matrix{CoverBlock}} = [
         DynamicCoralCoverModel.blocks_model.CoverBlock.(
-            reshape(C_cover[1, :, loc] .* site_data.area[loc], (n_sizes, n_groups))',
+            reshape(C_cover[1, :, loc] .* site_data.area[loc] .* site_data.k[loc], (n_sizes, n_groups))',
```
ConnectedSystems, line 608: Minor nitpick:

I would prefer relevant components to be clearly compartmentalized:

`C_cover[1, :, loc] .* (site_data.area[loc] .* site_data.k[loc])`

Makes explicitly clear what the intended relationship between the variables are.

ConnectedSystems (approved):

PR #761, opened 2024-05-22
Multidimensional cover

DanTanAtAims:
The commits reshape `C_cover` from `time x group_size x locs` to `time x group x size x locs`.

This involves providing duplicate implementations of processes that accept multidimensional inputs. The new implementations are for `proportional_adjustment`, `fecundity_scope`, `settler_DHW_tolerance`, `seed_corals`, `bleaching_mortality`, and `cyclone_mortality`. These changes are limited to functions called within `run_model`.

There is probably no rush to merge, to make the review easier compare the new function implementations with the non-multidimensional equivalents

PR #762, opened 2024-05-22
add range to for loop

DanTanAtAims:
This is the same fix added in #761.

Cyclone mortality was only being applied to the last function group of corals in the middle size classes.

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -830,7 +830,7 @@ function cyclone_mortality!(coral_cover, coral_params, cyclone_mortality)::Nothi
 
     # Mid class coral mortality
     coral_mid = hcat(collect(Iterators.partition(coral_params.mid, 4))...)
-    for i in size(coral_mid, 1)
+    for i in 1:size(coral_mid, 1)
         coral_deaths_mid = coral_cover[coral_mid[i, :], :] .* cyclone_mortality
         coral_cover[coral_mid[i, :], :] -= coral_deaths_mid
     end
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -830,9 +830,9 @@ function cyclone_mortality!(coral_cover, coral_params, cyclone_mortality)::Nothi
 
     # Mid class coral mortality
     coral_mid = hcat(collect(Iterators.partition(coral_params.mid, 4))...)
-    for i in 1:size(coral_mid, 1)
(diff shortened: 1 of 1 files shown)


src/scenario.jl
```diff
@@ -830,7 +830,7 @@ function cyclone_mortality!(coral_cover, coral_params, cyclone_mortality)::Nothi
 
     # Mid class coral mortality
     coral_mid = hcat(collect(Iterators.partition(coral_params.mid, 4))...)
-    for i in size(coral_mid, 1)
+    for i in 1:size(coral_mid, 1)
```
Zapiano, line 833: I'm ok with this fix, but maybe a better way would be to do

```
for i in eachrow(coral_mid)
    coral_deaths_mid = coral_cover[i, :] .* cyclone_mortality
    coral_cover[i, :] -= coral_deaths_mid
end
```

Zapiano (approved):
Thanks for catching that. I left a minor comment.

src/scenario.jl
ConnectedSystems: `eachrow` doesn't produce an index. For that you'd need to use `enumerate`

Zapiano: But in this case we don't need an index. The index is only used to access a row of the matrix. I'm proposing replacing:

```
for i in 1:size(coral_mid, 1)
    coral_deaths_mid = coral_cover[coral_mid[i, :], :] .* cyclone_mortality
     coral_cover[coral_mid[i, :], :] -= coral_deaths_mid
end
```` 

by:

```
for i in eachrow(coral_mid)
    coral_deaths_mid = coral_cover[i, :] .* cyclone_mortality
    coral_cover[i, :] -= coral_deaths_mid
end
```

DanTanAtAims: `coral_mid` is a matrix of indices so eachrow should work in this case

Zapiano (approved):
LGTM

src/scenario.jl
ConnectedSystems: The confusion there is you say you don't need the index, but then use an index?

> for i in eachrow(coral_mid)

> coral_cover[i, :]

ConnectedSystems: It makes sense now looking at Dan's changes, thanks.

PR #770, opened 2024-06-09
Change assisted adaptation to a Discrete Uniform value

ConnectedSystems:
Previously the assisted adaptation factor - indicating the level of enhancement/efficacy in terms of increased DHW tolerance - was defined as a Triangular Distribution.

I am changing this to a DiscreteOrderedUniform distribution to reflect that no one knows what the likely level of enhancement is.


DanTanAtAims (approved):
Works as expected.

PR #771, opened 2024-06-09
Use underlying data instead of YAXArray type

ConnectedSystems:
Addresses issue where data is misplaced in the figure

image

With fix:

image


DanTanAtAims (approved):
Works as expected.

PR #772, opened 2024-06-11
Increase compression level for ResultSet

ConnectedSystems:
Increasing compression level for slightly more manageable ResultSet sizes.

### DIFF
--- a/src/io/ResultSet.jl
+++ b/src/io/ResultSet.jl
@@ -212,6 +212,7 @@ function combine_results(result_sets...)::ResultSet
                 nrow(all_inputs),
                 size(rs1.seed_log, :timesteps),
                 size(rs1.seed_log, :sites),
+                36 # TODO: Find a way to remove this constant
             ))...
     )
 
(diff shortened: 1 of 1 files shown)


src/io/ResultSet.jl
```diff
@@ -212,6 +212,7 @@ function combine_results(result_sets...)::ResultSet
                 nrow(all_inputs),
                 size(rs1.seed_log, :timesteps),
                 size(rs1.seed_log, :sites),
+                36 # TODO: Find a way to remove this constant
```
ConnectedSystems, line 215: Oops, this may have been a defunct change that crept in.

Zapiano: I only see one method for this function `setup_logs` and it sccepts 5 arguments: `setup_logs(z_store, unique_sites, n_scens, tf, n_sites)`. I think this line should be removed, no?

ConnectedSystems: Yeah I'm trying to work out how it crept in...

ConnectedSystems:
This really was meant to be a one line change, not sure how the older line crept in!

All fixed now.

Zapiano (approved):
🚀

PR #774, opened 2024-06-11
Add methods to extract deployment location info

ConnectedSystems:
Helpers to extract details around number of deployment locations for scenarios that were run.

```julia
dim = ADRIA.load_domain(...)
freq_rank = ADRIA.decision.selection_ranks(rs.ranks, :seed; desc=true)

# Get details of locations ordered by their selection frequency.
rs.site_data[freq_rank, :]

deployment_summary = ADRIA.decision.n_deployment_locations(rs.ranks, :seed)
```


src/decision/location_selection.jl
```diff
@@ -441,7 +441,11 @@ function selection_frequency(
 end
 
 """
-    selection_ranks(ranks::YAXArray{T, 3}, iv_type::Union{Symbol,Int64}; desc::Bool=true)::Vector{Int64}
+    selection_ranks(
```
Zapiano, line 444: I thought the function signature in the docstring should be in one line.

src/decision/location_selection.jl
```diff
@@ -441,7 +441,11 @@ function selection_frequency(
...
+        ranks::YAXArray{T,4},
+        iv_type::Union{Symbol,Int64};
+        desc::Bool=true
+    )::Vector{Int64}
```
Zapiano, line 448:
```suggestion
    )::Vector{Int64} where {T<:Union{Int64, Float32, Float64}}
```

src/decision/location_selection.jl
```diff
@@ -439,3 +439,99 @@ function selection_frequency(
...
+rs = ADRIA.run_scenarios(dom, 128, "45")
+
+deployment_summary = ADRIA.decision.n_deployment_locations(rs.ranks, :seed)
+```
+
+# Arguments
+- `ranks` : Rankings from `ADRIA.decision.selection_ranks()`
+- `iv_type` : `:seed` or `:fog`
+"""
+function n_deployment_locations(
```
Zapiano, line 513: I found this name somewhat confusing. In your example you called this function and assigned it to a variable called `deplopyment_summary`. I think this would be a better name. Or `deployment_locs_summary` maybe.

ConnectedSystems: Good suggestion! 

I never liked the name but the method extracts the number of deployment locations, so that's what I went with.

I think `deployment_locs_summary` is ambiguous though: summary of what? Some property of the deployment locations?

I'm reading `deployment_summary` as a variable to be "summary of the number of deployment locations" (`deployment_summary = n_deployment_locations()`).

How about `count_deployment_locations()` ?

Though it's not a count either....

src/decision/location_selection.jl
ConnectedSystems: We don't follow that rule in other places 🤷 

Happy to replace with a single line though.

src/decision/location_selection.jl
Zapiano: Hum, but if I understand, this function doesn't return the number of deployment locations at all, right? It return this array called `summarized` with "Min, Mean, Median, Max, stdev" for each scenario. It's similar to the output of the DataFrame function `describe(df)`

image

We could call it `describe_deployment` or `describe_deployment_loc`...?

ConnectedSystems: You're right, it's an indication of the number of deployment locs.

I still don't like the blanket use of the term "describe" though (describe what about deployments?).

`scenario_deployment_stats()` ?

Thought doesn't convey what about deployments again...

Zapiano: I think when people say "summary stats" it's typically mean, median, and sometimes confidence interval, right? So `deployment_summary_stats` makes sense to me. The only thing is that the stats in question can either be for each scenario or for each location - and both can be useful...

ConnectedSystems: > So deployment_summary_stats makes sense to me. 

The issue for me is that number of deployments is not the only aspect of deployments that it could be referring to. It could be magnitude (number of coral for example), or spatial spread, or area deployed, or potentially the number of boats involved for seeding activities, etc. 

Here, it's attempting to provide summary stats for each scenario's deployment...

But maybe I'm overthinking it...

Zapiano: No, I see what you mean... it makes sense.

In this cases, I would expect to have the same function name, e.g. `summary_stats` or whatever, and the arguments passed to the function would tell me what is the function summarizing.

In our case, I think that would work because, despite you saying the it is summarizing "number of deployment locations", we can also say that it is summarizing the ranks, right? If we had other summary statistics we could use the same name but the arguments would be different.

----

I think that's already clear, but: my concert with `n_deployment_locations` is that I expect a function like this to return an `Int`, or maybe a `Array{Int}`, with the count of something - and that's not the case here.

ConnectedSystems: > I think that's already clear, but: my concert with n_deployment_locations is that I expect a function like this to return an Int, or maybe a Array{Int}, with the count of something - and that's not the case here.

Completely agree with you

ConnectedSystems: How about we go with `deployment_summary_stats()` for now.

ConnectedSystems: Changed to name for clarity.

Zapiano (approved):
🚀

PR #777, opened 2024-06-11
Fix visualization type error

DanTanAtAims:
example_err

The update to use `Dict{Symbol,<:Any}` in `ADRIA.viz.map` for optional arguments caused type mismatches in visualisations that were using mapping. Specifically, spatial visualization of clustering results.

This pull requests fixes this mismatch and allows tests to be run without erroring again.

### DIFF
--- a/ext/AvizExt/viz/clustering.jl
+++ b/ext/AvizExt/viz/clustering.jl
@@ -113,9 +113,9 @@ function ADRIA.viz.map(
     rs::Union{Domain,ResultSet},
     data::AbstractArray{<:Real},
     clusters::Union{BitVector,AbstractVector{Int64}};
-    opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(),
-    fig_opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(),
-    axis_opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(),
+    opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
+    fig_opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
+    axis_opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
 )::Figure
     f = Figure(; fig_opts...)
     g = f[1, 1] = GridLayout()
@@ -128,8 +128,8 @@ function ADRIA.viz.map!(
     rs::Union{Domain,ResultSet},
     data::AbstractVector{<:Real},
     clusters::Union{BitVector,Vector{Int64}};
-    opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(),
-    axis_opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(),
+    opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
+    axis_opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
 )::Union{GridLayout,GridPosition}
     # Although this function is called scenario_clusters, here we have locations clusters
     loc_groups::Dict{Symbol,BitVector} = ADRIA.analysis.scenario_clusters(clusters)
--- a/ext/AvizExt/viz/clustering.jl
+++ b/ext/AvizExt/viz/clustering.jl
@@ -92,8 +92,8 @@ function ADRIA.viz.clustered_scenarios!(
 end
 
 """
-    map(rs::Union{Domain,ResultSet}, data::AbstractMatrix, clusters::AbstractVector{Int64}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
-    map(g, rs, data, clusters; opts::Dict=Dict(), axis_opts::Dict=Dict())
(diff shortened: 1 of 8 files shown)


ConnectedSystems:
As you mentioned in a different conversation, we might as well migrate all use of `Dict()` to an internally defined `const` of `Dict{Symbol,<:Any}`.

We could also define an empty dict for all methods to use as a default value.

Something like:

```julia
const OPT_TYPE = Dict{Symbol,<:Any}
const DEFAULT_OPT_VAL = Dict{Symbol,Any}()

function some_viz_function(x, y; opt::OPT_TYPE=DEF_OPT_VAL)
end
```

This *should* work...

Zapiano:
Is there a reason why we don't use NamedTuples instead of Dicts for this?

ConnectedSystems:
> Is there a reason why we don't use NamedTuples instead of Dicts for this?

```julia
julia> x = (pedro=100, award=true)
(pedro = 100, award = true)

julia> x.another_option = 200
ERROR: type NamedTuple has no field another_option
Stacktrace:
 [1] setproperty!(x::@NamedTuple{pedro::Int64, award::Bool}, f::Symbol, v::Int64)
   @ Base .\Base.jl:39
 [2] top-level scope
   @ REPL[10]:1
```

Hypothetically you could use NamedTuples by recreating a NamedTuple everytime a method needs to add/update an option, but it will get messy...

Using Dicts was the most straightforward approach I could think of. Open to other options.

DanTanAtAims:
Something I found unexpected with `DEFAULT_OPT_VAL` is although its declared constant, functions that edit the `opt` dictionary creates side effects for others. For example

```julia
ADRIA.viz.clustered_scenarios(args...; opts=DEFAULT_OPT_VAL)
```
contains the following line
```julia
opts[:histogram] = false
```

Then if you call any other function with default axis options,
```julia
ADRIA.viz.scenarios(rs, s_tac) # axis_opts=DEFAULT_OPT_VAL
```
 it will error as `:histogram` is not a valid kwarg for the `Axis` constructor. Somehow we obtain a non-constant reference to a constant global which seems like a bug in Julia.

An alternative is to instead define the default type
```julia
const OPT_TYPE = Dict{Symbol,<:Any}
const DEFAULT_OPT_TYPE= Dict{Symbol,Any}
```
and use
```julia
function viz_func(;
    opts::OPT_TYPE=DEFAULT_OPT_TYPE()
)
```

ConnectedSystems:
Ah, the constant value itself is mutable. Same with Vectors, you can declare a Vector and add to it, but you can never replace it.

I think your alternative approach is what I actually meant to do...

DanTanAtAims:
Ready for review.

ConnectedSystems (approved):
Awesome, looks all good to me!

ext/AvizExt/viz/clustering.jl
```diff
@@ -1,8 +1,8 @@
 using Statistics
 
 """
-    scenarios(outcomes::AbstractMatrix, clusters::Vector{Int64}; opts::Dict=Dict(), fig_opts::Dict=Dict(), axis_opts::Dict=Dict())
-    scenarios!(g::Union{GridLayout,GridPosition}, outcomes::AbstractMatrix, clusters::Vector{Int64}; opts::Dict=Dict(), axis_opts::Dict=Dict())
+    scenarios(outcomes::AbstractMatrix, clusters::Vector{Int64}; opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(), fig_opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}(), axis_opts::Dict{Symbol,<:Any}=Dict{Symbol,Any}())
```
Zapiano, line 4: Just to clarify, the idea is that the Docstring will not be identical to the function declaration in these cases then?

ConnectedSystems: As you might have guessed, I decided to just leave it

DanTanAtAims: I think it is better to leave it to what it resolves to. In the case of a new user, they will most likely miss the global type declaration and lead to more confusion. 

The global declaration allows us to easily maintain compatibility between viz functions, preventing type errors from ambiguously constructed dictionaries.

ext/AvizExt/viz/scenarios.jl
```diff
@@ -122,10 +122,13 @@ function ADRIA.viz.scenarios!(
...
     outcomes::YAXArray;
-    opts::Dict=Dict(),
-    axis_opts::Dict=Dict(),
-    series_opts::Dict=Dict(),
+    opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
+    axis_opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
+    series_opts::OPT_TYPE=DEFAULT_OPT_TYPE()
 )::Union{GridLayout,GridPosition}
+    if isdefined(Main, :Infiltrator)
+      Main.infiltrate(@__MODULE__, Base.@locals, @__FILE__, @__LINE__)
```
Zapiano, line 130: @DanTanAtAims was that intentional?

DanTanAtAims: Sorry I'll remove.

ConnectedSystems: Bah, sorry I missed this

PR #779, opened 2024-06-12
constrain growth by available area

DanTanAtAims:
Add growth constraints based on available area. 

The initial implementation in the main branch uses `actual_growth = growth * log(1 + leftover space)`. 

This implementation uses `actual_growth = growth * log(2, 1 + leftover space)`. Using log base 2 so that when leftover space is 100% the log is 1 and growth is unconstrained.

This assumes the growth rates being used are unconstrained growth rates which probably isn't the case and depends on available space when the measurement took place. 

### Unconstrained RCP4.5 (Before changes)

unconstrained_45

### Constrained RCP4.5 (After Changes)

constrained_45

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -671,8 +671,7 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
         )
 
         # Constrain growth by available area
-        # Constraint Coefficient -> log(1 + left over space)
-        growth_spatial_constraint .= log.(2, 1 .+ relative_leftover_space(
+        growth_spatial_constraint .= log2.(1 .+ relative_leftover_space(
             dropdims(sum(C_t; dims=(1, 2)), dims=(1, 2))
         ))
 
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -658,6 +658,8 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
     C_t::Array{Float64, 3} = zeros(n_groups, n_sizes, n_locs)
     cover_copy = zeros(n_groups, n_sizes, n_locs)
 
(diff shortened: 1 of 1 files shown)


Zapiano:
I like the change to log base 2, good thinking!

src/scenario.jl
```diff
@@ -668,12 +670,18 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
             site_data.area .* site_data.k, (1, 1, n_locs)
         )
 
+        # Constrain growth by available area
+        # Constraint Coefficient -> log(1 + left over space)
```
ConnectedSystems, line 674: Is this an unnecessary (and outdated) comment?

DanTanAtAims: I was concerned the code didn't make the formula obvious, but it can be removed or updated to reflect the base of the log.

ConnectedSystems: Why not use `log2()` ? Wouldn't that make the code clearer?

src/scenario.jl
```diff
@@ -668,12 +670,18 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
...
+        growth_spatial_constraint .= log.(2, 1 .+ relative_leftover_space(
+            dropdims(sum(C_t; dims=(1, 2)), dims=(1, 2))
+        ))
```
ConnectedSystems, line 677: Hmm... not sure how much this will change things but technically this could be moved into the loop over locations below.

Keep the `growth_spatial_constraint` but make it so each thread is calculating/updating/using the cache separately.

This way we're only looping over $n$ (where $n$ is the number of reefs) once instead of twice.

ConnectedSystems (approved):
I'm approving, but leaving the final decision to @Zapiano in case he catches anything else.

Zapiano (approved):
LGTM

PR #786, opened 2024-06-19
Assortment of improvements to SIRUS and Time Series Clustering 

Zapiano:
- Add option to pass `lambda` param when running SIRUS
- Add option to choose distinct clustering algorithm with `cluster_series` and `cluster_scenarios`
- Add option to use distinct distance measurement with `cluster_series` and `cluster_scenarios`
- Standardize `target` and `non-target` plot colors (use the same colors as in SIRUS scatter plot)`

### DIFF
--- a/src/analysis/rule_extraction.jl
+++ b/src/analysis/rule_extraction.jl
@@ -98,15 +98,17 @@ end
 
 
 """
-    cluster_rules(clusters::Vector{T}, X::DataFrame, outcomes::AbstractMatrix{F}, max_rules::Int64; n_trees::Int64=1000) where {T<:Integer,F<:Real}
+    cluster_rules(clusters::Vector{T}, X::DataFrame, max_rules::T; seed::Int64=123, kwargs...) where {T<:Integer,F<:Real}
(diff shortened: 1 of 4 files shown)


src/analysis/rule_extraction.jl
```diff
@@ -123,13 +123,13 @@ A StableRules object (implemented by SIRUS).
    https://doi.org//10.1214/20-EJS1792
 """
 function cluster_rules(clusters::Vector{T}, X::DataFrame, max_rules::T;
-    n_trees::T=1000, seed=123) where {T<:Int64}
+    n_trees::T=1000, seed=123, lambda=1e-3) where {T<:Int64}
```
ConnectedSystems, line 126: Would it be more convenient to specify a generic `kwargs...` argument, given `lambda` is simply passed to SIRUS?

The upside is it will support any other args you might need that SIRUS has.

Zapiano: Yes, that makes sense.

ConnectedSystems (approved):
Approved, with change of specific keyword args to general `kwargs`

src/analysis/rule_extraction.jl
Zapiano: Done

PR #787, opened 2024-06-19
Add default option to _sort_keys

### LINKED ISSUE #769: Order of scenarios plotted are inconsistent
Scenario visualizations order scenario groupings based on variance.

This causes an artifact where plots do not maintain consistent ordering. It should always be "counterfactual", "unguided", "guided".

https://github.com/open-AIMS/ADRIA.jl/blob/437d51a030bd58f0d794dd46d03056a70f8fa09c/ext/AvizExt/viz/scenarios.jl#L221

Zapiano:
As in title. Default order is "counterfactual", "unguided", "guided".

> NOTE: When plotting clusters, the legend is always in the same order, but the plot order might change depending on `sort_by`

Below are some plots of scenarios and clusters, sorted by default, variance and size:

| `sort_by` | Scenarios | Cluster |
| -------- | --------- | ------- |
| `:default` | image | image |
| `:default` | image | image |
| `:variance` | image | image |
| `:variance` | image | image |
| `:size` | image | image |
| `:size` | image | image |

Grouped by RCP:

| `sort_by` | summarize = true | summarize = false |
| --- | --- | --- |
| `:default` | image | image |
| `:variance` | image | image |
  
 Closes #769

### DIFF
--- a/ext/AvizExt/viz/clustering.jl
+++ b/ext/AvizExt/viz/clustering.jl
@@ -44,6 +44,8 @@ function ADRIA.viz.scenarios!(
 
     scen_groups = ADRIA.analysis.scenario_clusters(clusters)
     opts[:histogram] = false
+    opts[:legend_labels] = sort(collect(keys(scen_groups)))
+
     return ADRIA.viz.scenarios!(
         g,
         ax,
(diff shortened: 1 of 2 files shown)


ConnectedSystems (approved):
Thanks

PR #760, opened 2024-05-21
Updates to make ADRIA compatible with new coral cover model

Zapiano:

PR #790, opened 2024-06-25
Julia format GH workflow, format all files

arlowhite:
Adds [julia-format](https://github.com/julia-actions/julia-format) GitHub workflow action.
Add `.JuliaFormatter.toml`
Update `development_setup.md`
format all files.

Also see [Notion notes](https://www.notion.so/Julia-Linter-Formatter-e12b052b5bef4796a888badeafc24050?pvs=4)

## Developer Setup Change

Recommendation is to uninstall the Julia Formatter plugin (or disable for your ADRIA.jl workspace if you use it in other projects). The primary VSCode Julia plugin uses `.JuliaFormatter.toml`, making the plugin redundant; the plugin only uses its own config and ignores `.JuliaFormatter.toml`. So having both installed/active adds confusion with around which formatting config is being applied.

## Formatter Config Changes

Changes from what we were using before with the formatter plugin. These settings reduce formatter changes:
* short_to_long_function_def=false
* conditional_to_if=false
* always_for_in="nothing"

Note: `always_for_in` config is strange. `false` seems to be always `=`, `true` is `in` (or value of `for_in_replacement`). `nothing` makes no change.


docs/make.jl
```diff
@@ -3,10 +3,9 @@ push!(LOAD_PATH, "../src/")
 
 using Documenter, ADRIA
 
-
-makedocs(
+makedocs(;
```
arlowhite, line 6: This is weird, not sure why the semicolon on opening parenthesis.

ext/AvizExt/AvizExt.jl
```diff
@@ -69,19 +66,19 @@ function main_menu()
 
     logo = image(
         f[1, 1],
-        rotr90(load(convert(String, LOGO))),
+        rotr90(load(convert(String, LOGO)));
```
arlowhite, line 69: comma converted to semicolon?

arlowhite (commented):
seems like there's a rule that a semicolon always goes before named args. in some cases this seems awkward to me.

ext/AvizExt/AvizExt.jl
ConnectedSystems: Is there a named argument afterwards? If so, that's why

docs/make.jl
ConnectedSystems: If the only arguments are keyword arguments, this is clear to me:

```julia
function blah(;
    foo=100
)
```

And is preferred over some other weirdness

```julia
function blah(
    ; foo=100
)
```

ext/AvizExt/AvizExt.jl
arlowhite: yes, there is, it just looks funny to me when it does `foo(; x = 1)` but I'm not seeing any option to omit the leading semicolon if all args are kwargs.

ConnectedSystems:
> seems like there's a rule that a semicolon always goes before named args. in some cases this seems awkward to me.

Yes, this is Julia syntax. Keyword arguments are specified after a semi-colon.

ext/AvizExt/AvizExt.jl
ConnectedSystems: It'll be a syntax error without the semi-colon, won't it? At least I thought it would be...

ODE_preferences.jl
```diff
@@ -3,9 +3,22 @@ using OrdinaryDiffEq
 
 set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileNonStiff" => true)
 set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileStiff" => false)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileAutoSwitch" => false)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileLowStorage" => false)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileDefaultSpecialize" => true)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileAutoSpecialize" => false)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileFunctionWrapperSpecialize" => false)
-set_preferences!(UUID("1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"), "PrecompileNoSpecialize" => false)
+set_preferences!(
```
arlowhite, line 6: For example, this could be a place where we disable formatting so they're always one line. 
```
#! format: off
code
#! format: on
```

arlowhite:
I was searching to see if there's a way to preserve git blame history. The best way seems to be `--ignore-revs-file`
https://medium.com/codex/how-to-introduce-a-code-formatter-without-messing-up-git-history-4a16bd074c10
So IDE git blame features would ignore this commit.

ext/AvizExt/AvizExt.jl
```diff
@@ -511,21 +548,25 @@ function ADRIA.viz.explore(rs::ResultSet)
...
+            2
+        )
     end
 
     @info "Displaying UI"
     gl_screen = display(f)
     # DataInspector()
 
-    wait(gl_screen)
+    return wait(gl_screen)
```
arlowhite, line 563: would you prefer this instead?
```julia
wait(gl_screen)
return
```

ODE_preferences.jl
arlowhite: though in this particular case, better to extract the variable.

ConnectedSystems: Don't stress too much about this particular file, it'll be removed after the new model gets integrated (@Zapiano) as we won't need the package it is configuring

ext/AvizExt/AvizExt.jl
ConnectedSystems: Explicit return is better imo.

And yes, what you suggest is fine with me

Zapiano:
Since we have a very critical/delicate PR, almost ready to merge, changing more than 20 files, I think say we should wait for that one to be merged before this.

ConnectedSystems:
@arlowhite @Zapiano is the blocker for this PR now resolved?

Zapiano:
> @arlowhite @Zapiano is the blocker for this PR now resolved?

My understanding we decided to wait until #760 is merged to main.

arlowhite:
In our last meeting, we discussed getting this in sooner than later and allowing PRs to merge with this check failing if necessary. I've rebased and reformatted.

arlowhite:
@Zapiano do you still want me to leave `src/scenario.jl` untouched?
or are you ok with these formatting changes?
https://github.com/open-AIMS/ADRIA.jl/pull/790/files#diff-bd5b8c618631550cda060d53f80beee397a2975f98de66d5a123ed45199ed86c

Zapiano:
> @Zapiano do you still want me to leave `src/scenario.jl` untouched? or are you ok with these formatting changes? https://github.com/open-AIMS/ADRIA.jl/pull/790/files#diff-bd5b8c618631550cda060d53f80beee397a2975f98de66d5a123ed45199ed86c

I am going to merge #817 today, after that you can format the entire project and I review it. Is that ok?

ext/AvizExt/viz/taxa_dynamics.jl
```diff
@@ -187,12 +200,16 @@ function taxonomy_by_intervention!(
...
     # Plot and calculate confidence intervals
     confints = zeros(n_timesteps, n_functional_groups, 3)
-    for (idx, taxa) in enumerate(relative_taxa_cover.taxa)
+    for (idx, taxa) ∈ enumerate(relative_taxa_cover.taxa)
         confints[:, idx, :] = series_confint(relative_taxa_cover[taxa=At(taxa)])
-        show_confints ? band!(
-            ax, 1:n_timesteps, confints[:, idx, 1], confints[:, idx, 3];
-            color=(colors[idx], 0.4)
-        ) : nothing
+        if show_confints
```
arlowhite, line 205: not sure I like it changing if/else to elvis operator. what are your thoughts?

ConnectedSystems: I think It should leave these alone. If I use the ternary operator, it's for a reason, and vice versa.

arlowhite:
What's your preference here?
returning assignment expressions seems a bit weird to me:
`return data .= (data .- mi) ./ (ma - mi)`
so, I'm thinking this is better:
```
data .= (data .- mi) ./ (ma - mi)
return data
```

ConnectedSystems:
Agree with your thinking

src/analysis/rule_extraction.jl
```diff
@@ -88,15 +92,18 @@ compact form.
 - `rules` : Vector of Rule objects
 """
 function print_rules(rules::Vector{Rule{Vector{Vector},Vector{Float64}}})::Nothing
-    condition(rule) = [c[2] == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
+    function condition(rule)
```
arlowhite, line 95: `function` expansion. if you dislike this, I can look into options to prevent it.

ConnectedSystems: It's definitely more verbose than I would prefer but I guess it's okay...

In fact, probably a sign that the function should be extracted, as currently it gets recompiled every call to `print_rules()`

ConnectedSystems:
Not a fan of these forced replacement of `in` to `∈` but I guess it's okay...

```julia
for r ∈ 1:n_rows
        for c ∈ 1:n_cols
```

src/analysis/rule_extraction.jl
ConnectedSystems: (that wasn't to say we should refactor it in this PR)

src/decision/location_selection.jl
```diff
@@ -281,26 +299,26 @@ Selection score
 """
 function selection_score(
     ranks::YAXArray{T,3},
-    iv_type::Union{Symbol,Int64},
-)::YAXArray where {T<:Union{Int64, Float32, Float64}}
+    iv_type::Union{Symbol,Int64}
+)::YAXArray where {T<:Union{Int64,Float32,Float64}}
```
arlowhite, line 303: funny that it wants comma-space elsewhere, but not with types.

ext/AvizExt/AvizExt.jl
arlowhite: I've added `return nothing` in appropriate places. I think I've caught them all, but it won't hurt anything if I missed one since the formatter is preserving current behavior.

Zapiano:
@arlowhite Now you can rebase and format all files. Thanks for waiting (and sorry it took so long).

arlowhite:
I've disabled the `for in` change and `function` expansion. The documented defaults are false, but I think blue style turns them on.
* https://domluna.github.io/JuliaFormatter.jl/stable/#conditional_to_if
* https://domluna.github.io/JuliaFormatter.jl/stable/#short_to_long_function_def
* https://domluna.github.io/JuliaFormatter.jl/stable/#always_for_in

arlowhite:
> Not a fan of these forced replacement of `in` to `∈` but I guess it's okay...
> 
> ```julia
> for r ∈ 1:n_rows
>         for c ∈ 1:n_cols
> ```

I've disabled it.

arlowhite:
Only one place in code doesn't use for `in`
src\io\rme_result_io.jl
`location_centroids = [centroid(multipoly) for multipoly ∈ geodata.geom]`

So we might want to do
```
always_for_in=true
for_in_replacement="in"
```

ext/AvizExt/viz/taxa_dynamics.jl
```diff
@@ -189,7 +199,8 @@ function taxonomy_by_intervention!(
     confints = zeros(n_timesteps, n_functional_groups, 3)
     for (idx, taxa) in enumerate(relative_taxa_cover.taxa)
         confints[:, idx, :] = series_confint(relative_taxa_cover[taxa=At(taxa)])
-        show_confints ? band!(
+        show_confints ?
```
Zapiano, line 202: Same here, probably if/else would be better.

src/ExtInterface/ReefMod/RMEDomain.jl
```diff
@@ -185,7 +189,8 @@ function load_domain(::Type{RMEDomain}, fn_path::String, RCP::String)::RMEDomain
 
     functional_groups = functional_group_names()
     cyc_scens::YAXArray{Float64} = ZeroDataCube(;
-        T=Float64, timesteps=timeframe_range, locs=loc_ids, species=functional_groups, scenarios=[1]
+        T=Float64, timesteps=timeframe_range, locs=loc_ids, species=functional_groups,
```
Zapiano, line 192:
```suggestion
        T=Float64, 
        timesteps=timeframe_range, 
        locs=loc_ids, 
        species=functional_groups,
```

src/ExtInterface/ReefMod/ReefModDomain.jl
```diff
@@ -214,13 +213,14 @@ function load_initial_cover(
...
-    size_class_weights = (cdf_integral[2:end] .- cdf_integral[1:(end-1)])
+    size_class_weights = (cdf_integral[2:end] .- cdf_integral[1:(end - 1)])
     size_class_weights = size_class_weights ./ sum(size_class_weights)
 
     # Take the mean over repeats, as suggested by YM (pers comm. 2023-02-27 12:40pm AEDT).
     # Convert from percent to relative values.
     # YAXArray ordering is [time ⋅ location ⋅ scenario]
-    icc_data = ((dropdims(mean(init_cc_per_taxa; dims=:scenario); dims=:scenario)) ./ 100.0).data
+    icc_data =
+        ((dropdims(mean(init_cc_per_taxa; dims=:scenario); dims=:scenario)) ./ 100.0).data
```
Zapiano, line 223: I'm not against it but this a little weird... maybe 

```julia
mean_init_icc = dropdims(mean(init_cc_per_taxa; dims=:scenario); dims=:scenario)
icc_data = (mean_init_icc ./ 100).data
```

src/analysis/rule_extraction.jl
```diff
@@ -88,15 +92,15 @@ compact form.
 - `rules` : Vector of Rule objects
 """
 function print_rules(rules::Vector{Rule{Vector{Vector},Vector{Float64}}})::Nothing
-    condition(rule) = [c[2] == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
+    condition(rule) =
+        [c[2] == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition]
```
Zapiano, line 96: What about

```
    condition(rule) = [
        c[2] == :L ? "$(c[1]) < $(c[3])" : "$(c[1]) ≥ $(c[3])" for c in rule.condition
    ]
```

?

src/decision/Criteria/SeedCriteria.jl
```diff
@@ -358,5 +361,7 @@ function _update_state(cluster_ids::Vector, num_locs::Int64, max_members::Int64)
     exceeded_clusters = collect(keys(cluster_frequency))[rule_violators_idx]
     potential_alternatives = cluster_ids[cluster_ids .∉ Ref(exceeded_clusters)]
 
-    return selected_clusters, cluster_frequency, rule_violators_idx, exceeded_clusters, potential_alternatives
+    return selected_clusters,
+    cluster_frequency, rule_violators_idx, exceeded_clusters,
```
Zapiano, line 365:
```suggestion
    cluster_frequency, 
    rule_violators_idx, 
    exceeded_clusters,
```

src/factors/Factors.jl
```diff
@@ -33,27 +33,28 @@ function Factor(val; kwargs...)::Param
...
+function _check_has_required_info(kwargs::NT) where {NT<:NamedTuple}
     @assert haskey(kwargs, :ptype) "Missing factor field `ptype`"
     @assert haskey(kwargs, :dist) "Missing factor field `dist`"
     @assert haskey(kwargs, :dist_params) "Missing factor field `dist_params`"
     @assert haskey(kwargs, :name) "Missing factor field `name`"
     @assert haskey(kwargs, :description) "Missing factor field `description`"
 
     param_dist_types = [
-        "continuous", "ordered categorical", "unordered categorical", "ordered discrete", "discrete"
+        "continuous", "ordered categorical", "unordered categorical", "ordered discrete",
```
Zapiano, line 56:
```suggestion
        "continuous", 
        "ordered categorical", 
        "unordered categorical", 
        "ordered discrete",
```

src/io/rme_result_io.jl
```diff
@@ -169,23 +170,25 @@ groupings for compatibility with ADRIA scenario plotting.
...
+)::Dict{Symbol,BitVector}
     if size(inputs) == (0, 0)
         return Dict(:counterfactual => BitVector([true for _ in 1:n_scenarios]))
     end
 
     counterfactual_scens::BitVector = BitVector([
         p_a == 0.0 && e_a == 0 for (p_a, e_a)
-            in zip(inputs.outplant_area_pct, inputs.enrichment_area_pct)
+        in
+        zip(inputs.outplant_area_pct, inputs.enrichment_area_pct)
```
Zapiano, line 181: I think a better way of doing this would be:

```
inputs.outplant_area_pct .== 0.0 .&&  inputs.enrichment_area_pct .== 0
```

I know, this is not just formatting. Maybe we can ignore this for now, but I just wanted to say that I really don't like this formatting (but again, I'm happy to ignore it).

src/io/rme_result_io.jl
```diff
@@ -169,23 +170,25 @@ groupings for compatibility with ADRIA scenario plotting.
...
     ])
 
     # Intervened if not counterfactual
     intervened_scens::BitVector = BitVector([
         p_a != 0 || e_a != 0 for (p_a, e_a)
-            in zip(inputs.outplant_area_pct, inputs.enrichment_area_pct)
+        in
```
Zapiano, line 187: similar comment as above

src/main_app.jl
```diff
@@ -101,34 +97,33 @@ function _indicative_result_display(res)
...
     upper = Y_g[:upper_95]
     lower = Y_g[:lower_95]
 
-    plot!(upper, fillrange=lower, color=:lightseagreen, alpha=0.4, label="")
-    plot!(Y_g[:median],
+    plot!(upper; fillrange=lower, color=:lightseagreen, alpha=0.4, label="")
+    plot!(Y_g[:median];
         label="Guided median", linecolor=:green, alpha=0.4,
         xlabel="Year", ylabel="Relative Cover",
         xticks=(1:75, year_axis))
```
Zapiano, line 117:
```suggestion
        label="Guided median", 
        linecolor=:green, 
        alpha=0.4,
        xlabel="Year", 
        ylabel="Relative Cover",
        xticks=(1:75, year_axis))
```

src/main_app.jl
```diff
@@ -101,34 +97,33 @@ function _indicative_result_display(res)
...
 
-    p2 = plot(Y_ung[:mean] - Y_no[:mean], label="Guided - No Deployment (μ)",
+    p2 = plot(Y_ung[:mean] - Y_no[:mean]; label="Guided - No Deployment (μ)",
         xlabel="Year", ylabel="δ Relative Cover",
         xticks=(1:75, year_axis), color=:red
     )
```
Zapiano, line 122:
```suggestion
    p2 = plot(
        Y_ung[:mean] - Y_no[:mean]; 
        label="Guided - No Deployment (μ)",
        xlabel="Year", 
        ylabel="δ Relative Cover",
        xticks=(1:75, year_axis), 
        color=:red
    )
```

Zapiano (approved):
I left a lot of comments but, essentially, I don't mind if you ignore all of them. Whatever we decide, I'm happy to follow.

ext/AvizExt/AvizExt.jl
```diff
@@ -134,9 +133,9 @@ end
 function remove_loader(fig, task)
     Base.throwto(task, InterruptException())
     empty!(fig)
```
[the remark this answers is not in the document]
arlowhite: I don't think so. It's given as an argument. There's no current uses, but there is a commented one, which doesn't use a return value:
```
        # Clear loading animation
        # remove_loader(traj_display[2, 1], loader_anim_display)
```

src/main_app.jl
ConnectedSystems: This `main_app` file is marked for removal. It was an earlier attempt to spin up a desktop GUI but isn't relevant for us any more.

ext/AvizExt/viz/rule_extraction.jl
```diff
@@ -130,38 +130,39 @@ function ADRIA.viz.rules_scatter!(
     end
 
     # Create Legend
-    markers = MarkerElement[MarkerElement(color=_c, marker=marker) for _c in colors]
+    markers = MarkerElement[MarkerElement(; color=_c, marker=marker) for _c in colors]
     Legend(
-        sub_g[1, n_cols+1],
+        sub_g[1, n_cols + 1],
```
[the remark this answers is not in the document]
arlowhite: up to you guys.
https://domluna.github.io/JuliaFormatter.jl/stable/#whitespace_ops_in_indices

Also, if we want whitespace in types 
https://domluna.github.io/JuliaFormatter.jl/stable/#whitespace_typedefs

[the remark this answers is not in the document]
ConnectedSystems: It's a config if I recall. I don't mind either way - sometimes with spaces is with more readable, other times without spaces is. Generally I prefer with spaces.

ext/AvizExt/viz/location_selection.jl
```diff
@@ -75,7 +75,7 @@ function ADRIA.viz.ranks_to_frequencies!(
             strokecolor=:grey,
             strokewidth=0.5,
             linestyle=:solid,
-            overdraw=true,
```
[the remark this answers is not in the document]
ConnectedSystems: I have a dissenting opinion.

Moving arguments around is something that's done during prototyping. Once prototyping is done, leaving the comma is untidy. If you have to adjust again (ideally months to years afterwards) then you just add the comma while prototyping.

But I think this discussion is moot because the point of having a formatting workflow is that you can do your thing and once you're finished the formatter standardises.

src/analysis/rule_extraction.jl
arlowhite: that's ok too, if I change to that, the formatter will leave it

src/decision/location_selection.jl
arlowhite: This is a separate rule https://domluna.github.io/JuliaFormatter.jl/stable/#whitespace_typedefs

PR #799, opened 2024-07-10
use apply_mortality not apply_survival

DanTanAtAims:
use `apply_mortality` instead of `apply_survival`.

Relates to [PR in CoralBlox](https://github.com/open-AIMS/CoralBlox.jl/pull/17)


Zapiano (approved):
LGTM

PR #802, opened 2024-07-11
Fix depth min/offset being sampled for counterfactual scenarios

ConnectedSystems:
As in title.

Thanks to @BG-AIMS for catching this.


DanTanAtAims (approved):
Looks good to me.

PR #805, opened 2024-07-18
Use CoralBlox linear_extension_scale_factor

Zapiano:
Also, standardize units to meters.

As in title.

### DIFF
--- a/src/ecosystem/corals/Corals.jl
+++ b/src/ecosystem/corals/Corals.jl
@@ -38,31 +38,33 @@ end
 """
     linear_extensions()
 
-Linear extensions for each
+Linear extensions. The values are converted from `cm` to the desired unit.
+The default unit is `m`.
 """
-function linear_extensions()::Matrix{Float64}
+function linear_extensions(; unit=:m)::Matrix{Float64}
     return [
         0.609456 1.07184 2.55149 5.07988 9.45091 16.8505 0.0;
         0.768556 1.22085 1.86447 2.82297 3.52938 3.00422 0.0;
         0.190455 0.343747 0.615467 0.97477 1.70079 2.91729 0.0;
         0.318034 0.47385 0.683729 0.710587 0.581085 0.581085 0.0;
         0.122478 0.217702 0.382098 0.718781 1.24172 2.08546 0.0
-    ]
+    ] .* linear_scale(:cm, unit)
 end
 
 """
     bin_edges()
 
-Helper function defining coral colony diameter bin edges in cm.
+Helper function defining coral colony diameter bin edges. The values are converted from `cm`
+to the desired unit. The default unit is `m`.
 """
-function bin_edges()
+function bin_edges(; unit=:m)
     return Matrix([
         0.0 5.0 7.5 10.0 20.0 40.0 100.0 150.0;
         0.0 5.0 7.5 10.0 20.0 35.0 50.0 100.0;
         0.0 5.0 7.5 10.0 15.0 20.0 40.0 50.0;
         0.0 5.0 7.5 10.0 20.0 40.0 50.0 100.0;
         0.0 5.0 7.5 10.0 20.0 40.0 50.0 100.0
-    ])
+    ]) .* linear_scale(:cm, unit)
 end
 # function bin_edges()
 #     return Matrix([
@@ -77,11 +79,10 @@ end
 """
     bin_widths()
 
-Helper function defining coral colony diameter bin widths in cm.
+Helper function defining coral colony diameter bin widths.
 """
 function bin_widths()
-    bins::Matrix{Float64} = bin_edges()
-    return bins[:, 2:end] .- bins[:, 1:end-1]
+    return bin_edges()[:, 2:end] .- bin_edges()[:, 1:end-1]
 end
 
 """
@@ -121,7 +122,7 @@ Generate colony area data based on Bozec et al., [1].
 """
 function colony_areas()
     # The coral colony diameter bin edges (cm) are: 0, 2, 5, 10, 20, 40, 80
-    edges = bin_edges()
+    edges = bin_edges(; unit=:cm)
 
     # Diameters in cm
     mean_cm_diameters = edges[:, 1:end-1] + (edges[:, 2:end] - edges[:, 1:end-1]) / 2.0
@@ -129,7 +130,7 @@ function colony_areas()
     # To convert to cover we locate bin means and calculate bin mean areas
     colony_area_mean_cm2 = colony_mean_area(mean_cm_diameters)
 
-    return colony_area_mean_cm2, (mean_cm_diameters ./ 100.0)
+    return colony_area_mean_cm2, linear_scale.(mean_cm_diameters, :cm, :m)
 end
 
 function bins_bounds(mean_diam::Matrix{Float64})::Matrix{Float64}
--- a/src/ecosystem/corals/Corals.jl
+++ b/src/ecosystem/corals/Corals.jl
@@ -36,6 +36,21 @@ function colony_mean_area(colony_diam_means::Array{T})::Array{T} where {T<:Float
 end
 
 """
+    linear_extensions()
+
+Linear extensions for each
+"""
+function linear_extensions()::Matrix{Float64}
+    return [
+        0.609456 1.07184 2.55149 5.07988 9.45091 16.8505 0.0;
+        0.768556 1.22085 1.86447 2.82297 3.52938 3.00422 0.0;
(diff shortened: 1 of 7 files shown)


src/scenario.jl
```diff
@@ -648,51 +660,48 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
...
+            dropdims(sum(C_cover[tstep-1, :, :, habitable_locs], dims=(1, 2)), dims=(1, 2))
+        )
+
+        lin_ext_scale_factors::Vector{Float64} = linear_extension_scale_factors(
+            C_cover_t[:, :, habitable_locs],
+            habitable_loc_areas,
+            linear_extensions(),
+            bin_edges(),
+            max_projected_cover[habitable_locs],
+        )
```
ConnectedSystems, line 676: Move `linear_extensions()` and `bin_edges()` to outside the loop to reduce memory allocations

src/scenario.jl
```diff
@@ -939,8 +949,8 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
 
     # Final reshape
     raw = reshape(permutedims(
-        C_cover, (1, 3, 2, 4)
-    ), (tf, n_group_and_size, n_locs))
+            C_cover, (1, 3, 2, 4)
+        ), (tf, n_group_and_size, n_locs))
```
ConnectedSystems, line 953: Minor formatting: last parenthesis should be on next line.

src/ecosystem/corals/Corals.jl
```diff
@@ -35,19 +35,36 @@ function colony_mean_area(colony_diam_means::Array{T})::Array{T} where {T<:Float
     return pi .* ((colony_diam_means ./ 2.0) .^ 2)
 end
 
+"""
+    linear_extensions()
+
+Linear extensions. The values are converted from `cm` to the desired unit.
+The default unit is `m`.
+"""
+function linear_extensions(; unit=:m)::Matrix{Float64}
```
ConnectedSystems, line 44: Waiting on further clarification as to why the unit conversion is necessary, and why meters over centimeters

src/ecosystem/corals/growth.jl
```diff
@@ -860,7 +860,7 @@ Calculates coral recruitment for each species/group and location.
 function recruitment_rate(larval_pool::AbstractArray{T,2}, A::AbstractArray{T};
     α::Union{T,Vector{T}}=2.5, β::Union{T,Vector{T}}=5000.0)::Matrix{T} where {T<:Float64}
 
-    sd = settler_density.(α, β, larval_pool) .* A
+    sd = settler_density.(α, β, larval_pool) .* A'
```
ConnectedSystems, line 863: Obviously this works but puzzled as to why the transform is necessary when it wasn't before?

I don't see a change in what is passed into `recruitment_rate` (in this PR at least)

src/scenario.jl
```diff
@@ -149,18 +151,26 @@ function run_scenarios(
...
+            eachrow(zeros(n_groups, n_sizes))
         ) for _ in 1:n_locs
     ]
 
+    loc_habitable_area = site_k_area(dom)
+    max_projected_cover = CoralBlox.max_projected_cover(
+        linear_extensions(),
+        bin_edges(),
+        loc_habitable_area
+    )
```
ConnectedSystems, line 169: Micro-performance change:

`bin_edges()` could be called once (assign to a variable), to avoid allocating three separate times.

src/scenario.jl
```diff
@@ -586,10 +598,10 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
     end
 
     # Set up distributions for natural adaptation/heritability
-    c_mean_t_1::Array{Float64, 3} = repeat(_to_group_size(
-        domain.coral_growth, corals.dist_mean
-    ), 1, 1, n_locs)
-    c_std::Array{Float64, 2} = _to_group_size(
+    c_mean_t_1::Array{Float64,3} = repeat(_to_group_size(
+            domain.coral_growth, corals.dist_mean
+        ), 1, 1, n_locs)
```
ConnectedSystems, line 603: Okay to ignore, just flagging this formatting looks weird to me?

src/scenario.jl
ConnectedSystems: Doing so will address this comment too: https://github.com/open-AIMS/ADRIA.jl/pull/805/files#r1682389979

ConnectedSystems (commented):
Minor comments, the most important one being the request to provide more detail/context on the linear scaling.

src/scenario.jl
```diff
@@ -648,51 +660,48 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
 
     FLoops.assistant(false)
     for tstep::Int64 in 2:tf
-        survival_rate_slices = [@view survival_rate_cache[:, :, loc] for loc in 1:n_locs]
-        apply_mortality!.(functional_groups, survival_rate_slices)
-        recruitment .*= reshape(survival_rate_cache[:, 1, :], (n_groups, n_locs))
-
-        # Determine absolute area for coral model
+        # Convert covert to absolute values to use within CoralBlox model
```
DanTanAtAims, line 663:
```suggestion
        # Convert cover to absolute values to use within CoralBlox model
```

src/scenario.jl
```diff
@@ -389,11 +400,11 @@ NamedTuple of collated results
...
 - `coral_dhw_log` : Array, Log of DHW tolerances / adaptation over time (only logged in debug mode)
 """
-function run_model(domain::Domain, param_set::DataFrameRow, functional_groups::Vector{Vector{FunctionalGroup}})::NamedTuple
+function run_model(domain::Domain, param_set::DataFrameRow, functional_groups::Vector{Vector{FunctionalGroup}}, max_projected_cover::Vector{Float64})::NamedTuple
     ps = DataCube(Vector(param_set); factors=names(param_set))
-    return run_model(domain, ps, functional_groups)
+    return run_model(domain, ps, functional_groups, max_projected_cover)
 end
-function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vector{Vector{FunctionalGroup}})::NamedTuple
+function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vector{Vector{FunctionalGroup}}, max_projected_cover::Vector{Float64})::NamedTuple
```
ConnectedSystems, line 407: Hang on, this now breaks usage.

Now there's no way to run the model directly without having to also specify these `functional groups` and `max_projected_cover`

Possible fix is to define a `run_model()` method that handles this.

Zapiano: Since this is already happening in the other branches that integrate with `CoralBlox`, I propose that we fix that on a separate PR.

src/ecosystem/corals/Corals.jl
Zapiano: When we convert from relative to absolute cover, we use `habitable_loc_areas′`. This area comes from `domain.site_data.area`. We can use `linear_extension` and `bin_edges` in cm, but then we also need to convert this area to cm^2.

src/ecosystem/corals/growth.jl
Zapiano: tl dr;
Before we were doing, inside scenario.jl 

```
loc_coral_cover = dropdims(sum(C_cover_t; dims=(1, 2)), dims=1)
```

and now this is being done inside the function `_loc_coral_cover` as :

```
dropdims(sum(C_cover_t; dims=(1, 2)), dims=(1, 2))
```

----- 

This is because `A`, that is called `leftover_space` inside the  function`settler_cover`, comes from `leftover_space_m²`, in `scenario.jl`: https://github.com/open-AIMS/ADRIA.jl/blob/d45cade0b6ca8f3a88edef47764f09d399289fe7/src/scenario.jl#L724

The argument that is being passed to the function `relative_leftover_space` is `loc_coral_cover`, which now is being computed inside the new function `_loc_coral_cover`. I implemented this function because we do this operation in different places so I extracted that to a function.

To standardize I had to drop dims 1 and 2, and not only 2.

src/scenario.jl
ConnectedSystems: Okie dokies

src/ecosystem/corals/Corals.jl
ConnectedSystems: Lets resolve in a separate PR then

ConnectedSystems (approved):
Okay with me, but two further changes (maintaining units in cm, and fixing `run_model`) to be addressed in separate PRs.

PR #807, opened 2024-07-24
Fix add run_model method that do not require functional_groups

Zapiano:
As in title.

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -153,7 +153,7 @@ function run_scenarios(
     n_locs::Int64 = dom.coral_growth.n_locs
     n_sizes::Int64 = dom.coral_growth.n_sizes
     n_groups::Int64 = dom.coral_growth.n_groups
-    _bin_edges = bin_edges()
+    _bin_edges::Matrix{Float64} = bin_edges()
     functional_groups = [
         FunctionalGroup.(
             eachrow(_bin_edges[:, 1:end-1]),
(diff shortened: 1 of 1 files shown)


src/scenario.jl
```diff
@@ -646,7 +660,7 @@ function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vecto
 
     # Cache matrix to store potential settlers
     potential_settlers = zeros(size(fec_scope)...)
-    linear_extension = _to_group_size(domain.coral_growth, corals.linear_extension)
+    _linear_extensions = linear_extensions() # _to_group_size(domain.coral_growth, corals.linear_extension)
```
ConnectedSystems, line 663: This breaks linear extension values being provided by the sampler.

src/scenario.jl
```diff
@@ -401,11 +395,31 @@ NamedTuple of collated results
...
+    param_set::DataFrameRow,
+    functional_groups::Vector{Vector{FunctionalGroup}}
+)::NamedTuple
     ps = DataCube(Vector(param_set); factors=names(param_set))
-    return run_model(domain, ps, functional_groups, max_projected_cover)
+    return run_model(domain, ps, functional_groups)
 end
-function run_model(domain::Domain, param_set::YAXArray, functional_groups::Vector{Vector{FunctionalGroup}}, max_projected_cover::Vector{Float64})::NamedTuple
+function run_model(
+    domain::Domain, param_set::YAXArray, functional_groups::Vector{Vector{FunctionalGroup}}
```
ConnectedSystems, line 421: Is there a reason why `functional_groups` cannot be created inside `run_model()`?

It would simplify the chain of functions being called.

ConnectedSystems (changes requested):

src/scenario.jl
Zapiano: They are being created outside so we can use as a cache.

src/scenario.jl
Zapiano: Ok, I'll fix that.

Zapiano:
@ConnectedSystems I fixed the `linear_extension` issue you mentions. About your second comment, there is an option to call `run_model` without passing `functional_groups` but I kept the version that accepts it as argument. Do you want me to change anything?

Zapiano:
@ConnectedSystems this branch is rebased and ready for a review.

ConnectedSystems (approved):

PR #811, opened 2024-07-29
fix permutation of initial cover dimensions

DanTanAtAims:
Unfortunately, I appeared to have incorrectly loaded RMEDomain initial cover when migrating to use CoralBlox. @Zapiano @ConnectedSystems 

This wouldn't impact calibration runs as initial cover is overwritten.

### Before

The first taxa has very small initial cover which is actually the initial cover values for the small size classes and the last taxa has very large cover values.

bugged_init_cover

2 counterfactuals
bug_not_fix

### After

bug_fix_init_cover

bug_fix


Zapiano:
@DanTanAtAims I think this need a rebase

DanTanAtAims:
@Zapiano Rebased on to updated `DanTan/CircularBuffer`. Ready for review.

Zapiano (approved):
LGTM

PR #812, opened 2024-07-30
Breaks new initial_cover.nc file into size classes assuming a LogNormal distribution

Zapiano:
This responsibility was moved from `rrap-dg` to `ADRIA`. This depends on a new version of the datapackage `0.7.0-rc` with the new `coral_cover.nc` file generated with `https://github.com/open-AIMS/rrap-dg/pull/14` . 

Running 128 scenarios with this new cover we get:

image

Edit: depends on #813


DanTanAtAims:
@Zapiano This branch should be rebased now

src/io/initial_coral_cover.jl
```diff
@@ -0,0 +1,96 @@
...
+    NetCDF,
+    YAXArrays
+
+"""
+    load_cover(data_fn::String)::YAXArray
+    load_cover(n_species::Int64, n_sites::Int64)::YAXArray
+
+Load initial coral cover data from netCDF.
+"""
+function load_cover(data_fn::String)::YAXArray
```
DanTanAtAims, line 11: ReefmodDomain and RMEDomain implement a function called `load_initial_cover`. Should the naming be the same as the others?
```suggestion
function load_initial_cover(data_fn::String)::YAXArray
````
I'm happy with either naming

**Review by DanTanAtAims — COMMENTED** [review]
Looks good. Runs fine.

File: src/io/initial_coral_cover.jl (hunk 1 above)
Comment (ConnectedSystems) [reply src/io/initial_coral_cover.jl:11]: Yes, best to keep it inconsistent

File: src/io/inputs.jl [hunk 2]
```diff
@@ -1,6 +1,7 @@
-using JSON
-using YAXArrays
-using NetCDF
+using Distributions,
+    JSON,
+    NetCDF,
+    YAXArrays
```
ConnectedSystems, line 4: Minor note, can be ignored for now but in future I'd prefer imports to be grouped by conceptual purpose:

```
using Distributions
using
    JSON,
    NetCDF,
    YAXArrays

ConnectedSystems (approved):
Looks good to me

src/io/initial_coral_cover.jl
Zapiano: I was actually thinking about doing this change but for another reason, so I'm happy to change.

Zapiano:
@ConnectedSystems @DanTanAtAims this is ready to be merged, but please wait for #813 to be merged first.

Zapiano:
Quick comment: I've rebased this onto #813 and now the tests are passing.

PR #823, opened 2024-08-13
Fix docs builder, add .git-blame-ignore-revs and document

arlowhite:
The doc builder failed:
https://github.com/open-AIMS/ADRIA.jl/actions/runs/10361183116/job/28681006319
I've regenerated the `docs` Manifest, which should fix this.

Also, added `.git-blame-ignore-revs`, which ignores the "re-format entire project" commit from #790

Documented how to activate `.git-blame-ignore-revs`


Zapiano (approved):
LGTM

PR #824, opened 2024-08-15
fix helper dimension size error

DanTanAtAims:
This should allow RMEDomain and ReefModDomain runs to load successfully.

### DIFF
--- a/src/ecosystem/corals/CoralGrowth.jl
+++ b/src/ecosystem/corals/CoralGrowth.jl
@@ -18,7 +18,7 @@ end
 Implements temporary hardcoded caches for a scenario with 35 'species' (split into 5 groups).
 """
 function CoralGrowth(n_locs::Int64)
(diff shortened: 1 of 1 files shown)


src/ecosystem/corals/CoralGrowth.jl
```diff
@@ -18,7 +18,7 @@ end
 Implements temporary hardcoded caches for a scenario with 35 'species' (split into 5 groups).
 """
 function CoralGrowth(n_locs::Int64)
-    n_groups::Int64, n_sizes::Int64 = 5, 6
+    n_groups::Int64, n_sizes::Int64 = 5, 7
```
Zapiano, line 21:
```suggestion
    n_groups::Int64, n_sizes::Int64 = size(linear_extensions())
```

Zapiano (approved):

PR #827, opened 2024-08-19
Add reshape to fix dimension mismatch when logging dhw tolerances

### LINKED ISSUE #826: Dimension mismatch when logging DHW tolerances in debug mode
Possibly due to recent model updates, a dimension mismatch error now occurs running in debug mode in the most recent version of the model and domain (0.7.0). 
```julia dom = ADRIA.load_domain( "Moore_2024-08-09_v070_rc1","45")
scens = ADRIA.sample(dom, 2^8)
rs = ADRIA.run_scenarios(dom, scens, ["45"])
```

Error message:
```julia ERROR: LoadError: DimensionMismatch: array could not be broadcast to match destination```

Which can be traced to this line in `scenario.jl`
https://github.com/open-AIMS/ADRIA.jl/blob/22e3c2877ea9248a0f24d95449cf5ea74e82fdd2/src/scenario.jl#L762-L764

Seem like a small update needed to the storage container for dhw tolerance logs.

Rosejoycrocker:
Resolves #826

Reshape retains order of size classes then species. Seems some repetitiveness though in logged values, maybe it can be further aggregated?

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -761,7 +761,9 @@ function run_model(
 
             if in_debug_mode
                 # Log dhw tolerances if in debug mode
-                dhw_tol_mean_log[tstep, :, :] .= mean.(c_mean_t)
+                dhw_tol_mean_log[tstep, :, :] .= reshape(
+                    mean.(c_mean_t), size(dhw_tol_mean_log)[2:3]
+                )
             end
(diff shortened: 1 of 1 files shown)


Zapiano (approved):
🚀

PR #832, opened 2024-08-21
Fix docs errors

arlowhite:
This fixes the errors that appeared after upgrading `Documenter` from `0.27.25` to `1.5.0`

Example Error:
```
┌ Error: invalid local link/image: file does not exist in src\usage\analysis.md
│   link =
│    @ast MarkdownAST.Image("../assets/imgs/analysis/single_rank_plot.png \"title text breaks\"", "") do
│      MarkdownAST.Text("Rank frequency plots for single rank")
│    end
```

Also, ran Pkg `update`, so now we're on `1.6.0`

# Summary

title text doesn't work in current docs, and `raw=true` seems to have no effect for our largest png on GitHub. So I don't think we're losing anything with the changes in this PR.

# Notes

These errors appeared because Documenter 1.0 switched to strict mode.
https://github.com/JuliaDocs/Documenter.jl/blob/v1.6.0/CHANGELOG.md#version-v100---2023-09-15

Also, Documenter does not support image titles. 
`alt text`

If we really want to use title (hover tooltip), we could ask Documenter to implement it or submit the feature. But maybe better to have this text as a caption to the image.

Documenter documentation states that `./assets` should work, but it doesn't seem to.

Also, needed to remove `?raw=true`. This makes GitHub serve the full raw image instead of a possibly downscaled one.
However, looking at one of our images deployed now, with and without raw has the same size.
https://open-aims.github.io/ADRIA.jl/dev/assets/imgs/analysis/rules_scatter.png?raw=true  
_(this is our largest image)_

## Alternative

If we want title and/or `raw=true`, Documenter lets us change this error to a warning as mentioned in 1.0 release notes.
`warnonly = :cross_references`

### DIFF
--- a/docs/Manifest.toml
+++ b/docs/Manifest.toml
@@ -272,12 +272,6 @@ git-tree-sha1 = "6c834533dc1fabd820c1db03c839bf97e45a3fab"
 uuid = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 version = "0.10.14"
 
-[[deps.Calculus]]
-deps = ["LinearAlgebra"]
-git-tree-sha1 = "f641eb0a4f00c343bbc32346e1217b86f3ce9dad"
-uuid = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
-version = "0.5.1"
-
 [[deps.CategoricalArrays]]
 deps = ["DataAPI", "Future", "Missings", "Printf", "Requires", "Statistics", "Unicode"]
 git-tree-sha1 = "1568b28f91293458345dabba6a5ea3f183250a61"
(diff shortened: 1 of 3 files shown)


arlowhite:
Actually, `title` isn't working in current docs either:
https://open-aims.github.io/ADRIA.jl/dev/usage/analysis/
```html
<img src="/ADRIA.jl/dev/assets/imgs/analysis/scenarios_tac.png?raw=true &quot;Quick scenario plot&quot;" alt="Quick scenario plots">
```
It's part of `src` but being ignored. I'm surprised this doesn't break.

ConnectedSystems:
We can try this, though not sure how well it will go.

Relevant conversation/workaround: https://github.com/JuliaDocs/Documenter.jl/issues/921#issuecomment-1369660732

ConnectedSystems (approved):

PR #838, opened 2024-08-27
Cscape interface improvements

ConnectedSystems:
Reduced ResultSet load times from \~2:10 mins to \~27 seconds for a 20 scenario dataset.
Note, improvements will be more noticeable if more each C~Scape dataset holds more than a single scenario.

Achieved by:
1. A "smarter" indexing strategy used to read C~Scape results into a common data structure
    Precalculates the indices for the scenarios in each dataset to avoid `if` branching in the loop
2. avoiding loading/calculating data where results are all zeros, 
3. reading in data into memory prior to any calculations

I've also organized the function definitions so that methods with the same names are together in the file, as per style guide.

I also applied the formatter so it passes the format check we now have in place.

### DIFF
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -704,14 +704,14 @@ function _cscape_relative_cover(dataset::Dataset)::Array
     ) .* reshape(
         area[1, :] .* dataset.k[:],
         reshape_tuple
-    ) ./ 100
+    ) ./ 100.0
 
     relative_cover .+= _drop_sum(
         dataset.cover[intervened=2], dim_sum
     ) .* reshape(
         area[2, :] .* dataset.k[:],
         reshape_tuple
-    ) ./ 100
+    ) ./ 100.0
 
     # ADRIA assumes a shape of [timesteps ⋅ locations ⋅ scenarios]
     if multi_scenario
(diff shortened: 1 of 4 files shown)


src/io/cscape_result_io.jl
```diff
@@ -724,36 +766,49 @@ function _n_scenarios(dataset::Dataset)::Int64
...
+    ends[1] = n_per_dataset[1]
+
+    for i in 2:n
+        starts[i] = ends[i-1] + 1
+        ends[i] = starts[i] + n_per_dataset[i] - 1
     end
-    return relative_cover
+
+    return collect(zip(starts, ends))
 end
```
ConnectedSystems, line 812: I would appreciate a logic check here 😄

ConnectedSystems:
Whoops, @DanTanAtAims I neglected to update the docstring detailing the expected directory structure for the C~Scape datasets.

Can't do it today but please hold off any approvals until that is done.

ConnectedSystems:
@DanTanAtAims I've added a link to the section to the docs which explains the expected directory structure.

I expect this should work but we'll see.

DanTanAtAims (approved):
Thanks, looks good to me.

PR #844, opened 2024-09-09
Overly constrained argument types causing tests to fail

ConnectedSystems:
Recent PR #702 included a type that was overly constrained which causes tests to fail.

Resolved by allowing any AbstractVector holding values <: Real.


ConnectedSystems:
pinging @PeterBaker0 - this should resolve the failing test you encountered.

DanTanAtAims (approved):
Tests pass. Looks good to me.

PR #846, opened 2024-09-10
fix reefmod domain load initial cover

### LINKED ISSUE #825: ReefModDomain loading is not working
When I try to load ReefModDomain:

```julia
reefmod_dom = ADRIA.load_domain(ReefModDomain, <path_to_reefmoddomain>, "45")
```

I get the following error: 

```
ERROR: ArgumentError: Can not construct YAXArray, supplied data size is (234, 3806) while axis lengths are (36, 3806)
Stacktrace:
 [1] YAXArrays.Cubes.YAXArray(axes::Tuple{…}, data::Matrix{…}, properties::Dict{…}, chunks::DiskArrays.GridChunks{…}, cleaner::Vector{…})
   @ YAXArrays.Cubes C:\Users\pribeiro\AIMS\Code\YAXArrays.jl\src\Cubes\Cubes.jl:114
 [2] YAXArray
   @ C:\Users\pribeiro\AIMS\Code\YAXArrays.jl\src\Cubes\Cubes.jl:136 [inlined]      
 [3] YAXArrays.Cubes.YAXArray(axes::Tuple{…}, data::Matrix{…})
   @ YAXArrays.Cubes C:\Users\pribeiro\AIMS\Code\YAXArrays.jl\src\Cubes\Cubes.jl:136
 [4] DataCube(data::Matrix{Float64}; kwargs::@Kwargs{species::UnitRange{Int64}, locs::Vector{String}})
   @ ADRIA c:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\io\datacubes.jl:14
 [5] load_initial_cover(::Type{…}, dom_data::YAXArrays.Datasets.Dataset, site_data::DataFrames.DataFrame, loc_ids::Vector{…}, init_yr::Int64)
   @ ADRIA c:\Users\pribeiro\AIMS\Code\ADRIA.jl\src\ExtInterface\ReefMod\ReefModDomain.jl:234
``` 

What seems to be happening is that on this line https://github.com/open-AIMS/ADRIA.jl/blob/8808b0b0a13df6585213275b935f66399cdb4f36/src/ExtInterface/ReefMod/ReefModDomain.jl#L216https://github.com/open-AIMS/ADRIA.jl/blob/8808b0b0a13df6585213275b935f66399cdb4f36/src/ExtInterface/ReefMod/ReefModDomain.jl#L216

We should be doing `size_class_weights = (cdf_integral[2:end] .- cdf_integral[1:(end - 1)])`

Also, the number of size classes is hard coded (which we could avoid) and we are not ignoring the first functional group. So I this change I'm proposing won't fix everything.

DanTanAtAims:
Corrects the domain to only load 5 functional groups instead of 6. 

closes #825

### DIFF
--- a/src/ExtInterface/ReefMod/ReefModDomain.jl
+++ b/src/ExtInterface/ReefMod/ReefModDomain.jl
@@ -247,7 +247,9 @@ function load_initial_cover(
     icc_data = icc_data ./ site_data.k'
     icc_data_sum = dropdims(sum(icc_data, dims=1), dims=1)
     if any(icc_data_sum .> 1.0)
-        @warn "Initial cover exceeds habitable area, there is most likely an issue with the data package. Constraining to 1.0"
+        msg = "Initial cover exceeds habitable area, "
+        msg *= "there is most likely an issue with the data package. Constraining to 1.0"
+        @warn msg
         icc_data[:, icc_data_sum .> 1.0] ./= icc_data_sum[icc_data_sum .> 1.0]'
     end
 
--- a/src/ExtInterface/ReefMod/ReefModDomain.jl
+++ b/src/ExtInterface/ReefMod/ReefModDomain.jl
@@ -194,15 +194,24 @@ end
 function load_initial_cover(
     ::Type{ReefModDomain},
     dom_data::Dataset,
-    location_data::DataFrame,
+    site_data::DataFrame,
     loc_ids::Vector{String},
     init_yr::Int=2022
 )::YAXArray
     if !haskey(dom_data.cubes, :coral_cover_per_taxa)
         @error "coral_cover_per_taxa variable not found in ReefMod data"
     end
-    init_cc_per_taxa::YAXArray = Cube(dom_data[["coral_cover_per_taxa"]])[time=At(init_yr)]
 
+    c_axes = caxes(dom_data.coral_cover_per_taxa)
+    init_cc_per_taxa::YAXArray = DataCube(
+        read(dom_data.coral_cover_per_taxa);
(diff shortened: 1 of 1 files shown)


src/ExtInterface/ReefMod/ReefModDomain.jl
```diff
@@ -194,15 +194,24 @@ end
...
+    site_data::DataFrame,
     loc_ids::Vector{String},
     init_yr::Int=2022
 )::YAXArray
     if !haskey(dom_data.cubes, :coral_cover_per_taxa)
         @error "coral_cover_per_taxa variable not found in ReefMod data"
     end
-    init_cc_per_taxa::YAXArray = Cube(dom_data[["coral_cover_per_taxa"]])[time=At(init_yr)]
 
+    c_axes = caxes(dom_data.coral_cover_per_taxa)
```
Zapiano, line 205: You can leave it this way if you want but you could also do something like:

```
cover_axes_names = ADRIA.axes_names(dom_data.coral_cover_per_taxa) 
cover_axes = ADRIA.axis_labels.([dom_data.coral_cover_per_taxa], cover_axes_names)
```

And we could also write a `ADRIA.axes_labels` that accepts just the datacube and returns an array of labels for all axes. This way you don't have to do this `Int64.(collect(..))`

Zapiano (approved):
Just two minor comments.

src/ExtInterface/ReefMod/ReefModDomain.jl
DanTanAtAims: I believe this is done because of an issue with DiskArrays and Floating point axis labels.

PR #847, opened 2024-09-12
Add metadata to metrics YAXArrays

Zapiano:
- Update metrics within the following files:
  - scr/metrics/metrics.jl
  - scr/metrics/scenario.jl
  - scr/metrics/reef_indices.jl
- Update setup_result_store! and load_results to save and read this metadata to/from result set
- Add new method to helper methods DataCube and ZeroDataCube to allow creating YAXArrays with properties
- Add tests to all metrics updated
- Add MockData module to tests


src/metrics/metrics.jl
```diff
@@ -32,8 +34,12 @@ Metric(f, d) = Metric(f, d, "")
 Makes Metric types callable with arbitary arguments that are passed to associated function.
 """
 function (f::Metric)(raw, args...; kwargs...)::YAXArray
-    data = DataCube(raw, (:timesteps, :species, :locations, :scenarios)[1:ndims(raw)])
-    return f.func(data, args...; kwargs...)
+    axes::Tuple = (:timesteps, :species, :locations, :scenarios)[1:ndims(raw)]
+    properties::Dict{Symbol,Any} = Dict(
+        :metric_name => metric_label(f),
+        :metric_unit => f.unit
+    )
```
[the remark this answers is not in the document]
Zapiano: I think so

test/factories/coral_cover.jl
```diff
@@ -0,0 +1,60 @@
+using ADRIA: YAXArray, DataCube, ZeroDataCube
+
+"""Raw model coral cover results, proportional to k-area."""
+function coral_cover(
```
Zapiano, line 4: I just realized that the one I'm calling `coral_cover` is the one that sums up to one and the one I'm calling `full_coral_cover` is the one that doesn't. I'll fix that.

test/metrics/metrics.jl
```diff
@@ -0,0 +1,215 @@
...
+
+function test_relative_metric(metric::metrics.Metric, params::Tuple)::Nothing
+    return _test_metrics(metric, params; relative=true)
+end
+
+function test_absolute_metric(metric::metrics.Metric, params::Tuple)::Nothing
+    return _test_metrics(metric, params; relative=false)
+end
+
+@testset "metrics.jl" begin
```
ConnectedSystems, line 27: This needs a more informative name for the test set

test/factories/coral_cover.jl
```diff
@@ -0,0 +1,60 @@
...
+"""Raw model coral cover results, proportional to k-area."""
+function coral_cover(;
+    n_timesteps::Int64=10,
+    n_group_sizes::Int64=35,
+    n_locations::Int64=5,
+    n_scenarios::Int64=32
+)::YAXArray{Float64,4}
+    coral_cover = Factories.full_coral_cover(;
+        n_timesteps, n_group_sizes, n_locations, n_scenarios
+    )
```
ConnectedSystems, line 12: `Factories` is not a great name, but also, is this exported directly from ADRIA? I don't see how it's imported for use here.

test/metrics/metrics.jl
```diff
@@ -0,0 +1,215 @@
...
+        coral_cover, zero_coral_cover, full_coral_cover
+    ]
+    k_area::Vector{Float64} = round.(rand(n_locations) .* 100; digits=2)
+
+    @testset "relative_cover" begin
+        test_relative_metric(metrics.relative_cover, (TEST_RS,))
+        for cover in test_covers
+            test_relative_metric(metrics.relative_cover, (cover,))
+        end
+    end
```
ConnectedSystems, line 57: Should we split this file according to the metric types so the tests are more organised?

e.g., a file for the various kinds of `cover`, another for `shelter`, another for `biodiversity` (which is currently just evenness), etc?

test/metrics/metrics.jl
Zapiano: Hum, it's supposed to be the test for all metrics within metrics.jl file. Do you think maybe "matrics.jl file" would be better?

test/metrics/metrics.jl
Zapiano: In my opinion we should reorganize the metrics module and then reorganize this to reflect that. The way it is now it seems easier for me to find a particular test I'm looking for if it is organized by file. The problem, again, is that the file organization is not so good right now.

test/factories/coral_cover.jl
Zapiano: I've used Factories inspired by Factories in Ruby [1]. But I'm happy to change to something else. Any suggestions?

[1] https://stackoverflow.com/questions/37364454/what-is-a-ruby-factory-method

Zapiano: About how it is imported, I'm including it in `runmodel.jl`.

ConnectedSystems: > I've used Factories inspired by Factories in Ruby [1]. But I'm happy to change to something else. Any suggestions?
> 
> [1] https://stackoverflow.com/questions/37364454/what-is-a-ruby-factory-method

Yes, I'm familiar with the factory pattern (it's not specific to Ruby by the way).

The current name is too general. In the example linked above they name it `PersonFactory` which at least indicates what the factory is for, but even that is for an example. Equally, (I hope) you'd agree `PersonConstructor` is not a great name.

What does the factory do?
From what I can see it creates mock coral cover data, so this specific use could be `mock_full_coral_cover`.

Some more questions:

1. Where is `runmodel.jl`? I don't see it in the list of files modified by this PR
2. I would call `full_coral_cover` -> `maximum_coral_cover`. Or at least explain more in the docstring ("represent condition where all locations have maximum possible coral cover").

test/metrics/metrics.jl
ConnectedSystems: You could equally organise the tests and change the metric module to reflect the tests.

> The way it is now it seems easier for me to find a particular test I'm looking for if it is organized by file.

I'm not sure I follow. The top-level name is `metrics.jl` which the corresponding file, as you say (and I agree) is not organised well, so I'm not sure how it helps?

I `ctrl+shift+f` and search for the specific metric.

test/metrics/metrics.jl
```diff
@@ -0,0 +1,215 @@
...
+    return nothing
+end
+
+function test_relative_metric(metric::metrics.Metric, params::Tuple)::Nothing
+    return _test_metrics(metric, params; relative=true)
+end
+
+function test_absolute_metric(metric::metrics.Metric, params::Tuple)::Nothing
+    return _test_metrics(metric, params; relative=false)
+end
```
ConnectedSystems, line 25: This could be simplified by checking if `absolute` or `relative` appears in the metric name.

I believe all coral cover metrics should follow that naming pattern.

Then you don't need the specific `test_relative_*`/`test_absolute_*` functions.

test/metrics/metrics.jl
```diff
@@ -0,0 +1,215 @@
...
+        end
+    end
+
+    @testset "relative_loc_taxa_cover" begin
+        for cover in test_covers
+            test_relative_metric(
+                metrics.relative_loc_taxa_cover, (cover[:, :, :, 1], k_area, n_groups)
+            )
+        end
+    end
```
ConnectedSystems, line 83: For these ones that don't need any special setup I wonder if you can simply generate the tests leveraging the metadata?

```julia
using ADRIA

all_metrics = names(metrics, all=true)
metric_identifiers = ADRIA.human_readable_name(all_metrics)

relative_metrics = occursin.("relative", metric_identifiers)
absolute_metrics = occursin.("absolute", metric_identifiers)
test_m_id = vcat(findall.([relative_metrics, absolute_metrics])...)

for m_id in test_m_id
    metric_to_test = getfield(metrics, all_metrics[m_id])
    params = ...  # infer appropriate mock data from metric metadata?

    # This would apply the appropriate test if relative or not
    _test_metrics(metric_to_test, params)

    # Or keep the flag
    _test_metrics(metric_to_test, params; is_relative=relative_metrics[m_id])
end


```

ConnectedSystems: One issue I have with the current approach is that you have to manually maintain all the tests, which will be a pain.

The "vision" for the metrics were to define a "base" set of metrics and the rest would be auto-generated (including tests). I just never got around to sitting down and working out the necessary meta-programming to do this.

test/viz.jl
```diff
@@ -0,0 +1,322 @@
...
+using ADRIA
+using ADRIA.metrics: total_absolute_cover
+using Statistics
+
+if !@isdefined(TEST_RS)
+    const TEST_DOM, TEST_N_SAMPLES, TEST_SCENS, TEST_RS = test_rs()
+end
+
+"""Test larger scenario run with figure creation"""
+function test_rs_w_fig(rs::ADRIA.ResultSet, scens::ADRIA.DataFrame)
```
ConnectedSystems, line 11: This test wasn't specifically for vizualizations, it's intended as a high-level pass at an end-to-end analysis.

It was included because:

1. Historically we're bad at making sure analyses that we have done in the past did not break
2. We're generally bad at making sure changes did not affect visualizations.

In testing terms, this is our stand-in for an integration test.

test/metrics/metrics.jl
ConnectedSystems: Personally, I would split the test set into the specific aspects of the metrics that are being tested.

src/utils/text_display.jl
```diff
@@ -27,6 +28,11 @@ end
 function human_readable_name(name::Vector{Symbol}; title_case::Bool=false)::Vector{String}
     return human_readable_name(String.(name); title_case=title_case)
 end
+function human_readable_name(
+    names::NTuple{N,Symbol}; title_case::Bool=false
+)::NTuple{N,String} where N
+    return human_readable_name.(String.(names); title_case=title_case)
+end
```
ConnectedSystems, line 35: I think I prefer either mandating that the `names` argument should always be a `Vector{Symbol}` or expanding the allowable data type on the existing implementation.

Currently you have duplicate implementations

```julia
function human_readable_name(names::Union{Vector{T}, NTuple{N, T}}; title_case::Bool=false)::Vector{String} where {T<:Symbol,N<:Int64}
    return human_readable_name(String.(name); title_case=title_case)
end
```

PS: I could be wrong but I think `::NTuple{N, String} where N` may be equivalent to `NTuple{Any, String}`

src/utils/text_display.jl
```diff
@@ -27,6 +28,11 @@ end
 function human_readable_name(name::Vector{Symbol}; title_case::Bool=false)::Vector{String}
     return human_readable_name(String.(name); title_case=title_case)
 end
```
ConnectedSystems, line 30: Not related to this PR but now I'm looking at this, I don't think this function is working as expected.

Should this really be:

```julia
function human_readable_name(names::Vector{Symbol}; title_case::Bool=false)::Vector{String}
    return human_readable_name.(names; title_case=title_case)
end
```

src/utils/text_display.jl
Zapiano: about the PS:
image

ConnectedSystems: What I meant was that `N` is interpreted to be `Any`. I wasn't claiming that it was valid syntax

Zapiano: Oh, I see. So, is there a way (that you know of) of declaring a NTuple without specifying the number of elements?

ConnectedSystems: In my snippet above, I declare `N` as only subtyping `Int64` so the compiler can specialise on that

test/factories/coral_cover.jl
ConnectedSystems: You said `runmodel.jl`

> About how it is imported, I'm including it in `runmodel.jl`.

arlowhite: I agree that `Factories` is too vague. If this is for generating mock data, then why not a name like `MockData`?
`MockData.full_coral_cover`?
I think it would be cleaner to have more modules with meaningful names rather than putting lots of misc things into Factories.

As far as static constructor functions, I've seen "build" or "create" in names. e.g. `Person.create`, so if you prefer:
`MockData.create_full_coral_cover`

or `mock_full_coral_cover` as Takuya suggested.

Zapiano: I don't mind switching to `MockData`. I don't like using "create" because all functions in this module will be to create mock data so all of them will have this prefix and that will make the function names too big. 

Do we agree with the module being called `MockData` then?

Zapiano:
Update 1: I've added updates to other metrics as well.
Update 2: I've updated Factories to MockData 

FYI @ConnectedSystems @arlowhite

src/io/result_io.jl
```diff
@@ -400,54 +400,51 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
     end
 
-    met_names = [
-        :relative_cover,
-        :relative_shelter_volume,
-        :absolute_shelter_volume,
-        :relative_juveniles,
-        :juvenile_indicator,
-        :coral_evenness
+    outcome_metrics::Vector{metrics.Metric} = [
```
arlowhite, line 403: Does this need to be updated when a new metric is added? Is that documented?
Also, should it be at the top of the file or in some other file?

Zapiano: I'm not sure how much you know about that so I'm sorry if I'm explaining something you already know: one thing is the `metrics` module and all metrics inside it; another thing are the outcomes that are saved in the ResultSet after the model is run. This vector you are pointing to has the metrics that correspond to these outcomes. 

About your questions: Everytime we decide to save a new outcome to the ResultSet this vector needs to be updated, but I don't think that this happened in the last year or so. All metrics within the `metrics` module are documented (some better than others). The set of outcomes saved in the ResultSet is not. This vector is only relevant in the context of this function that is building the `result_store` (the zarr that will hold this outcomes, among other things). 

I don't see a reason to move this to another file or to the top of this file, but I don't mind doing it. I don't think this needs to be documented, because in the end of the day what the user needs to know is all the metrics we have (and, again, these are documented). Some of these metrics just access these outcomes directly and that's how we expect the user to have access to them, and not by looking at `rs.outcomes[:relative_coral_cover]` direclty, for example. 

Let me know what do you think. :)

arlowhite: Since someone adding a future resultset metric will probably be looking at this function anyway, then it's probably obvious.

arlowhite (commented):
looks good to me, though not sure I have enough expertise in Julia to Approve it.

renaming `X` is a bit of pain, just a suggestion if you feel like doing more renaming in this PR.

src/metrics/metrics.jl
```diff
@@ -645,7 +687,7 @@ function _relative_shelter_volume(rs::ResultSet)::YAXArray
     return rs.outcomes[:relative_shelter_volume]
 end
 relative_shelter_volume = Metric(
-    _relative_shelter_volume, (:timesteps, :locations, :scenarios)
+    _relative_shelter_volume, (:timesteps, :locations, :scenarios), "m³"
```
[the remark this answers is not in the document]
Zapiano: Good idea, I'll update it. Tks

test/metrics/metrics.jl
Zapiano: few comments about this approach: 
1. there are some metrics that I consider to be "absolute" that don't have the word "absolute" in the name, like `juvenile_indicator` or `coral_evenness`. But we could manually add those to the vector of metrics.
2. The setup for each metric is slightly different. For the functions that accept just a ResultSet as input, that's ok. But for the ones that I need to pass a list of arguments, I would actually need to do this setup one by one. 
3. These are just an initial set of tests, but we might want to write more specific tests for each metric (like the `relative_shelter_volume` that has a test scenario only with coral with the maximum shelter density).

test/viz.jl
Zapiano: I'm not sure I understand what actions do you want me to take here. I can move this to `result_set.jl` (I can also do that and rename `result_set.jl`, if you have any suggestions). 

I just really think we shouldn't have them at `runtests.jl` because that makes it hard to run just the tests you are interested on when developing.

ConnectedSystems: You moved it from where it was to this `viz.jl` file. 

I don't mind moving it, but you can find a more appropriate place for it.

test/metrics/metrics.jl
ConnectedSystems: > About item 2: I see in your code snippet you wrote params = ...  # infer appropriate mock data from metric metadata?. I guess my question to that is: how?

We can give as much detail in the metadata we need to make this all work seamlessly. Marking each metric as to absolute or relative nature could also address the need to infer it from their name alone, which is a more robust approach too...

I'm throwing out ideas, that's all.

ConnectedSystems: I should clarify that the auto-generation of tests doesn't necessarily need to be in this PR.

src/utils/text_display.jl
Zapiano: 1. I just updated this code to avoid the duplicate implementation that Takuya was mentioning
2. With the `where {T<:Symbol, N<:Integer}` if you have multiple inputs you can assure that they have the same element type. Also, `T` and `N` are available to you inside the scope of your function in case you need to variables with the same type/element type as T or N.

src/metrics/metrics.jl
```diff
@@ -88,123 +94,133 @@ total_absolute_cover = Metric(
...
-                "Number of groups must divide n_group_and_size. n_group_and_size $(n_group_and_size), n_groups: $(n_groups)"
-            )
-        )
-    end
-    n_sc::Int64 = Int64(n_group_and_size / n_groups)
+)::AbstractArray{T,3} where {T<:Real}
+    n_timesteps, n_group_sizes, n_locs = size(X)
+    _n_sizes::Int64 = n_sizes(n_groups, n_group_sizes)
 
     taxa_cover::YAXArray = ZeroDataCube(
```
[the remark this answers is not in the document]
Zapiano: Maybe in the future, but since the metric is currently called `relative_loc_taxa_cover`, we would need to change the metric name as well - which would be good, but maybe not in this PR

src/metrics/metrics.jl
```diff
@@ -46,20 +49,23 @@ end
...
+    result::YAXArray = dropdims(sum(X; dims=2); dims=2)
+    return result
 end
-function _relative_cover(rs::ResultSet)::AbstractArray{<:Real}
+function _relative_cover(rs::ResultSet)::YAXArray{<:Real}
     return rs.outcomes[:relative_cover]
 end
 relative_cover = Metric(_relative_cover, (:timesteps, :locations, :scenarios))
 
+#TODO
```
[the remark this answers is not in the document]
Zapiano: 🤷‍♂️

src/metrics/metrics.jl
```diff
@@ -88,123 +94,133 @@ total_absolute_cover = Metric(
...
+n_locations)
+- `k_area` : The coral habitable area.
+- `n_groups` : Number of function coral groups.
+
+# Returns
+Coral cover, grouped by taxa for the given scenario, for each timestep and location,
+relative to location k area.
 """
 function _relative_loc_taxa_cover(
     X::AbstractArray{T}, k_area::Vector{T}, n_groups::Int64
```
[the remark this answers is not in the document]
Zapiano: I agree with you that `X` is not a good name. But since I think we need to have a deeper discussion about this `metrics` module and `Metric` struct, and probably this will lead to some changes, I propose we change this in a separate PR.

src/ecosystem/corals/Corals.jl
```diff
@@ -495,3 +495,19 @@ function to_coral_spec(inputs::DataFrameRow)::DataFrame
...
+                "n_group_sizes: $(n_group_sizes), n_groups: $(n_groups)"
+            )
+        )
+    end
+    return Int64(n_group_sizes / n_groups)
+end
+
+function group_indices(n_sizes::Int64, n_group_sizes::Int64)::Vector{UnitRange{Int64}}
+    return [i:(i + (n_sizes - 1)) for i in 1:n_sizes:n_group_sizes]
+end
```
ConnectedSystems, line 513: These deserve docstrings

src/io/result_io.jl
```diff
@@ -383,13 +383,13 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
     n_scenarios = nrow(scen_spec)
 
     # Set up stores for each metric
-    function dim_lengths(metric_structure)
+    function dim_lengths(metric_structure::NTuple{3,Symbol})
         dl = []
-        for d in metric_structure
+        for d in string.(metric_structure)
```
ConnectedSystems, line 388: Isn't it better to avoid casting to string and instead update the comparisons in the if/elseif chain to compare against Symbols?

e.g., `d == :timesteps`

src/io/result_io.jl
```diff
@@ -657,43 +654,46 @@ function load_results(result_loc::String)::ResultSet
     )
 
     outcomes = Dict{Symbol,YAXArray}()
+    outcomes_properties = [:metric_name, :metric_unit, :axes_names, :axes_units]
```
ConnectedSystems, line 657: Grammatically correct name:

```suggestion
    outcome_properties = [:metric_name, :metric_unit, :axes_names, :axes_units]
```

"properties" signifies the plural

src/io/result_io.jl
```diff
@@ -657,43 +654,46 @@ function load_results(result_loc::String)::ResultSet
...
-                    push!(st, res.attrs["unique_loc_ids"])
-                else
-                    push!(st, 1:sz[i])
-                end
+        data = zopen(sd; fill_as_missing=false)
+        data_size = size(data)
+
+        # Construct dimension names and metadata
+        dims = []
+        for (idx, dim) in enumerate(data.attrs["structure"])
```
ConnectedSystems, line 665: I'm okay leaving this as is as it probably does not matter in this specific case, just be aware you're overwriting methods here in this scope (I'm 90% sure `dims` and `dim` are methods).

src/io/result_io.jl
```diff
@@ -400,54 +400,51 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
     ]
 
-    _unique_loc_ids = unique_loc_ids(domain)
-
-    dim_struct = Dict(
-        :structure => string.((:timesteps, :locations, :scenarios)),
-        :unique_loc_ids => _unique_loc_ids
+    metric_symbols::Vector{Symbol} = metrics.to_symbol.(outcome_metrics)
+    metric_names::Vector{String} = metrics.to_string.(outcome_metrics; is_titlecase=true)
+    metric_units::Vector{String} = [metric.unit for metric in outcome_metrics]
```
ConnectedSystems, line 415: Don't feel like you need to accept this, but you can use the same broadcasting sugar as the above two lines here too.

```suggestion
    metric_units::Vector{String} = getfield.(outcome_metrics, :unit)
```

src/io/result_io.jl
```diff
@@ -400,54 +400,51 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
+    axes_names::Vector{NTuple{3,Symbol}} = fill(
+        (:timesteps, :locations, :scenarios), length(outcome_metrics) - 1
     )
-    result_dims::Tuple{Int64,Int64,Int64} = dim_lengths(dim_struct[:structure])
+    push!(axes_names, (:timesteps, :species, :scenarios))
```
ConnectedSystems, line 419: Add a comment explaining why this last one is slightly different (I can infer it, but I know the context here, others likely won't)

src/io/result_io.jl
```diff
@@ -400,54 +400,51 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
-    _unique_loc_ids = unique_loc_ids(domain)
-
-    dim_struct = Dict(
-        :structure => string.((:timesteps, :locations, :scenarios)),
-        :unique_loc_ids => _unique_loc_ids
+    metric_symbols::Vector{Symbol} = metrics.to_symbol.(outcome_metrics)
+    metric_names::Vector{String} = metrics.to_string.(outcome_metrics; is_titlecase=true)
+    metric_units::Vector{String} = [metric.unit for metric in outcome_metrics]
+    axes_names::Vector{NTuple{3,Symbol}} = fill(
+        (:timesteps, :locations, :scenarios), length(outcome_metrics) - 1
```
ConnectedSystems, line 417: I don't love this implied mixed used of Vectors and Tuples, it complicates handling later on by having to have method definitions to handle both types, doesn't it?

Unless I'm wrong and it's very straight forward, or it has a very narrow use case, I keeping everything in the same data type.

src/io/result_io.jl
```diff
@@ -400,54 +400,51 @@ function setup_result_store!(domain::Domain, scen_spec::DataFrame)::Tuple
...
             fill_value=nothing,
             fill_as_missing=false,
             path=joinpath(z_store.folder, RESULTS, string(m_name)),
-            chunks=(result_dims[1:(end - 1)]..., 1),
-            attrs=dim_struct,
+            chunks=(result_dims[idx][1:(end - 1)]..., 1),
+            attrs=outcomes_attrs[idx],
             compressor=COMPRESSOR
-        ) for m_name in met_names
+        ) for (idx, m_name) in enumerate(metric_symbols)
```
ConnectedSystems, line 445: This is a nice clean up, thank you.

src/metrics/scenario.jl
```diff
@@ -60,19 +64,38 @@ end
 scenario_relative_cover = Metric(_scenario_relative_cover, (:timesteps, :scenarios))
 
 """
-    scenario_relative_juveniles(data::YAXArray, coral_spec::DataFrame, k_area::AbstractVector{<:Real}; kwargs...)::AbstractArray{<:Real}
+    scenario_relative_juveniles(data::YAXArray{<:Real,3}, coral_spec::DataFrame, k_area::AbstractVector{<:Real}; kwargs...)::AbstractArray{<:Real}
```
ConnectedSystems, line 67: You've renamed `data` to `X` in the implementation but it is left as is here.

src/metrics/scenario.jl
```diff
@@ -60,19 +64,38 @@ end
...
     scenario_relative_juveniles(rs::ResultSet; kwargs...)::YAXArray
 
 Calculate the mean relative juvenile population for each scenario for the entire domain.
+
+# Arguments
+- `X` : Raw data for a single scenario.
+- `coral_spec` : Coral spec DataFrame.
+- `k_area` : K_area.
+- `rs` : Resultset.
```
ConnectedSystems, line 76: I would put `rs` after `X` here, otherwise it gives the impression it comes after the other arguments.

By the way, these docstrings for `_metric_name` methods are generally temporary. They don't get picked up when querying `?metric_name` as the docstring is associated with the "wrong" method. There are ways to address this, but needs some investigation.

src/metrics/utils.jl
```diff
@@ -70,17 +80,40 @@ Convenience method that slices the data in the specified manner.
 - `metric` : Function, the metric function to apply to "raw" data.
 - `data` : YAXArray, data to pass into `metric`
 - `args` : Additional positional arguments to pass into `metric`
-- `dims` : dummy keyword argument, not used but defined to allow use with other methods
+- `kwargs` : Additional keyword arguments to pass into `slice_results`
+    - `dims` : dummy keyword argument, not used but defined to allow use with other methods
 """
 function call_metric(metric::Union{Function,Metric}, data::YAXArray, args...; kwargs...)
-    dims = haskey(kwargs, :dims) ? kwargs[:dims] : nothing
+    dims = get(kwargs, :dims, nothing)
```
ConnectedSystems, line 87: Yes, this is better!

src/metrics/utils.jl
```diff
@@ -70,17 +80,40 @@ Convenience method that slices the data in the specified manner.
...
         return metric(slice_results(data; kwargs...), args...)
     else
         return metric(slice_results(data; kwargs...), args...; dims=dims)
     end
 end
 
+"""
+    fill_axes_properties(datacube::YAXArray)::YAXArray
+
+Fill :axes_names and :axes_units properties of datacube
```
ConnectedSystems, line 98:
```suggestion
Fill `:axes_names` and `:axes_units` properties of the datacube.
```

src/metrics/utils.jl
```diff
@@ -70,17 +80,40 @@ Convenience method that slices the data in the specified manner.
...
+
+# Arguments
+- `datacube` : YAXArray datacube
+"""
+function fill_axes_properties(metric::Metric, metric_result::YAXArray)::YAXArray
```
ConnectedSystems, line 103: Argument list in docstring does not match implementation?

test/metrics/metrics.jl
ConnectedSystems: Lets address this later

test/mock_data/coral_cover.jl
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    # For each timestep and scenario the total cover of each location sums up to 1
+    cover_data::Array{Float64,4} = (cover_weights ./ sum(cover_weights; dims=2))
+
+    # This is to prevent the cover to grow above 1 due to floating point precision error
+    cover_data[:, end, :, :] .=
+        1 .- dropdims(
+            sum(cover_data[:, 1:(end - 1), :, :]; dims=2); dims=2
+        )
```
ConnectedSystems, line 43: Actually, could this be addressed [with `clamp!()`](https://docs.julialang.org/en/v1/base/math/#Base.Math.clamp!) ?

test/result_set.jl
```diff
@@ -0,0 +1,58 @@
...
+    dom = ADRIA.load_domain(TEST_DOMAIN_PATH)
+
+    # Set min_iv_locations upper to 10 as this is the number of locations used in tests
+    dom = ADRIA.set_factor_bounds(dom, :min_iv_locations, (5.0, 10.0))
+
+    # Create some scenarios
+    # The number of scenarios set here seem to be the rough minimum for SIRUS to produce
+    # some results.
+    n_samples = 64
+    scens = ADRIA.sample(dom, n_samples)
```
ConnectedSystems, line 51: Any plans to tests/handling cases where the sample size is too small?

test/utils/text_display.jl
```diff
@@ -1,7 +1,36 @@
+using ADRIA: human_readable_name, get_scientific_factors
+
 @testset "text_display" begin
-    @testset "get_scientific_factors" begin
-        get_scientific_factors = ADRIA.get_scientific_factors
+    @testset "human_readable_name" begin
+        @testset "Symbols inputs" begin
```
ConnectedSystems, line 5: Remove double plural

```suggestion
        @testset "Symbol inputs" begin
```

ConnectedSystems (changes requested):
A bunch of minor comments to consider

test/viz.jl
Zapiano: I've changed the name of this file to `analysis.jl`. I don't love this name because it might look like I'm testing only the analysis module. An alternative would be to call it `end_to_end` (which is a weird name, but that's what this test is, like you said). Do you have other suggestions? @arlowhite do you have any alternative suggestions?

src/ecosystem/corals/Corals.jl
Zapiano: Done.

src/io/result_io.jl
Zapiano: Done.

src/io/result_io.jl
Zapiano: Done

src/io/result_io.jl
Zapiano: Done

src/io/result_io.jl
Zapiano: Done

src/metrics/scenario.jl
Zapiano: Done

src/metrics/scenario.jl
Zapiano: Done

src/metrics/utils.jl
Zapiano: Done

test/mock_data/coral_cover.jl
Zapiano: I don't think so, because it's not a vector with values that should be 1 but are 0.9998 or 1.0002, but a vector with values that should sum up to 1 (with respect to one of the dimensions) but sometimes sum up to 0.9998 or 1.0002. That's why I'm taking the last entry of each col (dimension 2) and replacing by 1 minus the sum of the other elements in each col.

test/result_set.jl
Zapiano: Do you think we need to include this as part of this PR? I like the idea of changing this function to be able to pass the number of samples as argument (and maybe other things) but do you think this is part of this task or should I open an issue to address that in the future?

Zapiano:
@ConnectedSystems  I think I've addressed almost all your comments. The only things left are:

1. About metric tests structure, which I understand you said we can tackle later
2. About `absolute` versus `relative` metric tests. Like I said, not always we can infer that information from the metric name. We could add that info to the metric metadata, but the problem is that in order to do that we need to change all metric methods to "manually" add that info (at least that's the only way I can think of). To do this automatically we would need to add this info to the metric struct attributes, and I think we should have a discussion about if we want to change the current metrics struct implementation before we do that.

test/mock_data/coral_cover.jl
ConnectedSystems: Yeah, makes sense

test/result_set.jl
ConnectedSystems: Sure

ConnectedSystems:
About point 2, is it fair to say anything with an SI unit would be absolute, and anything non-dimensional or as a percentage is relative?

Zapiano:
> About point 2, is it fair to say anything with an SI unit would be absolute, and anything non-dimensional or as a percentage is relative?

One counter example I know is count of things. Anything that is a count will be absolute, not relative, but will be non-dimensional.

Correct me if I'm wrong, but I think the three metrics below wouldn't follow this rule:

`coral_evenness` is a non-dimensional absolute metric
`scenario_absolute_juveniles` is a non-dimensional absolute metric
`scenario_evenness` seems to be a non-dimensional absolute metric

ConnectedSystems:
I'm not sure I would consider `coral_evenness` an absolute metric, but I take your point. I wouldn't call it relative either...

I agree `absolute_juveniles` is absolute, but is it non-dimensional? Shouldn't it be area taken up by juveniles? Or am I confusing myself?

Zapiano:
> I'm not sure I would consider `coral_evenness` an absolute metric, but I take your point. I wouldn't call it relative either...

Maybe that's not the best approach but I'm calling `absolute` everything that is not `relative`. 
 
> I agree `absolute_juveniles` is absolute, but is it non-dimensional? Shouldn't it be area taken up by juveniles? Or am I confusing myself?

Currently `scenario_absolute_juveniles` doesn't have a `unit` set, but that might be just a mistake. I just assumed that was intentional. I agree it should be `m^2`, since it's the average over `absolute_juveniles`, which has dimension `m^2`.

Zapiano:

test/utils/text_display.jl
```diff
@@ -1,7 +1,36 @@
+using ADRIA: human_readable_name, get_scientific_factors
+
 @testset "text_display" begin
-    @testset "get_scientific_factors" begin
-        get_scientific_factors = ADRIA.get_scientific_factors
+    @testset "human_readable_name" begin
```
ConnectedSystems, line 4: Test sets should be names (e.g., not snake case)

But leaving as is as it's very minor.

ConnectedSystems (approved):
Approving, but noting a "nice to have" is to use unitful values in the metadata (or make the `const` units unitful values)

https://painterqubits.github.io/Unitful.jl/stable/

PR #850, opened 2024-09-16
Update interface to use correct gpkg

### LINKED ISSUE #849: mismatch of naming in Cscape results set
@DanTanAtAims 

When creating the C~scape Results set the carrying capacity is named "location_max_coral_cover"
https://github.com/open-AIMS/ADRIA.jl/blob/cea9888dca09c9b321be6502e4adef5127182fb1/src/io/cscape_result_io.jl#L145

However, in another function that attempts to access the fields it is specified as "loc_max_coral_cover" 
https://github.com/open-AIMS/ADRIA.jl/blob/cea9888dca09c9b321be6502e4adef5127182fb1/src/io/ResultSet.jl#L390

I believe the second one is the correct naming as per other code implementations

DanTanAtAims:
1. Correct the calculation of maximum available coral cover in result loading
2. Remove the manual site data correction (new dpkg available on teams in ADRIA result sets, `cscape_dpkg_2024_09_16.zip`)
3. Use `loc` instead of `site` to maintain compatability with existing `ResultSet`

closes #849

### DIFF
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -125,7 +125,7 @@ function load_results(
         timeframe
     )
 
-    location_max_coral_cover = 1 .- geodata.k ./ 100
+    location_max_coral_cover = geodata.k ./ 100
     location_centroids = [centroid(multipoly) for multipoly ∈ geodata.geom]
 
     outcomes = Dict{Symbol,YAXArray}()
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -78,8 +78,6 @@ function load_results(
     gpkg_path = _get_gpkg_path(data_dir)
     geodata = GDF.read(gpkg_path)
 
-    geodata = _manual_site_additions(geodata, location_ids, raw_set)
-
     connectivity_path = joinpath(data_dir, "connectivity/connectivity.csv")
     connectivity = CSV.read(connectivity_path, DataFrame; comment="#", header=true)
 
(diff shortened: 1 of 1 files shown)


src/io/cscape_result_io.jl
```diff
@@ -778,7 +732,8 @@ function _cscape_relative_cover(datasets::Vector{Dataset})::YAXArray
     # all(diff(cumsum(n_scens)) .== 1)
 
     start_end_pos = _data_position(n_scens)
-    @showprogress desc="Loading datasets" for (ds_idx, dataset) in zip(start_end_pos, datasets)
+    @showprogress desc = "Loading datasets" for (ds_idx, dataset) in
+                                                zip(start_end_pos, datasets)
```
ConnectedSystems, line 736: @arlowhite Any way to configure formatter to ignore this case? I'd prefer argument assignment style over variable assignment style in this case (i.e., we want `x=100` here, not `x = 100`)

VHallerBull (approved):
In line 128 the name is location_max_coral_cover versus in line 14 loc_max_coral_cover similar with loc_centroids
Is this intentional?

src/io/cscape_result_io.jl
```diff
@@ -9,13 +9,13 @@ struct CScapeResultSet <: ResultSet
     name::String
     RCP::String
 
-    site_ids
-    site_area::Vector{Float64}
-    site_max_coral_cover::Vector{Float64}
-    site_centroids
+    loc_ids
+    loc_area::Vector{Float64}
+    loc_max_coral_cover::Vector{Float64}
```
DanTanAtAims, line 14: `location_max_coral_cover` in line 128 gets assigned to the field `loc_max_coral_cover` in the results set structure. The name`location_max_coral_cover` is the local variable name and doesn't effect the name used to access it from the result structure. 

```julia
# Load results
rs = ADRIA.load_results(CScapeResultSet, "<path>")

# Get max_coral_cover for each location
data = rs.loc_max_coral_cover

# Error
data = rs.location_max_coral_cover
```

VHallerBull: It all looks good to me then

VHallerBull (approved):

src/io/cscape_result_io.jl
arlowhite: Maybe if you want no spaces around all kwargs `=`?
https://domluna.github.io/JuliaFormatter.jl/stable/#whitespace_in_kwargs

ConnectedSystems: That might be it!

PR #853, opened 2024-09-17
correct calculation of relative cover

DanTanAtAims:
Corrections to relative cover calculations.

1. Divide by area of each location (summing intervened and non-intervened area)
2. Divide by 100 to convert percentage to proportion.

### DIFF
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -691,9 +691,9 @@ function _cscape_relative_cover(dataset::Dataset)::Array
         _drop_sum(
             dataset.cover[intervened=1][dummy_selector...], dim_sum
         ) .* reshape(
-            area[1, :] .* dataset.k[:],
+            area[1, :],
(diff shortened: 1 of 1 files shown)


VHallerBull (commented):
I think there are a few issues here:
1. C~scape has cover relative to total area of the reef so in the first step, you calculate absolute cover which should be `absolute_cover=cover_cscape*area` . You would need to divide this by 100 if cover_cscape is in the range 0-100 (check that).
2. Add together absolute_cover for intervened and non-intervened area
3. Convert `absolute_cover` back to `relative_cover` as this is relative to habitable area in ADRIA, you would use `absolute_cover/(area*k)` assuming k is in the range 0-1 otherwise `absolute_cover/(area*k/100)`

DanTanAtAims:
## Relative cover plot with 3 scenarios

image

VHallerBull (approved):

PR #854, opened 2024-09-18
Speed up CScape Result Interface

DanTanAtAims:
Use the [NetCDF.jl](https://github.com/JuliaGeo/NetCDF.jl) api instead of [YAXArrays](https://github.com/JuliaDataCubes/YAXArrays.jl) to load netcdf data.

Loading 22 files now takes around 1 second.

image

Loading a variable from 22 files takes around 1 second as well.

image

It would be good to see if this scales well to 2000 files.

## Relative cover of random files downloaded

image

### DIFF
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -280,7 +280,8 @@ function _create_inputs_dataframe(
     input_index::Int
 )::DataFrame
     functional_types::Vector{String} = _get_functional_types(nc_handle)
-    n_draws::Int = "draws" in keys(nc_handle.dim) ? length(get(nc_handle.dim, "draws", [1])) : 1
+    n_draws::Int =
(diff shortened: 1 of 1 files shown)


VHallerBull (approved):

src/io/cscape_result_io.jl
```diff
@@ -160,7 +159,7 @@ end
 
 Get the dimension names from a netcdf variable
 """
-function _ncvar_dim_names(::Type{String}, ncvar::NcVar)::Vector{Symbol}
+function _ncvar_dim_names(ncvar::NcVar)::Vector{Symbol}
     return [Symbol(d.name) for d in ncvar.dim]
```
ConnectedSystems, line 163: FYI, this could be:

```julia
Symbol.(getfield.(ncvar.dim, :name))
```

PR #860, opened 2024-09-23
add scenario id to scenario dataframe

DanTanAtAims:
Stores Scenario Id in scenario dataframe

### DIFF
--- a/src/io/cscape_result_io.jl
+++ b/src/io/cscape_result_io.jl
@@ -283,9 +283,9 @@ function _create_inputs_dataframe(
     n_draws::Int =
         "draws" in keys(nc_handle.dim) ? length(get(nc_handle.dim, "draws", [1])) : 1
 
-    dhws::Vector{Float64} = Float64.(fill(input_index, (n_draws,)))
-    cyclones::Vector{Float64} = Float64.(fill(input_index, (n_draws,)))
-    scenario_id::Vector{Int64} = Int64.(fill(scenario_spec.ID, (n_draws,)))
+    dhws::Vector{Float64} = Float64.(fill(input_index, n_draws))
+    cyclones::Vector{Float64} = Float64.(fill(input_index, n_draws))
+    scenario_id::Vector{Int64} = Int64.(fill(scenario_spec.ID, n_draws))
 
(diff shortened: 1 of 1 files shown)


VHallerBull (approved):
This looks good. We will have to revisit the upper limit down the road. It is ok for this set of results, but future sets will have Scenario IDs that are significantly larger (1e+9 would be a more appropriate upper limit).

src/io/cscape_result_io.jl
```diff
@@ -285,6 +285,7 @@ function _create_inputs_dataframe(
 
     dhws::Vector{Float64} = Float64.(repeat([input_index], n_draws))
     cyclones::Vector{Float64} = Float64.(repeat([input_index], n_draws))
+    scenario_id::Vector{Int64} = Int64.(repeat([scenario_spec.ID], n_draws))
```
ConnectedSystems, line 288: @DanTanAtAims `fill()` is better, it doesn't allocate

VHallerBull (approved):

src/io/cscape_result_io.jl
```diff
@@ -342,25 +343,26 @@ function _create_inputs_dataframe(
...
+        heritability1=fill(heritability1, (n_draws,)),
+        heritability2=fill(heritability2, (n_draws,)),
+        plasticity=fill(plasticity, (n_draws,)),
+        intervention_start=fill(intervention_start, (n_draws,)),
+        intervention_duration=fill(duration, (n_draws,)),
+        intervention_frequency=fill(frequency, (n_draws,)),
+        deployment_area=fill(deployment_area, (n_draws,)),
+        n_deployment_locations=fill(n_dep_locations, (n_draws,)),
+        enhancement_mean=fill(enhancement_mean, (n_draws,)),
+        enhancement_std=fill(enhancement_std, (n_draws,)),
```
ConnectedSystems, line 365: Don't think you need the implicit tuple, you can just do:

`fill(value, number)`

ConnectedSystems: Incidentally, the current way, I think you're allocating to create the one element tuple

PR #862, opened 2024-09-25
Improve summarize performance by using mapslices

### LINKED ISSUE #788: Improve loc_trajectory() function performance
Currently the `ADRIA.loc_trajectory()` function takes a long amount of time to run, particularly for GBR wide analyses. 
Possible speed up method is to use `mapslices()` function and specify the dimension to be reduced (`:scenarios`).

Zapiano:
The previous approach, using JuliannedArrays.Slices, is taking too much to run for some unknown reason. Using `mapslices` seems to fix the problem.

closes #788

### DIFF
--- a/src/metrics/site_level.jl
+++ b/src/metrics/site_level.jl
@@ -86,18 +86,25 @@ YAXArray with summary metric for the remaining axis.
 function summarize(
     data::YAXArray{D,T,N,A}, alongs_axis::Vector{Symbol}, metric::Function
 )::YAXArray where {D,T,N,A}
-    # Check if there's enough space to read(data)
+    # The approach using JuliennedArrays is faster then directly passing the YAXArray to
+    # mapslices only when we `read(data)`. Since `read(data)` loads all the data from disk
+    # into memory, we only want to use that approach when there is a reasonable amount of
+    # available space in RAM
     data_size = Base.summarysize(data)
     free_ram = Sys.free_memory()
-    if data_size > free_ram
+    proportional_usage = data_size / free_ram
+
+    # Only use this approach when data occupies more than 70% of available space in RAM
+    if proportional_usage > 0.7 * free_ram
+        # `D.` is ensuring the returned YAXArray has the same type as the input `data`
         return D.(mapslices(metric, data; dims=alongs_axis))
     end
 
     alongs = sort([axis_index(data, axis) for axis in alongs_axis])
 
     # Use of JuliennedArrays is in an attempt to speed up calculation of summary statistics.
-    #   We see a small but still worthwhile improvement in practice.
-    #   see: https://stackoverflow.com/a/62040897
+    # We see a small but still worthwhile improvement in practice.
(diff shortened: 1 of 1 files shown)


BG-AIMS (commented):
I think these are good changes.

Zapiano:
@ConnectedSystems @BG-AIMS I've changed the approach here, going back to JuliennedArrays, and now it's much faster than with `mapslices` (as mentioned here: https://github.com/open-AIMS/ADRIA.jl/issues/788#issuecomment-2373154482 )

Zapiano:
Just to give an idea:

Using `mapslices`:
image

Using JuliennedArrays with `read()`:
image

Zapiano:
@ConnectedSystems I still have to test this, but I tried dynamically determining if there's enough space to `read(data)`. Let me know if that's what you were thinking.

Zapiano:
Quick comment: I changed the threshold to be 70% of available space in RAM, to have a safety margin to account for memory spikes and other unpredicted memory usages that could cause a problem.

src/metrics/site_level.jl
```diff
@@ -86,12 +86,26 @@ YAXArray with summary metric for the remaining axis.
...
+    # available space in RAM
+    data_size = Base.summarysize(data)
+    free_ram = Sys.free_memory()
+    proportional_usage = data_size / free_ram
+
+    # Only use this approach when data occupies more than 70% of available space in RAM
+    if proportional_usage > 0.7 * free_ram
+        # `D.` is ensuring the returned YAXArray has the same type as the input `data`
+        return D.(mapslices(metric, data; dims=alongs_axis))
+    end
```
ConnectedSystems, line 101: Suggesting a more readable if statement

```suggestion
    data_size = Base.summarysize(data)
    
    # Only use this approach when data occupies more than 70% of available space in RAM
    available_threshold = Sys.free_memory() * 0.7
    proportional_usage = data_size / free_ram
    if proportional_usage > available_threshold 
        # `D.` is ensuring the returned YAXArray has the same type as the input `data`
        return D.(mapslices(metric, data; dims=alongs_axis))
    end
```

src/metrics/site_level.jl
```diff
@@ -86,18 +86,25 @@ YAXArray with summary metric for the remaining axis.
...
+    # The approach using JuliennedArrays is faster then directly passing the YAXArray to
+    # mapslices only when we `read(data)`. Since `read(data)` loads all the data from disk
+    # into memory, we only want to use that approach when there is a reasonable amount of
+    # available space in RAM
     data_size = Base.summarysize(data)
     free_ram = Sys.free_memory()
-    if data_size > free_ram
+    proportional_usage = data_size / free_ram
+
+    # Only use this approach when data occupies more than 70% of available space in RAM
```
ConnectedSystems, line 97:
```suggestion
    # Only use this approach when data occupies more than 70% of available RAM
```

For conciseness 😊

ConnectedSystems (approved):
Thanks @Zapiano

PR #864, opened 2024-09-30
Pin NetCDF.jl and Zarr.jl in order to pin DiskArrays.jl

Zapiano:
`NetCDF.jl` and `Zarr.jl` had to be pinned in order to pin `DiskArrays` to `0.3.23`. Once this PR https://github.com/meggart/DiskArrays.jl/pull/181 is merged we should unpin these two.


Zapiano:
@ConnectedSystems sorry, I think there's a problem with some tests, I'll let you know when this is ready for review.

Zapiano:
> @ConnectedSystems sorry, I think there's a problem with some tests, I'll let you know when this is ready for review.

Just confirming that the tests are passing here and this PR is ready for review.

ConnectedSystems (approved):
DiskArrays.jl is not explicitly pinned, but I'm guessing this is fine given the Zarr and NetCDF dependency at their pinned versions require the version we want/need.

On this basis I am approving.

Note: `update` needs to be run in Pkg for changes to be applied

PR #878, opened 2024-10-09
Apply CompatHelper suggestions, free NetCDF,Zarr; add & pin DiskArrays

arlowhite:
Apply the suggestions from CompatHelper.

add & pin DiskArrays instead of pinning packages that depend on it. (free NetCDF, Zarr)

Note NetCDF 0.12 requires later version of DiskArrays, so we're stuck on 0.11

`] update`

This results in DimensionalData updating from 0.27.9 to 0.28.2
https://github.com/rafaqz/DimensionalData.jl/releases

Note 0.28.0
> Breaking: Broadcast performance and correctness (https://github.com/rafaqz/DimensionalData.jl/pull/777)


Project.toml
```diff
@@ -87,7 +87,8 @@ CpuId = "0.3"
 DataFrames = "1"
 DataStructures = "0.18, 0.2"
 Dates = "1"
-DimensionalData = "0.25, 0.26, 0.27"
+DimensionalData = "0.25, 0.26, 0.27, 0.28"
+DiskArrays = "=0.3.23"
```
arlowhite, line 91: this may be redundant with pin? but just to be safe

Project.toml
```diff
@@ -115,6 +116,7 @@ MAT = "0.10"
 MLJ = "0.19.2, 0.20"
 Makie = "0.20.8, 0.21"
 ModelParameters = "0.4"
+NetCDF = "0.11, 0.12"
```
arlowhite, line 119: 0.11 needed to resolve a compatible version with DiskArrays@0.3.23

ConnectedSystems (approved):
Cheers

PR #903, opened 2024-12-25
Fix DHW scenario viz again

ConnectedSystems:
Trying to load RME data and found this issue has cropped up again.

Before:

image

After:

image

Before merging I will check the latest site-level Domains work as expected.

### DIFF
--- a/ext/AvizExt/viz/environment/cyclones.jl
+++ b/ext/AvizExt/viz/environment/cyclones.jl
@@ -1,7 +1,8 @@
 using Statistics
 
 function ADRIA.viz.cyclone_scenario(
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
Hi @DanTanAtAims when you get a chance, could you check that the above issue is resolved when plotting both Moore and RME datasets?

```julia
using GLMakie, GeoMakie, GraphMakie
using ADRIA


dom = ADRIA.load_domain("path", "45")
ADRIA.viz.dhw_scenario(dom, 1)
```

ConnectedSystems:
Ah, I just noticed that the last two datasets have the year tick labels missing. I think I saw this issue before, but not something we can fix here.

I suspect it's the domain being loaded with the year axis numbered by time step (1, 2, 3, etc) not year (1980, 1981, etc)

Issue #905, opened 2025-01-14
Seeding Intervention Habitable Area limit


DanTanAtAims:
Excess number of coral deployments will cause an assertion error in ADRIA when the area of corals deployed exceed the available habitable area. 

## Simple Fix
Cap the coral cover added by seeding to not exceed habitable area. This still leads to unrealistic scenarios where seeded corals are packed perfectly covering all available area. 

## Deployment density limit Fix
Add a limit to the density seeded corals can be deployed at.

## Proportion of Habitable Area Fix
Assume that coral deployments do not cover all habitable area. For example, coral seeding will only cover at most 80% of habitable area.

ConnectedSystems:
For the record, we went with the simple fix.

PR #906, opened 2025-01-14
Incorporate recruitment to growth constraints and cap seeding to habitable area

DanTanAtAims:
Relates to issue #905.

Quick fix to prevent crashes.

### DIFF
--- a/src/interventions/seeding.jl
+++ b/src/interventions/seeding.jl
@@ -19,17 +19,22 @@ function distribute_seeded_corals(
     available_space::Vector{Float64},
     seeded_area::YAXArray
 )::YAXArray
+    total_seeded_area::Float64 = sum(seeded_area)
+    total_available_space::Float64 = sum(available_space)
 
     # Proportion of available space on each site relative to available space at these
     # locations
-    prop_area_avail = available_space ./ sum(available_space)
+    prop_area_avail = available_space ./ total_available_space
+    if total_seeded_area > total_available_space
+        @warn "Seeded area exceeds available space. Restricting to available space."
+        seeded_area .*= total_available_space / total_seeded_area
+    end
 
     # Distribute seeded corals (as area) across locations according to available space
     # proportions:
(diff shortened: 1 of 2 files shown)


src/interventions/seeding.jl
```diff
@@ -19,17 +19,22 @@ function distribute_seeded_corals(
...
 )::YAXArray
+    total_seeded_area::Float64 = sum(seeded_area)
+    total_available_space::Float64 = sum(available_space)
 
     # Proportion of available space on each site relative to available space at these
     # locations
-    prop_area_avail = available_space ./ sum(available_space)
+    prop_area_avail = available_space ./ total_available_space
+    if total_seeded_area > total_available_space
+        @warn "Seeded area exceeds available space. Restricting to available space."
```
ConnectedSystems, line 29: I think worthwhile to include the available area and the seeded area to give an indication of the magnitude of the problem.

ConnectedSystems (approved):
One minor comment, can be merged after adjusting the warning message

src/interventions/seeding.jl
```diff
@@ -80,8 +85,7 @@ function seed_corals!(
         seeded_area
     )
 
-    # Seed each location and log
-    @views cover[seed_sc, seed_locs] .+= scaled_seed
+    # Log seeded corals
     Yseed[:, seed_locs] .= scaled_seed
```
ConnectedSystems, line 89: Hang on - this means in `seed_corals!()` we are no longer seeding corals?

ConnectedSystems (changes requested):
I think I found a conceptual mismatch here - although I think I agree with the general approach taken.

src/interventions/seeding.jl
ConnectedSystems: Actually nevermind - I'll fix up separately. Thank you.

ConnectedSystems (approved):
Any conceptual mismatches to be aligned in different PR/issue

PR #909, opened 2025-01-14
Remove defunct method `seed_corals!`

ConnectedSystems:
Code flow in `run_model()` is very convoluted so I extracted and refactored the key process for coral deployments so what is happening in the main loop is more obvious.

`seed_corals!()` is now consolidated into the `run_model()` method, allowing some defunct code to be cleaned up.

PR #910, opened 2025-01-14
Switch to logging number of coral deployed

ConnectedSystems:
**Please note: This PR requires #909 to be merged in first.**

Previously we were logging the proportional area increase for each deployment location.

For analyses, I think it makes more sense to assess the number of coral deployed at each location instead, particularly for bump hunting and feature analysis.

@DanTanAtAims sorry to overload you - I requested a review from @Zapiano initially, but realised because this PR relies on #909 , whoever reviews that one would have better context on the changes here.

Please check my math used to derive deployed number of corals with a fine-tooth comb.

### DIFF
--- a/src/interventions/seeding.jl
+++ b/src/interventions/seeding.jl
@@ -13,8 +13,6 @@ Distributes seeded corals according to current available space at each selected 
 - seed_volume : Absolute number of coral to deploy.
 
 # Returns
-YAXArray[taxa to seed ⋅ number of seed locations], area increased relative to k area.
-
 - YAXArray[taxa to seed ⋅ number of seed locations], Proportional increase in cover relative to locations' `k` area
 - Matrix[seed locations ⋅ taxa to seed], Number of coral deployed
 """
--- a/src/interventions/seeding.jl
+++ b/src/interventions/seeding.jl
@@ -23,7 +23,7 @@ function distribute_seeded_corals(
     available_space::Vector{Float64},
     seeded_area::YAXArray,
     seed_volume::Vector{Float64}
-)::Tuple{YAXArray, Matrix{Float64}}
+)::Tuple{YAXArray,Matrix{Float64}}
     total_seeded_area::Float64 = sum(seeded_area)
     total_available_space::Float64 = sum(available_space)
 
--- a/src/interventions/seeding.jl
+++ b/src/interventions/seeding.jl
@@ -45,7 +45,10 @@ function distribute_seeded_corals(
         locations=1:length(available_space)
     )
 
-    n_deployed_coral = prop_area_avail .* seed_volume'
+    n_deployed_coral =
+        prop_area_avail .* seed_volume' .* max(
+            total_available_space / total_seeded_area, 1.0
+        )
 
     return proportional_increase, n_deployed_coral
 end
--- a/src/interventions/seeding.jl
+++ b/src/interventions/seeding.jl
@@ -7,18 +7,23 @@ Calculate proportion of deployed corals to be seeded at each of the selected loc
 Distributes seeded corals according to current available space at each selected site.
 
 # Arguments
-- seed_loc_k_m² : carrying capacity area of locations to seed in m².
-- available_space : currently available space at each seed location in m².
-- seeded_area : area (in m²) of each coral type to be seeded with dim taxa.
+- seed_loc_k_m² : Carrying capacity area of locations to seed in m².
+- available_space : Currently available space at each seed location in m².
+- seeded_area : Area (in m²) of each coral type to be seeded with dim taxa.
+- seed_volume : Absolute number of coral to deploy.
 
 # Returns
 YAXArray[taxa to seed ⋅ number of seed locations], area increased relative to k area.
+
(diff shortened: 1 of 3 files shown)


src/scenario.jl
```diff
@@ -871,15 +872,19 @@ function run_model(
         has_fog_locs::Bool = !isempty(selected_fog_ranks)
 
         # Fog selected locations
-        if fog_decision_years[tstep] && has_fog_locs
+        if has_fog_locs  # fog_decision_years[tstep] &&
```
ConnectedSystems, line 875: Sorry, this got mixed in by accident.

It is a shift in how the scenarios are conceptualized.
Previously, we were changing how often deployments occur (once every n years, where n may be 1).

Now it seems the emphasis is on deployments occurring every year. So this change allows this, but still allows the selection of deployment locations to change every n years.

src/scenario.jl
```diff
@@ -939,25 +944,39 @@ function run_model(
         has_seed_locs::Bool = !isempty(selected_seed_ranks)
 
         # Apply seeding (assumed to occur after spawning)
-        if seed_decision_years[tstep] && has_seed_locs
+        if has_seed_locs  # seed_decision_years[tstep] &&
```
DanTanAtAims, line 947: Either we maintain the seed decision year check
```suggestion
        if has_seed_locs && seed_decision_years[tstep]
```

or the variable `selected_seed_ranks` must be assigned to be empty at the beginning of every timestep, otherwise ADRIA will attempt to carry out seeding when its not a seed year and will use the previous seeding years locations.

DanTanAtAims (changes requested):
The unit tests need to be updated to reflect the change to seeding. I can do this if you want.

src/scenario.jl
```diff
@@ -939,25 +944,39 @@ function run_model(
...
-                a_adapt,
-                @view(Yseed[tstep, :, :]),
-                c_std,
-                c_mean_t
+
+            available_space = leftover_space_m²[seed_locs]
+            seed_locs = seed_locs[findall(available_space .> 0.0)]
+
+            # Calculate proportion to seed based on current available space
+            proportional_increase, abs_seeded_area = distribute_seeded_corals(
```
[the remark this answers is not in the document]
ConnectedSystems: Yes, thanks for catching that.

src/interventions/seeding.jl
```diff
@@ -35,121 +39,49 @@ function distribute_seeded_corals(
...
+    proportional_increase = DataCube(
         scaled_seed;
         taxa=caxes(seeded_area)[1].val.data,
         locations=1:length(available_space)
     )
-end
 
-"""
-    seed_corals!(cover::Matrix{Float64}, loc_k_area::V, leftover_space_m²::V, seed_locs::Vector{Int64}, seeded_area::YAXArray, seed_sc::BitVector, a_adapt::V, Yseed::SubArray, stdev::V, c_dist_t::Matrix)::Nothing where {V<:Vector{Float64}}
+    n_deployed_coral = prop_area_avail .* seed_volume'
```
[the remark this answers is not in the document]
ConnectedSystems: Maybe I move it to the if/else block above? (or add a new one)

```
if total_seeded_area > total_available_space
    # adjusted calculation
else
   # previous calculation
end
```

src/scenario.jl
ConnectedSystems: Yes, that's the idea. See my comment.

Deployments should now occur every year, but only the change in locations vary every n years where n could be 0 (select locations once and keep deploying in the same locations), 1 (change every year), etc.

ConnectedSystems:
> The unit tests need to be updated to reflect the change to seeding. I can do this if you want.

Yes please, thank you Daniel. Apologies for the oversight.

src/interventions/seeding.jl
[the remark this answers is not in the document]
ConnectedSystems: My intention above was to try to avoid repeat calculations of `total_available_space / total_seeded_area` but maybe that's a micro-optimisation...

src/scenario.jl
```diff
@@ -939,28 +944,32 @@ function run_model(
...
         # Apply seeding (assumed to occur after spawning)
-        if seed_decision_years[tstep] && has_seed_locs
+        if has_seed_locs  # seed_decision_years[tstep] &&
             # Seed selected locations
             # Selected locations can fill up over time so avoid locations with no space
             seed_locs = findall(log_location_ranks.locations .∈ [selected_seed_ranks])
-            seed_locs = seed_locs[findall(leftover_space_m²[seed_locs] .> 0.0)]
+
+            available_space = leftover_space_m²[seed_locs]
+            seed_locs = seed_locs[findall(available_space .> 0.0)]
```
DanTanAtAims, line 953: In scenarios where seeding is quite, continuously seeding the same locations leads to them filling up. This causes `seed_locs = []` because `findall(leftover_space_m²[seed_locs] .> 0.0)` will return an empty list.

We should skip seeding when this happens.

src/scenario.jl
ConnectedSystems: Interesting that you get an error, I haven't come across it just yet.

DanTanAtAims (approved):
Tests passing. Looks good to me.

PR #912, opened 2025-01-15
Add option to input functional groups and size classes parameters

Rosejoycrocker:
Adds the option to input the number of functional groups and size classes and associated parameters as an input file when loading the domain. If no file is input, the domain loads the default parameters as normal.

- Changes to `CoralGrowth` so number of functional groups and size classes is not hard coded
- If no file is provided, default parameters and number of functional groups and size classes are loaded as normal
- `coral_spec()` requires an input dictionary with the base parameters required to build the spec, but the default can be loaded without inputs using `ADRIA.load_default_coral_spec()`

@ConnectedSystems, I'm not sure what you think of storing alternative coral parameters in a json. This was used due to being able to load directly into a dict, but open to other formats.

### DIFF
--- a/src/Domain.jl
+++ b/src/Domain.jl
@@ -18,16 +18,17 @@ mutable struct EnvLayer{S<:AbstractString,TF}
 end
 
 """
-    load_domain(path::String)
+    load_domain(path::String; coral_spec_path::String="")
 
 Load ADRIA domain specification from data package.
 No SSP/RCP data is preset.
 
 # Arguments
 - `path` : location of data package.
+- `coral_spec_path` : location of JSON with coral ecological parameters if not using default functional groups and size classes.
 """
-function load_domain(path::String)::Domain
-    return load_domain(path, "")
+function load_domain(path::String; coral_spec_path::String="")::Domain
+    return load_domain(path, ""; coral_spec_path=coral_spec_path)
 end
 
 function unique_loc_ids(d::Domain)::Vector{String}
--- a/src/ExtInterface/ADRIA/Domain.jl
+++ b/src/ExtInterface/ADRIA/Domain.jl
@@ -23,6 +23,7 @@ mutable struct ADRIADomain <: Domain
     const cluster_id_col::String  # column of unique cluster ids
     init_coral_cover::YAXArray  # initial coral cover dataset
     const coral_growth::CoralGrowth  # coral
+    const coral_params::Dict{Symbol,Any}  # coral params
     const loc_ids::Vector{String}  # Location IDs that are represented (i.e., subset of loc_data[:, location_id_col], after missing locations are filtered)
     const removed_locs::Vector{String}  # indices of locations that were removed. Used to align loc_data, DHW, connectivity, etc.
     dhw_scens::YAXArray  # DHW scenarios
@@ -47,6 +48,7 @@ function Domain(
     cluster_id_col::String,
     init_coral_cover::YAXArray,
     coral_growth::CoralGrowth,
+    coral_params::Dict{Symbol,Any},
     location_ids::Vector{String},
     removed_locations::Vector{String},
     DHW::YAXArray,
@@ -77,6 +79,7 @@ function Domain(
         cluster_id_col,
         init_coral_cover,
         coral_growth,
+        coral_params,
         location_ids,
         removed_locations,
         DHW,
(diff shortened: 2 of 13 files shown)


ConnectedSystems:
YAML is what we use else where and it supports in-file comments, but I'll take a look first

Rosejoycrocker:
> YAML is what we use else where and it supports in-file comments, but I'll take a look first

Sure, happy to change to that. I'll sort out the failing formatting test now.

ConnectedSystems:
Sorry, could you say a few words on what the intended use case being supported is?

Changing the spec won't affect the data or any of the modelling done via ADRIA, so just trying to understand the purpose better

src/Domain.jl
```diff
@@ -18,16 +18,17 @@ mutable struct EnvLayer{S<:AbstractString,TF}
...
 No SSP/RCP data is preset.
 
 # Arguments
 - `path` : location of data package.
+- `coral_spec_path` : location of JSON with coral ecological parameters if not using default functional groups and size classes.
 """
-function load_domain(path::String)::Domain
-    return load_domain(path, "")
+function load_domain(path::String; coral_spec_path::String="")::Domain
+    return load_domain(path, ""; coral_spec_path=coral_spec_path)
```
ConnectedSystems, line 31: Sorry, I don't see how this works.

The Domain datasets themselves hold data relevant to initial coral cover, susceptibility to cyclones and other info.

src/ExtInterface/ADRIA/Domain.jl
```diff
@@ -176,14 +181,26 @@ function Domain(
...
     cover_params = ispath(init_coral_fn) ? (init_coral_fn,) : (n_group_and_size, n_locs)
     init_coral_cover = load_initial_cover(cover_params...)
 
+    if size(init_coral_cover, 1) != n_group_and_size
+        throw(
+            DimensionMismatch(
+                "
+The number of functional groups and size classes specified in the coral parameters file
+does not match the size of the initial coral cover matrix.
+"
```
ConnectedSystems, line 199: Multi-line strings are triple quoted:

```
"""
like so
"""
```

src/ecosystem/corals/CoralGrowth.jl
```diff
@@ -25,33 +25,37 @@ function CoralGrowth(n_locs::Int64, n_groups::Int64, n_sizes::Int64)::CoralGrowt
...
-        zeros(n_groups, n_locs),           # rec
-        zeros(n_groups, n_sizes, n_locs),  # sXr
+        # cache matrices
+        zeros(n_groups, n_locs),  # rec
+        zeros(n_groups, n_sizes, n_locs), # sXr
         zeros(n_groups, n_sizes, n_locs),  # X_mb
-        zeros(n_groups, n_sizes),          # r
-        zeros(n_groups, n_sizes)           # mb
+        zeros(n_groups, n_sizes),  # r
+        zeros(n_groups, n_sizes)   # mb
```
ConnectedSystems, line 58: This entire bit needs to be revisited as many of these cache are not used any more (they were for the older ODE method).

src/ecosystem/corals/Corals.jl
```diff
@@ -80,8 +80,8 @@ end
 
 Helper function defining coral colony diameter bin widths.
 """
-function bin_widths()
-    return bin_edges()[:, 2:end] .- bin_edges()[:, 1:(end - 1)]
+function bin_widths(bin_edges)
+    return bin_edges[:, 2:end] .- bin_edges[:, 1:(end - 1)]
 end
```
ConnectedSystems, line 85: This will break any code that used the older implementation?

src/ecosystem/corals/Corals.jl
```diff
@@ -119,9 +119,7 @@ Generate colony area data based on Bozec et al., [1].
        management in the Whitsundays.
      https://doi.org/10.13140/RG.2.2.26976.20482
 """
-function colony_areas()
-    # The coral colony diameter bin edges (cm) are: 0, 2, 5, 10, 20, 40, 80
-    edges = bin_edges(; unit=:cm)
+function colony_areas(edges)
```
ConnectedSystems, line 122: Same comment here. This will break any code that relies on the previous method.

To be clear, anyone who had written `ADRIA.colony_areas()` will see their code break.

src/io/inputs.jl
```diff
@@ -170,7 +170,30 @@ function load_cyclone_mortality(timeframe::Vector{Int64}, loc_data::DataFrame)::
     return ZeroDataCube(;
         timesteps=1:length(timeframe),
         locations=sort(loc_data.reef_siteid),
-        species=ADRIA.coral_spec().taxa_names,
+        species=ADRIA.default_coral_spec().taxa_names,
         scenarios=[1]
     )
 end
```
ConnectedSystems, line 176: I'm unsure of the thinking here as this will only work for the default coral specification and nothing else.

ConnectedSystems (commented):
Still waiting on the above but for now I'll say I wish you discussed these changes with someone beforehand.

The principle concerns I have are:

- Relies on a separate (and currently undocumented?) JSON file that the user provides when loading a domain.
- I suspect this will create mismatches all over ADRIA as some code rely on the "default" coral specification, and others don't
- Adds a `coral_params` field to the `Domain` type that is not consistently used. It also duplicates the function of the `model_spec`.
- Requires even more changes to the `metrics` module than what is currently included in this PR.

Specific to just changing the number of functional groups and size classes, a cleaner approach would be to add methods to allow the `model_spec` and `coral_spec` to be changed/updated with new coral-relevant factors.

See relevant code, which you would have seen when working on this PR:

https://github.com/open-AIMS/ADRIA.jl/blob/50c3aa0540ecfe7b60e2d66e287b39d90980db14/src/ecosystem/corals/Corals.jl#L465

https://github.com/open-AIMS/ADRIA.jl/blob/50c3aa0540ecfe7b60e2d66e287b39d90980db14/src/ecosystem/corals/Corals.jl#L370-L371

https://github.com/open-AIMS/ADRIA.jl/blob/50c3aa0540ecfe7b60e2d66e287b39d90980db14/src/ecosystem/corals/Corals.jl#L412

ConnectedSystems:
Marking this PR as a draft as more work needs to be done to support the described feature.

PR #914, opened 2025-01-18
Add method to display all DHW scenarios for the currently set RCP/SSP

ConnectedSystems:
As in title.

```julia
ADRIA.viz.dhw_scenarios(dom)
```

image

I've also taken the opportunity to flesh out the docstring for `ADRIA.viz.dhw_scenario(dom, some_id)` which was missing.


src/analysis/feature_set.jl
```diff
@@ -0,0 +1,117 @@
+"""
+    _filter_constants(scens::DataFrame)::DataFrame
+
+Filter out features/factors that do not vary.
+"""
+function _filter_constants(scens::DataFrame)::DataFrame
+    varying_cols = []
+    for (i, col) in enumerate(eachcol(scens))
```
Rosejoycrocker, line 8: So the loop here is more efifcient than something like `findall(dropdims(sum(diff(Matrix(scens);dims=1),dims=1),dims=1).==0.0)` ?

Rosejoycrocker (approved):
Looks good to me :)

src/analysis/feature_set.jl
ConnectedSystems: Adjusted to use a comprehension instead of mutating an existing vector.

I had to get something done quickly so what was there before was the easiest way without using too much brain power.

This should address your concern re efficiency while balancing for code readability.

PR #896, opened 2024-10-28
Add DEA functions and plotting to ADRIA.analysis

Rosejoycrocker:
Adds function to perform output-oriented (default) Data Envelopment Analysis (DEA) given inputs X and output metrics Y. DEA is used to measure the performance of entities (scenarios), where inputs are converted to outputs via some process. Each scenario's "efficiency score" is calculated relative to an "efficiency frontier", a region representing scenarios for which outputs
cannot be further increased by changing inputs (scenario settings). Scenarios on the frontier serve as "benchmarks" or "peers", associated with best practice restoration scenarios. Scenarios with efficiencies not equal to 1 can be improved to be more efficient.

Allows usage with generic input functions, such as functions representing cost as a function of intervention scenario.

Flagging @ConnectedSystems as  we discussed this previously

Working example is:
```julia
using Revise, Infiltrator
using ADRIA
using DataFrames, YAXArrays
using GLMakie, GeoMakie, GraphMakie
using Distributions

dom = ADRIA.load_domain( "domain path", "45")

scens = ADRIA.sample(dom, 2^10)
rs = ADRIA.run_scenarios(dom, scens, "45")

n_scens = size(scens,1)

# Get cost of deploying corals in each scenario, randomised cost example
cost = YAXArray(collect(range(100,stop=1000000, length=n_scens)) .+ rand(Uniform(1000, 2000), n_scens))

# Get mean coral cover and shelter volume for each scenario
s_tac = dropdims(
    mean(ADRIA.metrics.scenario_total_cover(rs); dims=:timesteps); dims=:timesteps
)
s_sv = dropdims(
   mean(ADRIA.metrics.scenario_shelter_volume(rs); dims=:timesteps); dims=:timesteps
    )

# Do output oriented DEA analysis seeking to maximise cover and shelter volume for minimum
# deployment cost.
DEA_scens = ADRIA.economics.data_envelopment_analysis(cost, Array(s_tac), Array(s_sv))
dea_fig = ADRIA.viz.data_envelopment_analysis(rs, DEA_scens)

```

example_dea_fig

PR #915, opened 2025-01-18
Minor tweaks to taxa visualization

ConnectedSystems:
The following tweaks were applied:

- Reduced large display of plot (1200x1200 to 1200x600)
- Adjust title text to be in title case
- Change position of legend to left hand side to avoid hiding other plot elements. This should be moved outside the plot but I don't have time to make the necessary adjustments.

image

One change I needed but couldn't work out was changing the scenario type name to title case ("counterfactual" -> "Counterfactual").

Please note there was some chatter about `viz.taxonomy` in another issue, I'm not sure if the changes there will impact this PR.
https://github.com/open-AIMS/ADRIA.jl/pull/896#issuecomment-2597022820

### DIFF
--- a/ext/AvizExt/viz/taxa_dynamics.jl
+++ b/ext/AvizExt/viz/taxa_dynamics.jl
@@ -65,17 +65,23 @@ function ADRIA.viz.taxonomy(
     axis_opts::OPT_TYPE=DEFAULT_OPT_TYPE(),
     series_opts::OPT_TYPE=DEFAULT_OPT_TYPE()
 )::Figure
-    fig_opts[:size] = get(fig_opts, :size, (1200, 600))
-    f = Figure(; fig_opts...)
-
-    g = f[1, 1] = GridLayout()
-
     _scenarios = copy(scenarios[1:end .∈ [relative_taxa_cover.scenarios], :])
     scen_groups = if get(opts, :by_RCP, false)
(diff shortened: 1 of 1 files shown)


Zapiano:
Did you plot this using a result set containing only counterfactuals? Because if the result set contains guided, unguided and counterfactuals, the plot looks like this:

image

Also, if `:by_functional_groups` is `false`, that's how this plot looks like:

image


Maybe the default size of this plot should probably adapt to the number of scenario types and to the value of `:by_functional_groups`?

Zapiano:
Also, I made the scenario type title case.

ConnectedSystems:
> Did you plot this using a result set containing only counterfactuals? Because if the result set contains guided, unguided and counterfactuals, the plot looks like this

No I didn't because I was only looking at a single scenario type at the time. But more to the point, it should auto-adjust the size as needed.

Another improvement that could be easily made is to link the x-axis limits so if you move/zoom one plot, the other graphs follow.

Zapiano:
I have committed an adjustment, see what you think. Here are some examples of how it looks like with different `opts` and number of scenario types:

tmp_tax_all_fn

tmp_tax_all_rcp_fn

tmp_tax_all

tmp_tax_one

ConnectedSystems:
Another, probably more important, issue is that this method seems to be very slow. Suspiciously slow...

Zapiano (approved):
LGTM

PR #917, opened 2025-01-27
Updated factor bounds setting method

ConnectedSystems:
Previous implementation looped over each factor and set these individually, which is very slow.

Updated to set the factors in a single pass.

This is currently a draft as tests are needed.

### DIFF
--- a/src/io/sampling.jl
+++ b/src/io/sampling.jl
@@ -512,7 +512,7 @@ set_factor_bounds(dom, :wave_stress, (0.1, 0.2))
 ```
 """
 function set_factor_bounds(dom::Domain, factor::Symbol, new_dist_params::Tuple)::Domain
-    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog=1 _category=:deprecation
+    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog = 1 _category = :deprecation
 
     set_factor_bounds!(dom, factor, new_dist_params)
     return dom
@@ -532,7 +532,7 @@ function set_factor_bounds!(dom::Domain, factor::Symbol, new_dist_params::Tuple)
 end
 
 function set_factor_bounds(dom::Domain; factors...)::Domain
-    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog=1 _category=:deprecation
+    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog = 1 _category = :deprecation
 
     set_factor_bounds!(dom; factors...)
     return dom
--- a/src/io/sampling.jl
+++ b/src/io/sampling.jl
@@ -512,7 +512,8 @@ set_factor_bounds(dom, :wave_stress, (0.1, 0.2))
 ```
 """
 function set_factor_bounds(dom::Domain, factor::Symbol, new_dist_params::Tuple)::Domain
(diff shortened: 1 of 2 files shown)


src/io/sampling.jl
```diff
@@ -512,7 +512,8 @@ set_factor_bounds(dom, :wave_stress, (0.1, 0.2))
 ```
 """
 function set_factor_bounds(dom::Domain, factor::Symbol, new_dist_params::Tuple)::Domain
-    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog=1 _category=:deprecation
+    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog = 1
+    _category = :deprecation
```
ConnectedSystems, line 516: Doesn't the macro require all the arguments to be on one line? I suspect this breaks the use of the `@warn` macro?

src/io/sampling.jl
```diff
@@ -528,16 +529,17 @@ function set_factor_bounds!(dom::Domain, factor::Symbol, new_dist_params::Tuple)
...
 
     update!(dom, ms)
-    return nothing
+    return dom
 end
 
 function set_factor_bounds(dom::Domain; factors...)::Domain
-    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog=1 _category=:deprecation
+    Base.@warn "set_factor_bounds is deprecated, use set_factor_bounds! instead" maxlog = 1
+    _category = :deprecation
```
ConnectedSystems, line 537: Same comment here - I suspect adding a new line breaks the code.

src/io/sampling.jl
```diff
@@ -549,13 +551,15 @@ function set_factor_bounds!(dom::Domain; factors...)::Nothing
...
+    ms[factor_rows, :val] .= [
+        v isa Int ? round(n) : n for (v, n) in zip(old_vals, new_vals)
+    ]
 
     # Update `is_constant` column
     ms[!, :is_constant] .= (ms[!, :lower_bound] .== ms[!, :upper_bound])
 
     update!(dom, ms)
-    return nothing
+    return dom
```
ConnectedSystems, line 562: The `!` in the method name indicates side-effects should be expected.

The intention was to modify `dom` in place.

```julia
# Change this:
dom = set_factor_bounds(dom; factors...)

# to:

# `dom` is modified in place
set_factor_bounds(dom; factors...)
```

That said, I don't have an aversion to also returning the domain type again. Are you thinking of supporting method chaining/piping or similar?

pascal082: @ConnectedSystems  I was trying to keep flexibility for both styles—modifying dom in place while also returning it for method chaining. No strong preference, happy to align with your/team's style.

src/io/sampling.jl
pascal082: @ConnectedSystems - it's valid. In Julia, I can see the @warn macro supports multiple formatting styles (line breaks, semicolons, and parentheses ). I was following the Blue Style guideline which the project is currently moving to follow (written on the developer page), and this adheres to a 92-character line length limit. Do you want me to add a new test case to test the old function and ensure nothing breaks? - Happy to go with your preference.

ConnectedSystems: You have to wrap code in backticks, you've just tagged a random GitHub user into this conversation :P

ConnectedSystems: Can you point me to an example where `@warn` is split across multiple lines like this? As far as I can tell, Julia will interpret this as two different calls: One where the warning is raised and another where a `_category` variable to set.

Zapiano (approved):

PR #922, opened 2025-02-21
Pin DiskArrays to 0.4.9

vitorhirata:
YAXArrays uses DiskArrays and it is raising an error in its last update. To solve the problem we add and pin DiskArrays version to 0.4.9.
It also updates the Manifest. I ran tests to check the code is still working.


vitorhirata:

Zapiano (approved):
LGTM

PR #923, opened 2025-02-21
Fix tests of viz/spatial and metrics/spatial

vitorhirata:
Running metrics/spatial and viz/spatial tests raised this error: `ERROR: LoadError: UndefVarError: 'cf_difference_loc' not defined`. This error was caused because commit 46fd368 renamed the function cf_difference_loc to ensemble_loc_difference but it was still imported in the test function. The function was not being used.
This commit removes the import of cf_difference_loc function.


vitorhirata:

Zapiano (approved):

PR #924, opened 2025-02-21
Pin DiskArrays to 0.4.9 in docs Manifest and v1.11 Manifest

vitorhirata:
The workflow of building the documentation was failing. Following commit 323d65f, this commit add and pin DiskArrays to 0.4.9 in the docs Manifest and in Julia v1.11 Manifest. It also updates the docs manifest.

### DIFF
--- a/Manifest-v1.11.toml
+++ b/Manifest-v1.11.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.11.2"
 manifest_format = "2.0"
-project_hash = "4ffb592df61e4d217c5b2589556288d921d4ba0f"
+project_hash = "5b975ff5ee93b291c8ca4988afc831aa04ba535d"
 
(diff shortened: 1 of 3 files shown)


vitorhirata:

ConnectedSystems:
@vitorhirata how did you pin the package?

I ask as it doesn't seem to have modified the manifest file for Julia v1.11.
It might be be because you're still on Julia v1.10?

vitorhirata:
> @vitorhirata how did you pin the package?
> 
> I ask as it doesn't seem to have modified the manifest file for Julia v1.11. It might be be because you're still on Julia v1.10?

I have both installed but I was running the code using v1.10 because it is on the default manifest. I will update the PR to change both manifest files

ConnectedSystems (approved):
Thanks Vitor, this LGTM.

PR #925, opened 2025-02-23
Tentative Plotlylight Extension package

ConnectedSystems:
Initial support for visualizations based on PlotlyJS via [PlotlyLight](https://github.com/JuliaComputing/PlotlyLight.jl).

## Reason for this PR

Recently, the need to showcase some kind of results visualization to support RRAP-related work and future projects have been (re-)raised.

This PR provides an initial/tentative implementation of the `scenarios` visualization method.

I've elected to leverage Julia's extension package system to support a single plot type (scenario outcomes) using PlotlyLight given the issues outlined further below and the comparatively small visualization capability required. Analysts/Modellers can continue performing assessments "offline".

As it is an extension package, the usual workflow remains the same:

```julia
using PlotlyLight
using ADRIA

dom = ADRIA.load_domain(...)
scens = ADRIA.sample(dom, 128)

r_tac = ADRIA.metrics.scenario_total_cover(rs)
p = ADRIA.viz.scenarios(rs, r_tac)
```

image

## Further Background / Context

In the early days of ADRIA there was a strong push for some kind of user interface.

Attempts had been made previously in [MATLAB GUI](https://au.mathworks.com/discovery/matlab-gui.html) before my time on this project. In retrospect I should have pushed back on the idea of a UI completely as ADRIA was not mature enough to support anything of the sort at the time. ADRIA was barely more than a loose assortment of scripts and the underlying programmatic interfaces were constantly changing. Updates to formalize contracts/interfaces would most likely break the UI. Not to mention a likely change in languages.

The second "attempt" after the initial migration to Julia was a web-based dashboard written in Python Plotly. I do not know why Python was selected, as Plotly had become available for [Julia](https://plotly.com/julia/) some time before this (although based on some reports I saw at the time, it had some issues). But an initial Python implementation had been written (see [here](https://github.com/open-AIMS/ADRIA_dashboard)).

To support the web dashboard an Oxygen.jl web service was set up. What was lacking was a mature analysis/visualization pipeline. At the time, the analysis/visualization interfaces had not yet been fully written (what is now the `viz`, `metrics`, `analysis`, submodules), so again, the visualization methods were essentially a loose assortment of functions/scripts except now spread in two languages and, at times, with duplicate functionality. I tried to put in some structure but doing this and working on ADRIA itself, it was all a bit much.

Scenario data was being served by Oxygen for processing in Python. The full scenario set needed to be passed to Python/JS to support specific types of analyses which required the full dataset. While the data processing itself was not so time consuming (although at times orders of magnitude slower compared to Julia), the constant data passing and conversion of scenario outcomes from matrices to JSON was horrendously slow.

After this, I made to call to switch to Makie given time/effort were better spent:

1. Maturing analysis and assessment capabilities
2. Address issues with the ecosystem model
3. Building interfaces to support interoperability with other RRAP models
4. Data processing would ideally stay in Julia given the comparatively high volume of data
5. There wasn't actually a strong need for a user interface or interactivity at the time
6. Packages like PlotlyLight didn't exist at the time.

I was also banking on community efforts to support interactivity (either local or web-based) in the Makie ecosystem would mature in the intervening time (from memory this was ~3 years ago).

## Current state

The Makie ecosystem ADRIA depends on for visualizations, however, has not yet progressed to support client-side interactivity.
This was hoped to be done via WGLMakie and Bonito (then JSServe). While _possible_, WGLMakie requires an active Julia server/runtime to be running in the background for interactive 2D plots until functionality is ported across to Three.js.

Although we now have a fairly mature and stable set of methods to support analyses and assessments and a consistent set of visualization methods and a (semi)-standardized approach for their use, there are some concerns:

- Some methods have exploded in complexity (e.g., [taxonomy vizualizations](https://github.com/open-AIMS/ADRIA.jl/blob/dcd9a23c466335008554d512fb9d42055c6070cd/ext/AvizExt/viz/taxa_dynamics.jl#L91)) and could use a revisit/rewrite
- The Makie package itself is quite a heavy dependency
- The Plotly ecosystem is more mature and has greater support across languages

If this extension package is accepted, I suggest we only port/re-create the visualizations needed. It may be that we don't need web-based interactivity for the vast majority of plot types in the short-term.

PlotlyLight was selected over PlotlyJS.jl as @arlowhite is a javascript aficionado and the Javascript API is more mature compared to the Julia one and would hopefully ease cognitive burden.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -72,9 +72,11 @@ BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
 GeoMakie = "db073c08-6b98-4ee5-b6a4-5efafb3259c6"
 GraphMakie = "1ecd5474-83a3-4783-bb4f-06765db800d2"
 Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+PlotlyLight = "ca7969ec-10b3-423e-8d99-40f33abb42bf"
 
 [extensions]
 AvizExt = ["Makie", "GeoMakie", "GraphMakie"]
+PlotlyLightExt = ["PlotlyLight"]
 
 [compat]
 Aqua = "0.8, 0.9"
@@ -91,7 +93,7 @@ DataFrames = "1"
 DataStructures = "0.18, 0.2"
 Dates = "1"
 DimensionalData = "0.25, 0.26, 0.27"
-DiskArrays = "=0.3.23, ^0.4.5"
+DiskArrays = "^0.4.5"
 Distances = "0.10"
 Distributed = "1"
 Distributions = "0.25"
@@ -123,6 +125,7 @@ NetCDF = "0.11, 0.12"
 OnlineStats = "1.5"
 OrderedCollections = "1"
 PkgVersion = "0.2, 0.3"
+PlotlyLight = "0.11"
 PrecompileTools = "1"
 Printf = "1"
 ProgressMeter = "1"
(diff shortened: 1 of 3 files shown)


ConnectedSystems:
@arlowhite @DanTanAtAims 

See working example here:

https://github.com/open-AIMS/MADAME-api/blob/7e8b3ed98f8d7182722efe3f37c55468fdd19f55/src/routing/model_run.jl#L66


plotlylight_working_example

ConnectedSystems:
Closing in favour of the larger restructure going on in `pkg-reorg` branch

Zapiano:
Will be covered on a separate PR
FYI @ConnectedSystems

PR #929, opened 2025-02-28
Replace and deprecate remaining use of `site_k_area`

### LINKED ISSUE #564: Find and deprecate inconsistent methods
We have a set of functions to do with `site`, such as `site_k_area()`.

These should be deprecated and ultimately removed in favour of those using `location` terminology (e.g., `location_k_area()`.

Docstrings and comments should be updated accordingly as well!

ConnectedSystems:
Replaces remaining instances of `site_k_area()` with `loc_k_area()`.

Deprecates `site_k_area()` so any use gets auto-forwarded to `loc_k_area()` (with `@deprecated`, a new macro I discovered).

Closes #564

### DIFF
--- a/docs/src/usage/analysis.md
+++ b/docs/src/usage/analysis.md
@@ -211,7 +211,7 @@ plan_horizon = Int64(scen["plan_horizon"])
 decay = 0.99 .^ (1:(plan_horizon + 1)) .^ 2
 dhw_projection = ADRIA.decision.weighted_projection(dhw_scens, 1, plan_horizon, decay, 75)
 # Connectivity
-area_weighted_conn = dom.conn.data .* ADRIA.site_k_area(dom)
+area_weighted_conn = dom.conn.data .* ADRIA.loc_k_area(dom)
 conn_cache = similar(area_weighted_conn)
 in_conn, out_conn, network = ADRIA.connectivity_strength(
     area_weighted_conn, sum_cover, conn_cache
--- a/src/ADRIA.jl
+++ b/src/ADRIA.jl
@@ -89,7 +89,7 @@ export
     run_scenario, coral_spec,
     create_coral_struct, Intervention, SimConstants,
     SeedCriteriaWeights, FogCriteriaWeights,
-    loc_area, site_k_area,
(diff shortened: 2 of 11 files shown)


Zapiano:
@ConnectedSystems this PR need a rebase before it's merged.

ConnectedSystems:
@Zapiano done. Had to merge main then rebase, but thankfully the conflicts were easily handled.

Zapiano:
2 things:

1. When running the tests, I got a `ERROR: LoadError: MethodError: no method matching *(::Float64, ::typeof(loc_k_area))
The function `*` exists, but no method is defined for this combination of argument types.` on line `ADRIA.jl\src\scenario.jl:1022` https://github.com/open-AIMS/ADRIA.jl/blob/2af2792cdd2402a778e0233e048967d3c2d7c09b/src/scenario.jl#L1022
2. Also when running the tests, I got a warning followed by an error that, for some reason, did stop the tests from running saying

```
WARNING: Method definition site_k_area(Any) in module ADRIA at deprecated.jl:103 overwritten on the same line (check for duplicate calls to `include`).
ERROR: Method overwriting is not permitted during Module precompilation. Use `__precompile__(false)` to opt-out of precompilation.
```

image


Not sure how relevant these are.

ConnectedSystems:
Hmmm... there may be a poorly resolved conflict...

Kind of surprised about the view-based matrix multiplication, it wasn't an issue before I think.

ConnectedSystems:
>     1. When running the tests, I got a `ERROR: LoadError: MethodError: no method matching *(::Float64, ::typeof(loc_k_area)) The function `*` exists, but no method is defined for this combination of argument types.` on line `ADRIA.jl\src\scenario.jl:1022` https://github.com/open-AIMS/ADRIA.jl/blob/2af2792cdd2402a778e0233e048967d3c2d7c09b/src/scenario.jl#L1022

Got it. There was a variable named `loc_k_area`. Because we're now using the function, it's raising the fact that we can't multiple by a method name.

Fixing now.

ConnectedSystems:
@Zapiano should be okay to merge now.

Zapiano (approved):
LGTM

PR #934, opened 2025-03-16
Considering spatial dispersal of intervention locations

ConnectedSystems:
This PR removes the post-processing and instead relies solely on the MCDA assessment by introducing two factors (four in total as they are duplicated for fogging and seeding interventions).

- `cluster_diversity` : Locations are grouped into clusters through a pre-processing step. \
How under-represented the selected reef/site cluster is. The higher the value, the more "under-represented" the cluster is.

- `geographic_separation` : The mean distance to the closest `n` neighbors (where n := 10 for now) \
giving an indication of the local "centrality" of the location.

By default, both of these are given a weight of 0.5

For `cluster_diversity`, a higher weight gives preference to smaller clusters whose members are under-represented in the spatial domain.

For `geographic_separation` a higher weight preferences locations that are closer to its neighbors.

---

## Context

The previous approach to selecting locations required a post-processing of selected locations, swapping locations for next-best rank if too many locations were within the same cluster. The process was controlled by `max_members` (the maximum number of locations allowed within a single cluster).

While it ensured spatial spread, it made MCDA assessments difficult to interpret.

This approach simplifies analysis of MCDA outputs, removing `max_members` but adds four additional factors in total (cluster diversity and geographic separation for both seeding and fogging interventions).


ConnectedSystems:
@vitorhirata this approach should be much better for what you need to do.

@Zapiano I seem to be having issues with the DiskArrays.jl dependency, even with both the sandbox and ADRIA environments using v0.4.9 (as pinned). 

It seems in some places in code, Julia is not able to auto-cast from `DimVector` to `Vector`. 
In other places the access rules have regressed or changed so that `collect(some_yaxarray[some_selector])` does not work and now expects `collect(some_yaxarray[some_selector]).data`

Could you confirm tests fail on your end too?

This PR is marked as a draft as it expects #933 and because of the above issues.

vitorhirata:
> @vitorhirata this approach should be much better for what you need to do.

That is great, it will help me a lot!

test/decisions/location_spread.jl
```diff
@@ -0,0 +1,120 @@
...
+    geographic_separation,
+    decision_matrix,
+    select_locations
+
+@testset "Location selection" begin
+    """
+    Tests to ensure cluster diversity and geographic separation of intervention sites
+    are applied correctly.
+    """
+
```
vitorhirata, line 17: I would potentially add a simpler test just to cover basic functionalities of the select_locations and change the file name to location_selection. I know this is not related to this PR, but would improve coverage and won't give much work. For example, just testing, minimum locations and valid locations. I could add a commit with that if that helps

src/Domain.jl
```diff
@@ -34,6 +34,127 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+end
+function distance_matrix(loc_data::DataFrame)::Matrix{Float64}
+    return distance_matrix(centroids(loc_data))
+end
+
+"""
+    mean_distance(dist_matrix::Matrix{Float64})::Vector{Float64}
+
+Calculate the mean distance between a location and all other locations to give an
+indication of spatial spread.
```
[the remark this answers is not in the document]
ConnectedSystems: Hmm, I'll be more explicit then. I also realise the description is slightly incorrect as it's the `n` closest locations, not all locations.

test/decisions/location_spread.jl
vitorhirata: I didn't find the tests I mentioned in the file you linked but I agree this is out of scope for this PR. I will create a PR adding more tests later

ConnectedSystems: Well, it doesn't test select locations but rather the behaviour of the MCDA.

The principle is the same I think though.

But you're right, there should be a test to cover location selection at the interface level

src/decision/Criteria/DecisionPreferences.jl
```diff
@@ -140,7 +140,9 @@ function rank_by_index(
     end
 
     # Recreate preferences, removing criteria that are constant for this scenario
-    _dp = filter_criteria(dp, is_const)
+    if !all(.!is_const)
```
Zapiano, line 143:
```suggestion
    if any(is_const)
```

src/decision/Criteria/DecisionPreferences.jl
```diff
@@ -140,7 +140,9 @@ function rank_by_index(
...
+        _dp = filter_criteria(dp, is_const)
+    end
 
     # Assess decision matrix only using valid (non-constant) criteria
     res = solve(_dp, dm[criteria=.!is_const], method)
```
vitorhirata, line 148: If it doen't enter in the if `_dp` is not defined. Also, this was causing this error when I tried to run: `ERROR: UndefVarError: `_dp` not defined in local scope Suggestion: check for an assignment to a local variable that shadows
a global of the same name.`
You could just reassign `dp`
```suggestion
    if !all(.!is_const)
        dp = filter_criteria(dp, is_const)
    end

    # Assess decision matrix only using valid (non-constant) criteria
    res = solve(dp, dm[criteria=.!is_const], method)
```

Zapiano:
> @vitorhirata this approach should be much better for what you need to do.
> 
> @Zapiano I seem to be having issues with the DiskArrays.jl dependency, even with both the sandbox and ADRIA environments using v0.4.9 (as pinned).
> 
> It seems in some places in code, Julia is not able to auto-cast from `DimVector` to `Vector`. In other places the access rules have regressed or changed so that `collect(some_yaxarray[some_selector])` does not work and now expects `collect(some_yaxarray[some_selector]).data`
> 
> Could you confirm tests fail on your end too?
> 
> This PR is marked as a draft as it expects #933 and because of the above issues.
> 
> A corresponding metric indicating the spatial spread of intervention locations is probably a good idea and needs to added to [`feature_set()`](https://github.com/open-AIMS/ADRIA.jl/blob/c6ce2585e02bf133024f33bccc0319110206a780/src/analysis/feature_set.jl#L52-L53)

All green here.

image

src/decision/Criteria/DecisionPreferences.jl
ConnectedSystems: Ha, yes of course!

ConnectedSystems:
> All tests are passing here both on this branch and on the main branch.
> 
> image

Might be due to a mismatch between my sandbox and ADRIA environments.

```
(sandbox) pkg> status -m DiskArrays YAXArrays DimensionalData
Status `C:\Users\tiwanaga\projects\ADRIA.jl\sandbox\Manifest.toml`
  [0703355e] DimensionalData v0.29.13
⌃ [3c3547ce] DiskArrays v0.4.9
⌅ [c21b50f5] YAXArrays v0.5.14
```

```
Project ADRIA v0.11.0
Status `C:\Users\tiwanaga\projects\ADRIA.jl\Manifest-v1.11.toml`
⌅ [0703355e] DimensionalData v0.27.9
⌃ [3c3547ce] DiskArrays v0.4.9 ⚲
⌅ [c21b50f5] YAXArrays v0.5.12
```

Might be worth pinning YAXArrays as well as it seems somewhat fragile too.

ConnectedSystems:
Tests now pass when invoked from package environment in Julia v1.11.4 and when explicitly using package dependencies:

```julia-repl
] test
```

One minor note, not relevant to this PR: there seems to be a test that doesn't test anything.

```
Test Summary:                       | Total  Time
Visualization of spatial.jl metrics |     0  1.1s
```

Zapiano:
> @Zapiano when able, could you please share how you ran the tests above (in your sandbox or through `] test` ?) as well as the package versions?

I was running on ADRIA environment doing `]test`

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+give an indication of spatial spread.
+
+# See also
+- `nearest_neighbor_distances()`
+"""
+function mean_distance(dist_matrix::Matrix{Float64})::Vector{Float64}
+    n_locs = size(dist_matrix, 1)
+    mean_distances = vec(sum(dist_matrix; dims=2) ./ (n_locs - 1))
+
+    return mean_distances
```
Zapiano, line 99:
```suggestion
    return vec(sum(dist_matrix; dims=2) ./ (n_locs - 1))
```

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+# Returns
+Matrix of pairwise distances between locations
+"""
+function distance_matrix(coords::Vector{Tuple{Float64,Float64}})::Matrix{Float64}
+    n_locs = length(coords)
+    dist_matrix = zeros(n_locs, n_locs)
+
+    for i in 1:n_locs
+        for j in (i + 1):n_locs
+            if i == j
```
Zapiano, line 55: If `j` starts at `i+1` and goes up, it can never be equal `i`, can it?

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+                continue
+            end
+
+            dist = Distances.haversine(coords[i], coords[j])
+            dist_matrix[i, j] = dist
+            dist_matrix[j, i] = dist
```
Zapiano, line 61:
```suggestion
            dist_matrix[i, j] = dist_matrix[j, i] = Distances.haversine(coords[i], coords[j])
```

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+Vector of mean distances (in meters) to nearest `n_neighbors` for each location.
+"""
+function nearest_neighbor_distances(
+    dist_matrix::Matrix{Float64},
+    n_neighbors::Int
+)::Vector{Float64}
+    n_locs = size(dist_matrix, 1)
+
+    # Adjust n_neighbors if there aren't enough locations
+    effective_n = min(n_neighbors, n_locs - 1)
```
Zapiano, line 121: Maybe throw a warn in this case?

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+
+    # Calculate mean distance to n closest neighbors for each location
+    mean_distances = zeros(n_locs)
+    for i in 1:n_locs
+        # Get distances from this location to all others, excluding self
+        other_indices = [j for j in 1:n_locs if j != i]
```
Zapiano, line 127: Not sure how relevant it is to improve performance here but since there can be many locations: there could be a cache vector `other_indices = 2:n_locs` outside this loop and then you add, to the end of the loop: 
```
if i < n_locs
    other_indices[i] = i
end
```

src/Domain.jl
```diff
@@ -34,6 +34,130 @@ function unique_loc_ids(d::Domain)::Vector{String}
...
+        distances = dist_matrix[i, other_indices]
+
+        # Find the n closest neighbors
+        if length(distances) <= effective_n
```
Zapiano, line 131: `dist_matrix` is a square matrix, so this condition is going to be either always true or always false every time this function is called. I know is just a small thing but this `length(distances) <= effective_n` check can be done only once outside the look and save in a variable.

src/ExtInterface/ReefMod/ReefModDomain.jl
```diff
@@ -122,6 +122,9 @@ function load_domain(
     # GBRMPA zone types are not contained in matfiles
     spatial_data[:, :zone_type] .= ["" for _ in 1:nrow(spatial_data)]
 
+    dist_matrix = distance_matrix(spatial_data)
+    spatial_data.mean_to_neighbor .= nearest_neighbor_distances(dist_matrix, 10)
```
Zapiano, line 126: I feel like this `n_neighbors` should be a kwarg... 🤔

Zapiano (changes requested):
A few minor comments.

src/ExtInterface/ReefMod/ReefModDomain.jl
ConnectedSystems: Yeah probably - at the time I felt like we'd never really change it...

Zapiano: That's why a kwarg would be good, with a default value. but I'm happy to leave it this way of you prefer.

src/Domain.jl
ConnectedSystems: No actually, we do need it.

i loops from 1 to 10
j loops from 2 to 10 for each i

In the second loop, i will be equal to j (i = 2, j = 2)

Zapiano: Hum, but in your example when `i = 1`, `j` goes from `2` to `10`. Then When `i = 2`, `j` goes from `3` to `10`, etc. Then `i` is never equal `j`. Even though `i` goes from `1` to `10` and `j` goes from `2` to `10`, they are never equal.

src/Domain.jl
ConnectedSystems: Tested, seems there is effectively zero difference in this case - the time/allocation is spent in the comprehension. That said, there's probably a smart way of avoiding the comprehension completely...

```
julia> @be $x .= [j for j in 1:1000 if j != 100]
Benchmark: 459 samples with 10 evaluations
 min    2.304 μs (9 allocs: 21.844 KiB)
 median 4.075 μs (9 allocs: 21.844 KiB)
 mean   21.778 μs (9 allocs: 21.844 KiB, 0.62% gc time)
 max    3.228 ms (9 allocs: 21.844 KiB, 95.16% gc time)

julia> @be $y = [j for j in 1:1000 if j != 100]
Benchmark: 588 samples with 11 evaluations
 min    1.998 μs (9 allocs: 21.844 KiB)
 median 2.373 μs (9 allocs: 21.844 KiB)
 mean   15.022 μs (9 allocs: 21.844 KiB, 0.66% gc time)
 max    3.425 ms (9 allocs: 21.844 KiB, 97.95% gc time)
```

Zapiano: Sorry, what did you test exactly? In my suggestion there's no comprehension, you only change one element of the array at each iteration.

Zapiano: And you only create one vector, before the loop (which seems different from you test)

src/Domain.jl
ConnectedSystems: Actually, is this even necessary?

This is already taking the lower number:

```julia
sorted_indices = sortperm(distances)[1:effective_n]
```

So the case should be handled already...

src/ExtInterface/ReefMod/ReefModDomain.jl
ConnectedSystems: I don't mind either way really

src/decision/Criteria/SeedCriteria.jl
```diff
@@ -179,164 +188,79 @@ function select_locations(
             # Return empty vector to signify no ranks
             return String[]
         end
-
         rethrow(err)
     end
 
     if length(considered_locs) == 0
         return String[]
     end
```
vitorhirata, line 196: I just noticed that with the changes made in this PR the only place `considered_locs` is used is here. However, this validation is not necessary and the argument `considered_locs` can be removed from the `select_locations` function. 

In run_model there is already a validation checking if `length(considered_locs) > 0` ([here](https://github.com/open-AIMS/ADRIA.jl/blob/812d1a57e503e26249a9fcdabba19ef010b7611e/src/scenario.jl#L887)) and there is a filtering: `decision_mat[location=locs_with_space[_valid_locs]]` that would return an error if all locations are invalid. The same is valid for other uses of the `select_locations` function. 
If we want to be extra cautious we could a validation `if length(loc_names) == 0; return String[]`.

vitorhirata (commented):
Sorry to add more comments to this PR again, but I just noticed one of the arguments for select_locations could be removed

src/Domain.jl
ConnectedSystems:
```julia
x = zeros(999)
x .= [j for j in 1:1000 if j != 100]

# vs

y = [j for j in 1:1000 if j != 100]
```

If there is no difference for one iteration, it seems unlikely there will be a difference for $n$ iterations...

src/decision/Criteria/SeedCriteria.jl
ConnectedSystems: It is true that the only place it is currently being used right now is within `scenarios.jl`.

We do, however, have functionality to run a standalone location selection process, and that will need to be updated as well (not part of this PR).
I'm waiting to cleanly separate it out in the future when the code in the `scenario.jl` file gets neatly compartmentalized.

src/Domain.jl
ConnectedSystems: You're right! I thought I needed it for some reason but past late-night self was being smart and just forgot to delete the unnecessary check!

src/Domain.jl
ConnectedSystems: I'll put a separate warn/check for the standalone version.

I'd like to avoid spamming the terminal with warnings for the scenario-based method.

Zapiano (approved):
LGTM

PR #935, opened 2025-03-19
Pathway diversity analysis - switching probability

vitorhirata:
This PR begins the introduction of a new analysis using pathway diversity. Ref: https://www.ecologyandsociety.org/vol25/iss3/art19/. According to this concept a system is more resilient if actors have more options available and they are maintained over time.
Here we define one option as a set of weights given to the MCDA analysis. Below is a plot of the definition of the options:
```julia
options = ADRIA.analysis.option_seed_preference(; include_weights=true)
fig = ADRIA.viz.mcda_options(options)
```
option_weight_matrix

Each option connects to an instance of ADRIA.decision.SeedPreferences and considering `decision_matrix`, `mcda_method`, `valid_seed_locs`, `min_locs` are defined we can call `selected_locations` for all the optins:
```julia
options.selected_locations = [Vector{String}() for _ in 1:number_options]
for row in eachrow(options)
    row.selected_locations = ADRIA.select_locations(
        row.preference, decision_matrix, mcda_method, valid_seed_locs, min_locs
    )
    colors = [location == 0 ? :grey : :red for location in (dom.loc_ids .∈ [row.selected_locations])]
    fig = ADRIA.viz.map(dom.loc_data; geom_col=:geom, color=colors)
    save("seleceted_locations_$(row.option_name).png", fig)
end
```
Screenshot_2025-03-27_12-08-07

Lastly, we can call ADRIA.analysis.switching_probability to compute the probability of switching from and to all options:
```julia
number_options = size(options, 1)
prob_matrix = zeros(number_options, number_options)
for source in 1:number_options
    probs = ADRIA.analysis.switching_probability(options.selected_locations[source], decision_matrix, dom.loc_data, mcda_method, min_locs)
    prob_matrix[source,:] = 100 * probs.probability
end
# plot prob_matrix using heatmap!
```
probability_matrix

Next steps includes sample all possible combinations of options over time, passing a time series of options to follow at each timestamp, and make selec_locations follow the option SeedPreferences pre-define for the timestamp. Lastly, we compute the probability of each pathway and compute pathway diversity.

### DIFF
--- a/src/analysis/pathway_diversity.jl
+++ b/src/analysis/pathway_diversity.jl
@@ -1,4 +1,75 @@
 """
+    switching_probability(past_option::Symbol, decision_matrix::YAXArray)::Dict{Symbol, Float64}
+
+Compute switching probability for all defined pathway diversity options.
+
+# Arguments
+- `past_option` : Symbol with past option.
+- `decision_matrix` : Decision matrix with criterias used for MCDA analysis.
+- `loc_data` : Location data for all locations.
+- `mcda_method` : MCDA method used in selec_locations.
+- `min_locs` : Minimum number of locations to be selected by the MCDA algorithm
+- `ports` : Dataframe with name and coordinate of ports.
+- `weight` : Weigh used to cost_index. The complementary is used to weight the option_similarity.
+Defaults to 0.5
+
+# Returns
+DataFrame with probability of switching to each option or the probability value for one option if the option is passed.
+"""
+function switching_probability(
+    past_option::Symbol,
+    decision_matrix::YAXArray,
+    loc_data::DataFrame,
+    mcda_method::Union{Function,DataType},
+    min_locs::Int64;
+    ports::Union{DataFrame,Nothing}=nothing,
+    weight::Float64=0.5
+)::DataFrame
+    options = option_seed_preference()
+    options.probability = zeros(size(options, 1))
+    valid_locs = collect(1:size(loc_data, 1))
+
+    if ports == nothing
+        ports = _ports()
+    end
+
+    past_locations = ADRIA.select_locations(
+        options[options.option_name .== past_option, :preference][1], decision_matrix,
+        mcda_method, valid_locs, min_locs
+    )
+
+    for row in eachrow(options)
+        option_locations = ADRIA.select_locations(
+            row.preference, decision_matrix, mcda_method, valid_locs, min_locs
+        )
+        row.probability += (1 - cost_index(option_locations, loc_data, ports)) * weight
+        row.probability +=
+            option_similarity(option_locations, past_locations, loc_data) * (1 - weight)
+    end
+
+    # Normalize probabilities
+    sum_probs = sum(options.probability)
+    options.probability = options.probability / sum_probs
+
+    return options[:, [:option_name, :probability]]
+end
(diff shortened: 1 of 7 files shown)


src/analysis/pathway_diversity.jl
```diff
@@ -0,0 +1,211 @@
...
+    :abbot_point => (148.0948982, -19.9220318),
+    :hay_point => (149.2737503, -21.2934747),
+    :gladstone => (151.2993586, -23.8740713)
+)
+
+# Weighs used in MCDA critaria for all pathway diversity options.
+# The order of the weighs follow this criteria sorting
+# :seed_heat_stress, :seed_wave_stress, :seed_in_connectivity, :seed_out_connectivity,
+# :seed_depth, :seed_coral_cover, :seed_cluster_diversity, :seed_geographic_separation
+const OPTION_WEIGHS = Dict(
```
ConnectedSystems, line 17:
```suggestion
const OPTION_WEIGHTS = Dict(
```

ConnectedSystems: Probably better to make this a function so you can easily adjust.

`const` types require the Julia session to be restarted if you need to modify.

src/analysis/pathway_diversity.jl
```diff
@@ -0,0 +1,211 @@
...
+    :townsville => (146.8332585, -19.2529659),
+    :abbot_point => (148.0948982, -19.9220318),
+    :hay_point => (149.2737503, -21.2934747),
+    :gladstone => (151.2993586, -23.8740713)
+)
+
+# Weighs used in MCDA critaria for all pathway diversity options.
+# The order of the weighs follow this criteria sorting
+# :seed_heat_stress, :seed_wave_stress, :seed_in_connectivity, :seed_out_connectivity,
+# :seed_depth, :seed_coral_cover, :seed_cluster_diversity, :seed_geographic_separation
```
ConnectedSystems, line 16:
```suggestion
# Weights used in MCDA criteria for all pathway diversity options.
# The order of the weights follow this criteria sorting
# :seed_heat_stress, :seed_wave_stress, :seed_in_connectivity, :seed_out_connectivity,
# :seed_depth, :seed_coral_cover, :seed_cluster_diversity, :seed_geographic_separation
```

Looking at this though, wouldn't a dataframe or YAXArray be more appropriate?

ConnectedSystems: Please look over the rest and make sure `weights` is used appropriately.

src/analysis/pathway_diversity.jl
vitorhirata: Sure, I will change that

src/analysis/pathway_diversity.jl
vitorhirata: Yes, I was considering this as well. I might change to dataframe since this will run after the simulation and the performance shouldn't be a problem

vitorhirata:
I will continue to work on this analysis at my own fork of ADRIA. We will come up with a strategy to add this analysis to the main repository on a later stage, once the analysis is more stable and can be easily used.
For now, I am keeping this analysis in this PR of my fork: https://github.com/vitorhirata/ADRIA.jl/pull/1

PR #939, opened 2025-03-26
Update docs to configure Julia Formatter extension

ConnectedSystems:
Add clearer docs on how to set up the Julia Formatter extension.

The Julia Formatter package is only relevant if running the formatting manually or in CI/CD

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,7 +1,7 @@
 style="blue"
 indent=4
 margin=92
-always_for_in=false
+always_for_in="nothing"
 for_in_replacement="∈"
 whitespace_typedefs=false
 import_to_using=true
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,10 +1,11 @@
 style="blue"
 indent=4
 margin=92
-always_for_in="nothing"
+always_for_in=false
 for_in_replacement="∈"
 whitespace_typedefs=false
 import_to_using=true
+whitespace_in_kwargs=false
 align_struct_field=true
 align_assignment=false
 align_conditional=true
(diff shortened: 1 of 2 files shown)


arlowhite (changes requested):
I would rather have the formatting rules defined in one place, the `.JuliaFormatter.toml` file. I don't think this extension adds any value beyond the main Julia VSCode extension, unless it has added more features recently.

I'm not sure why the default has changed, but set this:
{8A36651E-A5D8-4449-95D1-8C10E6B49095}

`"editor.defaultFormatter": "julialang.language-julia"`
and VSCode will start using the `.JuliaFormatter.toml` file again. No need for step 1, install JuliaFormatter extension.

ConnectedSystems:
> `"editor.defaultFormatter": "julialang.language-julia"` and VSCode will start using the `.JuliaFormatter.toml` file again. No need for step 1,2,3, instead, just need to fix this setting.

I'm not sure this is working as intended - I'm 80% certain @sammatthews990 had it set to the default Julia option and no changes were happening but let me try again.

arlowhite:
Just for the `julia` language context. aka "language mode"
image

ConnectedSystems:
> This is a bit confusing though, because `None` was not working, then I reset this defaultFormatter option and both `None` and `Julia` work for me now.

Maybe the second time you change it, VS Code needs a restart to "forget". I suspect it loads up this setting on start up.

ConnectedSystems:
Okay, for me the issue was I installed the VS Code extension.

This comment by a helpful chap on the internet resolved the issue I was seeing: https://discourse.julialang.org/t/help-does-anyone-know-how-to-use-juliaformatter-in-vscode-editor/111714/9

arlowhite (approved):
LGTM

PR #940, opened 2025-03-26
Speed up timeseries clustering by checking for already completed calculations

eco-ben:
Matrix calculation code (L91-105) copied from `EcoReefClustering.jl` repo (I tried to extend `EcoReefClustering.distance_matrix!()` function to extend to using `_complexity_invariance` as a function input, but wasn't able to complete requiring `dist_fn` as well. 

The main changes check if CID calculation has already been done for a pair of reefs, allowing for large increases in speed in `cluster_series()`.

I'm making a PR onto GBR-wide-domain branch because I am using it for GBR wide runs and clustering. I can swap and PR to main if that is more suitable.

### DIFF
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -44,6 +44,12 @@ function correction_factor(ce_i::T, ce_j::T)::Float64 where {T<:Real}
     return max(ce_i, ce_j) / min(ce_i, ce_j)
 end
 
+function _complexity_invariance((data_x, complexity_x), (data_y, complexity_y), dist_fn)
+    ed = dist_fn(data_x, data_y)
+    cf = correction_factor(complexity_x, complexity_y)
+    return ed * cf
+end
+
 """
     complexity_invariance_distance(data::AbstractMatrix{<:Real}; distance=:euclidean)::AbstractMatrix{Float64}
 
@@ -71,6 +77,7 @@ function complexity_invariance_distance(
     # Create empty Matrix
     data_size = size(data, 2)
     cid_matrix::AbstractMatrix{Float64} = zeros(data_size, data_size)
+    dist_complexity = [(data[:, i], complexity[i]) for i in 1:data_size]
 
     local weights::Vector{Float64}
     if distance == :weuclidean
@@ -81,12 +88,20 @@ function complexity_invariance_distance(
 
     #? Do we want to normalize the amplitudes of all series?
     # Iterate over data matrix to compute CID (Complexity Invariance Distance)
-    for i in axes(data, 2)
-        @floop for j in axes(data, 2)
-            ed = dist_fn(data[:, i], data[:, j])
-            cf = correction_factor(complexity[i], complexity[j])
-            cid_matrix[i, j] = cid_matrix[j, i] = ed * cf
+    for ii in axes(cid_matrix, 1)
+        Threads.@threads for jj in axes(cid_matrix, 2)
+            if ii == jj || !iszero(cid_matrix[ii, jj])
+                continue
+            end
+
+            @views cid_matrix[ii, jj] = _complexity_invariance(
+                dist_complexity[ii], 
+                dist_complexity[jj], 
+                dist_fn
+            )
         end
+
+        cid_matrix[:, ii] .= cid_matrix[ii, :]
     end
 
     return cid_matrix


src/analysis/clustering.jl
```diff
@@ -81,12 +88,20 @@ function complexity_invariance_distance(
 
     #? Do we want to normalize the amplitudes of all series?
     # Iterate over data matrix to compute CID (Complexity Invariance Distance)
-    for i in axes(data, 2)
-        @floop for j in axes(data, 2)
-            ed = dist_fn(data[:, i], data[:, j])
-            cf = correction_factor(complexity[i], complexity[j])
-            cid_matrix[i, j] = cid_matrix[j, i] = ed * cf
+    for ii in axes(cid_matrix, 1)
+        Threads.@threads for jj in axes(cid_matrix, 2)
```
Zapiano, line 92: Why replace `@floop` with `Threads.@threads`? From FLoops documentation it looks like they basically use Threads.@threads by default in this case (but please, correct me if I'm wrong).

Zapiano, line 92: Alternatively we could start `jj` at `ii + 1` and do  
```
cid_matrix[ii, jj] = cid_matrix[jj, ii] = _complexity_invariance(
                dist_complexity[ii],
                dist_complexity[jj],
                dist_fn
            )
```

With that, this check `if` statement can be removed (right?) because `jj` will  never be equal `ii`. 


Also, can we replace `axes(data, 2)` by `data_size` (which a **terrible** name, btw)?

Zapiano: Also: since in Julia allocating memory is usually worse than doing more computation, I feel like creating this `dist_complexity = [(data[:, i], complexity[i]) for i in 1:data_size]` can increase run time here - but I haven't checked. Have you?

Zapiano (changes requested):

Zapiano:
@eco-ben I forgot to mention, but in my opinion this change should target the main branch. :)

eco-ben:
> @eco-ben I forgot to mention, but in my opinion this change should target the main branch. :)

@Zapiano Thanks for the suggestions, yeah I can do that, I'll just answer your comments here, or can I move the whole PR with git rebase? 

> Also, can we replace axes(data, 2) by data_size (which a terrible name, btw)?
`axes(data, 2)` == `1:data_size` (renamed n_locs in the following code).

`I couldn't work out where the speed up was coming from`. Here are a few tests using Revise, but it seems like Threads.@threads is what is giving the speed up as well as the !iszero() if statement (?)
Do you know why that would be the case?

Using the following code: 
```julia
for ii in axes(cid_matrix, 1)
     @floop for jj in 2:n_locs

             ed = dist_fn(data[:, ii], data[:, jj])
             cf = correction_factor(complexity[ii], complexity[jj])
             @views cid_matrix[ii, jj] = cid_matrix[jj, ii] = ed * cf
      end
end

@time ADRIA.analysis.cluster_series(rand(70, 3806), 3)
 # 15.248195 seconds (58.21 M allocations: 17.844 GiB, 19.81% gc time)
```

Using the following code:
```julia
for ii in axes(cid_matrix, 1)
        @floops for jj in axes(cid_matrix, 2)
            if ii == jj || !iszero(cid_matrix[ii, jj])
                continue
            end
            
            ed = dist_fn(data[:, ii], data[:, jj])
            cf = correction_factor(complexity[ii], complexity[jj])
            @views cid_matrix[ii, jj] = cid_matrix[jj, ii] = ed * cf
        end
end

@time ADRIA.analysis.cluster_series(rand(70, 3806), 3)
# 14.504362 seconds (29.25 M allocations: 8.997 GiB, 16.52% gc time)
```

Using the following code:
```julia
for ii in axes(cid_matrix, 1)
        Threads.@threads for jj in axes(cid_matrix, 2)
            if ii == jj || !iszero(cid_matrix[ii, jj])
                continue
            end
            
            ed = dist_fn(data[:, ii], data[:, jj])
            cf = correction_factor(complexity[ii], complexity[jj])
            @views cid_matrix[ii, jj] = cid_matrix[jj, ii] = ed * cf
        end
end

@time ADRIA.analysis.cluster_series(rand(70, 3806), 3)
# 6.782402 seconds (29.24 M allocations: 8.995 GiB, 44.14% gc time)
```

Using the following code:
```julia
for ii in axes(cid_matrix, 1)
        Threads.@threads for jj in 2:n_locs            
            ed = dist_fn(data[:, ii], data[:, jj])
            cf = correction_factor(complexity[ii], complexity[jj])
            @views cid_matrix[ii, jj] = cid_matrix[jj, ii] = ed * cf
        end
end

@time ADRIA.analysis.cluster_series(rand(70, 3806), 3)
# 11.026897 seconds (58.20 M allocations: 17.843 GiB, 50.57% gc time)
```

These time improvements matter in my case because I'm looping over different groupings of reefs (e.g. management area and bioregions) and clustering series within each one, so clustering time adds up quickly. I did a bit of profiling and it seemed like complexity_invariance_distance() was taking most of the time.

Zapiano:
Closing this since #949 fixes the same problem using the same approach.

PR #947, opened 2025-04-30
Fix rounding problem, leading to coral overgrowth, when cover plus recruits fill an entire reef

Zapiano:
In the rare case where adding the recruits to a location cover should fill an entire reef, so that the relative cover would be equal to 1, sometimes there's a rounding error causing the actual relative cover to be 1 plus a very tiny number (one of the times I've checked it was close to 1e10-11). That causes a few issues and the fix is to set a threshold (right now it's set to 0.95 but that can be changed) so that:

- If that location's relative cover is already above the threshold, we set the recruits to zero.
- If that location's relative cover is below the threshold but when adding the recruits it goes above the threshold, we rescale the recruits so that this sum is equal to the threshold
- Otherwise we do nothing

And that seems to fix the problem. 

Note that:

1. To cause this bug in the first place we seeded with a very high number of seeds and a very low number of locations (1e6 seeds to 15 locations).
2. There's also a non related issue with CoralBlox' `max_projected_cover` that can cause a bug precisely in this situations where a reef is filled with corals from a small size class. This issue doesn't cause the round error but also causes a coral overgrowth and that is going  to be fixed on a CoralBlox PR. Here's the link to the issue: https://github.com/open-AIMS/CoralBlox.jl/issues/33

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -930,7 +930,7 @@ function run_model(
         C_rel_cover = loc_coral_cover(C_cover_t) ./ habitable_loc_areas
         loc_recruits_rel_cover = loc_recruits_cover(recruitment) ./ habitable_loc_areas
 
-        # Set recruitment to 0 where C_cover_t is already round_threshold
+        # Set recruitment to 0 where C_cover_t is already above round_threshold
         cover_above_threshold_mask = C_rel_cover .> round_threshold
         recruitment[:, cover_above_threshold_mask] .= 0.0
 
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -1042,8 +1042,8 @@ function run_model(
         # Calculates scope for coral fedundity for each size class and at each location
         fecundity_scope!(fec_scope, fec_all, fec_params_per_m², C_cover_t, habitable_areas)
 
-        loc_coral_cover = loc_coral_cover(C_cover_t)
-        leftover_space_m² = relative_leftover_space(loc_coral_cover) .* vec_abs_k
+        _loc_coral_cover = loc_coral_cover(C_cover_t)
+        leftover_space_m² = relative_leftover_space(_loc_coral_cover) .* vec_abs_k
 
         # Reset potential settlers to zero
         potential_settlers .= 0.0
@@ -1157,14 +1157,14 @@ function run_model(
             # Determine connectivity strength weighting by area.
             # Accounts for strength of connectivity where there is low/no coral cover
             in_conn, out_conn, _ = connectivity_strength(
-                area_weighted_conn, vec(loc_coral_cover), conn_cache
+                area_weighted_conn, vec(_loc_coral_cover), conn_cache
             )
 
             update_criteria_values!(
                 decision_mat;
                 heat_stress=dhw_projection[_valid_locs],
                 wave_stress=wave_projection[_valid_locs],
-                coral_cover=loc_coral_cover[_valid_locs],  # Coral cover relative to `k`
+                coral_cover=_loc_coral_cover[_valid_locs],  # Coral cover relative to `k`
                 in_connectivity=in_conn[_valid_locs],  # area weighted connectivities for time `t`
                 out_connectivity=out_conn[_valid_locs]
             )
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -19,6 +19,7 @@ using ADRIA.metrics:
 using ADRIA.metrics: relative_juveniles, relative_taxa_cover, juvenile_indicator
 using ADRIA.metrics: coral_evenness
 using ADRIA.decision
+using ADRIA: loc_coral_cover, loc_recruits_cover
 
 """
     setup_cache(domain::Domain)::NamedTuple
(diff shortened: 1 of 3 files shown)


Zapiano:
Just a small comment: the tests are not passing but that's because in `global-calib` branch we still need to add the coral growth acceleration parameters to the model.

src/scenario.jl
```diff
@@ -1002,7 +1041,7 @@ function run_model(
         # Calculates scope for coral fedundity for each size class and at each location
         fecundity_scope!(fec_scope, fec_all, fec_params_per_m², C_cover_t, habitable_areas)
 
-        loc_coral_cover = _loc_coral_cover(C_cover_t)
+        loc_coral_cover = loc_coral_cover(C_cover_t)
```
ConnectedSystems, line 1044: Won't this replace the method or is this a non-issue?

I can't remember if Julia can distinguish between a variable and a function name...

src/scenario.jl
```diff
@@ -960,20 +1002,17 @@ function run_model(
...
         end
 
-        if !(sum(_loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas) == 0)
-            msk = _loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas
-            C_cover_t[:, :, msk .&& habitable_locs] .*=
+        # Check if size classes are inappropriately out-growing habitable area
+        if any(loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas)
+            @warn "Cover outgrowing habitable area at tstep $tstep. Constraining."
+            msk = loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas
+            C_cover_t[:, :, habitable_locs .&& msk] .*=
```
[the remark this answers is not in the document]
Zapiano: You're right, that causes a conflict, I don't know how was that running fine before...

Zapiano:
@ConnectedSystems I've addressed all your comments, you can have another look at this when you have time.

src/scenario.jl
```diff
@@ -915,15 +917,55 @@ function run_model(
...
+                        round_threshold .*
+                        habitable_loc_areas[agg_cover_above_threshold_mask]
+                    ) .- loc_coral_cover(C_cover_t[:, :, agg_cover_above_threshold_mask])
+                ) ./ loc_recruits_cover(recruitment[:, agg_cover_above_threshold_mask])
+
+            @assert !any(recruits_scale_factor .<= 0)
+
+            # Rescale recruits in target locations to fit within error bounds
+            recruitment[:, agg_cover_above_threshold_mask] .*= repeat(
+                recruits_scale_factor', 5
```
ConnectedSystems, line 958: `5` being the number of functional groups?

There's a variable that stores that waaaaaay above this line of code (I think `n_groups`?) - I'd prefer that be used instead of a magic number that can change.

(just check that `n_groups` is actually five first, it may be a mis-named variable 😛 )

src/scenario.jl
```diff
@@ -960,20 +1005,18 @@ function run_model(
             coral_cover(functional_groups[i], @view(C_cover_t[:, :, i]))
         end
 
-        if !(sum(_loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas) == 0)
-            msk = _loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas
-            C_cover_t[:, :, msk .&& habitable_locs] .*=
+        # Check if size classes are inappropriately out-growing habitable area
+        if any(loc_coral_cover(C_cover_t)[habitable_locs] .> habitable_loc_areas)
+            @warn "Cover outgrowing habitable area at tstep $tstep. Constraining."
+            outgrowing_locs_maks =
```
ConnectedSystems, line 1011: typo in variable name `maks` -> `mask`

src/scenario.jl
```diff
@@ -1002,8 +1045,8 @@ function run_model(
         # Calculates scope for coral fedundity for each size class and at each location
         fecundity_scope!(fec_scope, fec_all, fec_params_per_m², C_cover_t, habitable_areas)
 
-        loc_coral_cover = _loc_coral_cover(C_cover_t)
-        leftover_space_m² = relative_leftover_space(loc_coral_cover) .* vec_abs_k
+        _loc_coral_cover = loc_coral_cover(C_cover_t)
```
ConnectedSystems, line 1048: I know why this name, and I know it was similar before, but personally I think in this context the names are so similar, maybe best to use a shorthand (e.g., `lcc`)? The variable is used quite close to each other so context is maintained.

But otherwise, I'm also okay with leaving it as is.

src/scenario.jl
Zapiano: I can't believe I missed this 🤦🏻

src/scenario.jl
ConnectedSystems: Yes, it's cryptic but it's used in a very small area (within 5 lines of code from each other), which is what I meant by the context is maintained.

But it's not a huge issue, I'm okay to ignore this 😄

Zapiano: Ok, I've already addressed your other comments

ConnectedSystems (approved):
Thanks!

PR #949, opened 2025-05-13
Speed up timeseries clustering

Zapiano:
This is essentially the same as #940 but I've created a separate PR because the other branch was too behind the main and a rebase was creating a few unwanted commits for some reason.

The two main changes here are:

1. Since the distance matrix is symmetric, the inner loop limits were changing to avoid going through the same matrix elements more than once
2. `@floop` was replaces by `Threads.@threads`. Not sure why, but a simple banchmark using Chairmarks showed that it actually works (in the dfigure below, the first benchmark was done using `@loops` and the second using `Threads.@threads`):
image

Credits to @eco-ben who actually found this.

### DIFF
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -72,8 +72,8 @@ function complexity_invariance_distance(
     data::AbstractMatrix{<:Real};
     distance=:euclidean
 )::AbstractMatrix{Float64}
-    # Compute complexity vector
-    complexity = _complexity(data)
+    # Compute complexity vector (each element is the complexity of a distinct scenario)
+    complexity::Vector{Float64} = _complexity(data)
 
     # Create empty Matrix
     data_size = size(data, 2)
--- a/src/analysis/clustering.jl
+++ b/src/analysis/clustering.jl
@@ -44,10 +44,17 @@ function correction_factor(ce_i::T, ce_j::T)::Float64 where {T<:Real}
     return max(ce_i, ce_j) / min(ce_i, ce_j)
 end
 
-# function _complexity_invariance((data_x, complexity_x), (data_y, complexity_y), dist_fn)
-function _complexity_invariance(data, complexity, i, j, dist_fn)::Float64
-    ed = dist_fn(data[:, i], data[:, j])
-    cf = correction_factor(complexity[i], complexity[j])
+"""
+    _complexity_invariance(data_i, data_j, complexity_i, complexity_j, dist_fn)::Float64
+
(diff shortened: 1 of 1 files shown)


src/analysis/clustering.jl
```diff
@@ -44,6 +44,13 @@ function correction_factor(ce_i::T, ce_j::T)::Float64 where {T<:Real}
     return max(ce_i, ce_j) / min(ce_i, ce_j)
 end
 
+# function _complexity_invariance((data_x, complexity_x), (data_y, complexity_y), dist_fn)
```
ConnectedSystems, line 47: If not needed, remove commented function definition

ConnectedSystems:
> @floop was replaces by Threads.@threads. Not sure why, but a simple banchmark using Chairmarks showed that it actually works (in the dfigure below, the first benchmark was done using @loops and the second using Threads.@threads):

Yeah, I suspect @floops does some extra work behind the scenes to ensure compatibility with iterators that the base `@threads` macro does not support. In this case there are no "special" data structures or iterators to worry about so the `@threads` actually performs better as it does not do these additional checks.

So lesson learnt - if doing simple threaded tasks, don't be lazy and type out the full `Threads.@threads` instead of `@floops`.

Approving subject to two minor comments.

Zapiano:
FYI @eco-ben

src/analysis/clustering.jl
```diff
@@ -89,14 +89,14 @@ function complexity_invariance_distance(
...
+        weights = sqrt.(1 ./ (1:n_timesteps))
     end
     dist_fn(x, y) = (distance == :euclidean) ? euclidean(x, y) : weuclidean(x, y, weights)
 
     #? Do we want to normalize the amplitudes of all series?
     # Iterate over data matrix to compute CID (Complexity Invariance Distance)
-    for i in 1:n_timesteps
-        Threads.@threads for j in (i + 1):n_timesteps
+    for i in 1:n_scenarios
+        Threads.@threads for j in (i + 1):n_scenarios
```
ConnectedSystems, line 99: Sorry @Zapiano I just noticed - are these changes correct?

They may be weighting and incorrectly comparing these changes to the old implementation?

Zapiano: I think it's easier if we look at the aggregated diff after this entire PR (print below). I just replaced `size(data,1)` with `n_timesteps` and `size(data,2)` with `n_scenarios` where `n_timesteps, n_scenarios = size(data)`. The mistake that I mentioned in the previous comment (and that was fixed before merging) is that I've switched these two by accident. 

image

PR #950, opened 2025-05-15
add bin edges used in calibration to ADRIA

DanTanAtAims:
When eventually merged, the global-calib branch should have the size classes that are used in the calibration.

See https://github.com/ConnectedSystems/ltmp_calibration/blob/ecorrap/src/common/common.jl#L17-L21


Zapiano (approved):
LGTM

PR #951, opened 2025-05-15
Fix multiple tests

Zapiano:
A few tests were broken on the main branch and this PR fixes all of them - at least locally they are passing now.

### DIFF
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -208,13 +208,13 @@ n_locations)
 - `coral_spec` : Coral spec DataFrame
 """
 function _relative_juveniles(
-    X::AbstractArray{T,3}, coral_spec::DataFrame
-)::AbstractArray{T,2} where {T<:Real}
+    X::YAXArray{T,3}, coral_spec::DataFrame
+)::YAXArray{T,2} where {T<:Real}
     # Cover of juvenile corals (< 5cm diameter)
-    juv_groups =
-        X[species=coral_spec.class_id .== 1] .+ X[species=coral_spec.class_id .== 2].data
-
-    return dropdims(sum(juv_groups; dims=:species); dims=:species)
+    return dropdims(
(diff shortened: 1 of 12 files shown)


src/metrics/metrics.jl
```diff
@@ -208,13 +208,13 @@ n_locations)
...
+)::YAXArray{T,2} where {T<:Real}
     # Cover of juvenile corals (< 5cm diameter)
-    juv_groups =
-        X[species=coral_spec.class_id .== 1] .+ X[species=coral_spec.class_id .== 2].data
-
-    return dropdims(sum(juv_groups; dims=:species); dims=:species)
+    return dropdims(
+        sum(X[species=coral_spec.class_id .== 1]; dims=:species); dims=:species
+    ) .+
+           dropdims(sum(X[species=coral_spec.class_id .== 2]; dims=:species); dims=:species)
```
ConnectedSystems, line 217: I prefer the older approach with `AbstractArray` because sometimes it is handy to call `relative_juveniles` directly (e.g., with results loaded from a CSV from ReefMod).

ConnectedSystems (commented):
One comment about supporting alternate use case.

Once we come to an agreement on that, I think this can be merged.

FYI, I'm also thinking of splitting out the metrics into a separate sub-package.
It would be useful for others to be able to import those metrics without having to load the entire ADRIA ecosystem.

src/metrics/metrics.jl
Zapiano: You can still call this metric passing an Array (non-YAXArray). What happens is that this array is always converted to a YAXArray before this function is actually called (https://github.com/open-AIMS/ADRIA.jl/blob/9dbb94a20a27d280b7bb736eefa5c5edf37b8b4a/src/metrics/metrics.jl#L47)

ConnectedSystems (approved):

PR #952, opened 2025-05-16
Changes to enable recent RME domain loading

Zapiano:
Tested with "rme_ml_2023_03_30b", "rme_ml_2024_01_08" and the most recent on "rme_ml_2025_02_11" and it works with all of them.


src/ExtInterface/ReefMod/RMEDomain.jl
```diff
@@ -258,6 +265,22 @@ function _get_relevant_files(fn_path::String, ident::String)
...
+function _data_file_path(data_path, file_name; file_extension="csv")
+    data_file_names = readdir(data_path)
+    files_mask = [occursin(Regex(file_name), file) for file in data_file_names]
+    file_names = data_file_names[files_mask]
+    file_name = if length(file_names) == 1
+        file_names[1]
+    else
+        dhw_csv_mask = [
+            occursin(Regex(file_extension), csv_file) for csv_file in file_names
+        ]
```
ConnectedSystems, line 277: I prefer using globs over regexs partly for safety, partly for readability (but how it's used here seems readable).

I also prefer including the "." in extensions (e.g., `.csv`) - because this will match against any text in the filename.

src/ExtInterface/ReefMod/RMEDomain.jl
```diff
@@ -124,17 +124,24 @@ function load_domain(::Type{RMEDomain}, fn_path::String, RCP::String)::RMEDomain
...
+    end
+    cluster_id_col = reef_id_col
 
     reef_ids::Vector{String} = spatial_data[:, reef_id_col]
 
     # Load accompanying ID list
     # TODO: Create canonical geopackage file that aligns all IDs.
     #       Doing this removes the need for the manual correction below and removes the
     #       dependency on this file.
+    id_file = first(readdir(joinpath(data_files, "id")))
```
ConnectedSystems, line 142: Suggest being more targeted with the substring (`id_` or `id_list`?) to protect against unwanted coincidental matches.

ConnectedSystems (changes requested):
Minor suggestions to protect against coincidental text matches but otherwise looks good.

Thanks for doing this, it had been on my todo list for a long time.

src/ExtInterface/ReefMod/RMEDomain.jl
Zapiano: I think the function and variable names were misleading because here we are after a folder with "csv" in the name, not a file - so the extension won't help.

src/ExtInterface/ReefMod/RMEDomain.jl
Zapiano: Now I'm using Glob to better match the name and extension of the file.

ConnectedSystems (approved):

PR #955, opened 2025-05-22
Add plot for single scenario stratified by size class and functional group

Zapiano:


src/viz/viz.jl
```diff
@@ -15,6 +15,9 @@ function explore() end
 function scenarios() end
 function scenarios!() end
 
+# Plot a single scenario
```
ConnectedSystems, line 18: Single scenario of what?

ext/AvizExt/viz/scenarios.jl
```diff
@@ -189,6 +189,80 @@ function ADRIA.viz.scenarios!(
...
+# Plot scenario cover stratified by size class and functional group
+ADRIA.viz.scenario_stratified(reshape(m_rs.raw, n_timesteps, n_sizes, n_groups, n_locations))
+
+# Plot scenario bleaching mortality stratified by size class and functional group
+ADRIA.viz.scenario_stratified(
+    permutedims(m_rs.bleaching_mortality, (1,3,2,4));
+    axis_opts=Dict{Symbol,Any}(:limits=(nothing, (0.0, 1.0)))
+)
+
+# Plot scenario cover
```
ConnectedSystems, line 222: This seems out of place? Or the example is unfinished?

ext/AvizExt/viz/scenarios.jl
```diff
@@ -189,6 +189,80 @@ function ADRIA.viz.scenarios!(
...
+```
+"""
+function ADRIA.viz.scenario_stratified(
```
ConnectedSystems, line 225: I don't think `scenario_stratified` is a good name - there are many ways a scenario can be stratified - but I also understand it's hard to come up with a name that makes sense and isn't overly verbose.

Best alternative I can think of is `scenario_by_taxa` which at least indicates what will be shown.

ext/AvizExt/viz/scenarios.jl
```diff
@@ -189,6 +189,80 @@ function ADRIA.viz.scenarios!(
...
+"""
+    scenario_stratified(data::AbstractArray{Float64, 4})
+
+Plot a single scenario, averaged across all locations, stratified by size class and
+functional group.
+
+# Arguments
+- `data` : Either a 4D array comprising (timesteps, size_classes, functional_groups and
+locations) or a 3D array comprising (timesteps, size_groups, locations).
+- `axis_opts` : Opts to be passed to Makie.Axis.
```
ConnectedSystems, line 202:
```suggestion
- `axis_opts` : Opts to be passed to `Makie.Axis()`.
```

ext/AvizExt/viz/scenarios.jl
```diff
@@ -189,6 +189,80 @@ function ADRIA.viz.scenarios!(
...
+# TODO Add support for YAXArrays (`Makie.lines!` expects an Array). For now this works.
+"""
+    scenario_stratified(data::AbstractArray{Float64, 4})
+
+Plot a single scenario, averaged across all locations, stratified by size class and
+functional group.
+
+# Arguments
+- `data` : Either a 4D array comprising (timesteps, size_classes, functional_groups and
+locations) or a 3D array comprising (timesteps, size_groups, locations).
```
ConnectedSystems, line 201: It states 3D or 4D array but the function signature specifies it must be 4D?

ConnectedSystems (changes requested):
A few minor comments, thanks.

ext/AvizExt/viz/scenarios.jl
```diff
@@ -189,6 +189,80 @@ function ADRIA.viz.scenarios!(
...
+        :limits,
+        (nothing, (nothing, ceil(maximum(mean_data); digits=2)))
+    )
+    ylabel = pop!(axis_opts, :ylabel, "")
+
+    # This isn't technically an axis parameter because this title is going to be applied on
+    # as a Label to the whole figure, but left it like this for simplicity
+    title = pop!(axis_opts, :title, "")
+
+    for fg in 1:n_groups
```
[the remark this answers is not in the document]
Zapiano: I made this function accept a Array, instead of a YAXArray, because of a few operations that are different with both (like taking the `mean`\`median` passing the axis name or calling `lines!`, that expects an Array). I think it would be nice to accept both (either having two methods one for each of making this one more flexible somehow) but since this is meant for debugging purposes (for now) and the time limit I decided to go only with Arrays for now.

ext/AvizExt/viz/scenarios.jl
Zapiano: I also don't like `scenario_stratified` very much. I don't like `scenario_by_taxa` because it plots by functional group and by size class. And, in the future, this could have a flag so the user decides if they want it plotted by functional group, by size class or both. Alternatively 

1. We could simple call this `scenario_by_size_and_group` (I know, long name...) and in the future have a `scenario_by_group` and a `scenario_by_size`. 
2. We could call it `scenario` and have a flag `group_by` that can be `size_and_group`, `size`, `group` or `nothing`.

PR #957, opened 2025-06-17
Update Manifest

Zapiano:
In particular:
- Remove DiskArrays
- Update CoralBlox to 1.1.0

DiskArrays was added just so it could be pinned to fix an issue with a previous version. That was solved on a new version but to use that we needed to update YAXArrays and that was being blocked by CoralBlox. With the new CoralBlox version we were able to update YAXArrays and now DiskArrays is no longer needed.

Tests are passing locally.


DanTanAtAims (approved):
After trying `]instantiate` and `resolve` again the tests past.

PR #965, opened 2025-07-23
Fix tests prior to merging `global-calib` to `main`

Zapiano:
Context: The way the new factors, unique for each `CB_CALIB_GROUP`, were implemented was by adding them to the model when the domain is loaded. When sampling, we filter the factors that correspond to `CB_CALIB_GROUP`s not present (for example, when using ADRIA domain, all reefs belong to the same `CB_CALIB_GROUP`, so the factors relative to all other `CB_CALIB_GROUP`s are not inputs to the model in this case). Because of that, when we load ADRIA domain and generate scenarios, the model spec and the param_set will have different factors/params, and that may cause an issue in some places. 

This PR: The biggest change introduced in this PR is the addition of a few `model_spec` methods that remove these unused factors so we can continue having a `model_spec` and `param_set` with the same factors. 

When this is merged, the branch `global-calib` will be ready for a final review and merge.


DanTanAtAims (approved):
Tests passing. Looks good.

PR #967, opened 2025-08-12
Calibration and SA changes

Zapiano:
This implements a series of changes in ADRIA, some that directly affect the model, as a result of the calibration and sensitivity analysis processes. Those changes include:

- The loaded spatial geopackage is now  expected to have a `CB_CALIB_GROUPS` integer column assigning each location to a `CB_CALIB_GROUPS`.
- New "families" of factors/parameters specific to each `CB_CALIB_GROUP` were introduced: `mb_rate_scale_cb_group`, `linear_extension_scale_cb_group`, `growth_acceleration_cb_group_midpoint`, `growth_acceleration_cb_group_height`, `growth_acceleration_cb_group_steepness`
- Include recruits in the spatial competition 'max_projected_cover' 

# `global-calib` vs `main` comparison

n_scenarios = 1024

## RME Domain
| ---- | `calib` | `main` |
| ---- | ------- | ------ |
| Total Cover | rme__calib__stc | rme__main__stc |
| Shelter Vol. |  rme__calib__srsv | rme__main__srsv |
| Juveniles | rme__calib__sj | rme__main__sj |

## Moore Domain
| ---- | `calib` | `main` |
| ---- | ------- | ------ |
| Total Cover | moore__calib__stc | moore__main__stc |
| Shelter Vol. | moore__calib__srsv | moore__main__srsv |
| Juveniles | moore__calib__sj | moore__main__sj |

PR #970, opened 2025-08-22
Fix conn data load location selection when force_single_reef is true

Zapiano:
This was introduced in PR#967 as suggested in the review [1].

[1] https://github.com/open-AIMS/ADRIA.jl/pull/967#discussion_r2271792525


ConnectedSystems (approved):
LGTM

PR #985, opened 2025-10-08
Update calibrated parameter base values

Zapiano:
Update with parameters resulted from that calibration run on 01/08/2025.


Zapiano:
@ConnectedSystems all tests passing here.

PR #988, opened 2025-10-23
Fix factor update bug

ConnectedSystems:
In some situations it is possible for the factor details passed into `set_factor_bounds!()` to become misaligned.

The approach converts the passed in arguments into an OrderedDict which ensures ordered alignment of factor names and associated values.


ConnectedSystems:
@DanTanAtAims this is in draft as I'm still doing some final checks. Will let you know when ready for review 😄

ConnectedSystems:
@DanTanAtAims ready for review

PR #992, opened 2025-11-06
fix: Correct growth constraint calculation in scenario.jl

DanTanAtAims:
This pull request addresses a bug in the calculation of growth constraints for locations with low or zero coral cover.

Previously, `NaN` values were generated when locations with zero coral cover were passed to the `linear_extension_scale_factors` function. While correcting this, a more significant issue was discovered: the growth constraints for locations with low cover were not being correctly calculated by the `growth_acceleration` function.

This was due to a bug in the loop that iterates over calibration groups. The location mask used to identify locations with low cover was being mutated within the loop. As a result, the growth acceleration was only applied to locations in the first calibration group, and the growth constraints for locations in other calibration groups were not updated.

**Corrected Masking:** The loop that calculates growth constraints now uses a non-mutating approach to create the location mask for each calibration group. This ensures that the growth acceleration is applied correctly to all locations with low cover, regardless of their calibration group.
**Conditional Growth:** `linear_extension_scale_factors` are now calculated only for locations where coral cover meets a specified threshold, previously they were calculated and then overwritten. Similarly, `growth_acceleration` is calculated only for locations where coral cover is below that threshold.

These changes ensure that the growth constraints are calculated correctly for all locations, preventing `NaN` values and ensuring the model behaves as expected for locations with low or zero coral cover.

### DIFF
--- a/src/scenario.jl
+++ b/src/scenario.jl
@@ -220,7 +220,7 @@ function run_scenarios(
                 # Switch RCPs so correct data is loaded
                 dom = switch_RCPs!(dom, rcp)
                 target_rows = findall(
-                    scenarios_matrix[factors = At("RCP")] .== parse(Float64, rcp)
+                    scenarios_matrix[factors=At("RCP")] .== parse(Float64, rcp)
                 )
                 scen_args = _scenario_args(dom, scenarios_matrix, rcp, length(target_rows))
 
@@ -272,7 +272,7 @@ function growth_acceleration(
 end
 
 function _scenario_args(dom, scenarios_matrix::YAXArray, rcp::String, n::Int)
-    target_rows = findall(scenarios_matrix[factors = At("RCP")] .== parse(Float64, rcp))
+    target_rows = findall(scenarios_matrix[factors=At("RCP")] .== parse(Float64, rcp))
     rep_doms = Iterators.repeated(dom, n)
     return zip(
         rep_doms, target_rows, eachrow(scenarios_matrix[target_rows, :])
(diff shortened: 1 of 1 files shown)


src/scenario.jl
```diff
@@ -220,7 +220,7 @@ function run_scenarios(
                 # Switch RCPs so correct data is loaded
                 dom = switch_RCPs!(dom, rcp)
                 target_rows = findall(
-                    scenarios_matrix[factors=At("RCP")] .== parse(Float64, rcp)
```
Zapiano: I think this is not following the styleguide.

src/scenario.jl
```diff
@@ -860,33 +861,35 @@ function run_model(
             cache_habitable_max_projected_cover .+
             dropdims(sum(recruitment; dims=1); dims=1)
 
+        relative_habitable_cover = zeros(n_locs)
```
Zapiano: Would it be better to create this outside the loop?

src/scenario.jl
```diff
@@ -810,7 +810,8 @@ function run_model(
     FLoops.assistant(false)
     habitable_loc_idxs = findall(habitable_locs)
 
-    apply_growth_acc_mask::BitVector = trues(n_locs)
+    cover_threshold_mask::BitVector = trues(n_locs)
+    current_mask::BitVector = falses(n_locs)
```
Zapiano: I find these names `cover_theshold_mask` and `current_mask` a bit generic. Specially `current_mask`. Maybe `growth_mask_cache` and `growth threshold_mask_cache`..?

Zapiano:
@vitorhirata FYI

src/scenario.jl
Zapiano: Two questions:

1. Shouldn't `growth_constraints` be initialized as a `ones(Float64, n_locs)` vector instead? 
2. Since we use the same vector `growth_constraints` for all timesteps, shouldn't we do a `growth_constraints .= 1.0` at the end or beginning of each timestep?

Zapiano (commented):
I just have two questions before approving this.

src/scenario.jl
DanTanAtAims: From my understanding, theoretically the growth constraints values should be completely overwritten every loop, being replaced with either the linear extension scaling or the growth acceleration?

PR #996, opened 2025-11-26
Refactor out recruitment constraint from timestep loop

Zapiano:
As in title. This is just to improve readability of `run_model`.


Zapiano:
@ConnectedSystems just a quick comment: the tests are not passing on the main branch right now, but I assert that the model is running fine locally.

PR #998, opened 2025-11-27
Update GeoDataFrame handling

ConnectedSystems:
Switch to using handlers from GeoInterface.jl to identify columns with geometries.


DanTanAtAims (approved):
Works as intended. Thanks.

PR #1011, opened 2025-12-11
Improves scenario assessment workflow

ConnectedSystems:
Improves the scenario assessment workflow by adding a scenario screening method and increasing the number of representative DHW trajectories considered during sampling.
Normalizes data before applying rules in Pareto optimal identification and screening.

Introduces a `screen_scenarios()` method to filter scenarios based on specified criteria. This can be then passed in to identify scenarios that meet criteria for use in rule induction, scenario discovery, classification, etc.

Also made a slight tweak to scenario set sampling, expands the range for clustering scenarios to a maximum of 50 (previously 20)


ConnectedSystems:
@Zapiano for awareness

PR #1020, opened 2026-01-09
Mention use of shared depot paths

ConnectedSystems:
Give guidance on shared depot paths in HPC/parallel contexts.


ConnectedSystems:
Merging this in @Zapiano @DanTanAtAims

PR #1027, opened 2026-01-23
Improve fecundity

Zapiano:
Instead of using a cache Matrix, saving the results to that and then summing up in on of the dimensions and writing the sum to a matrix entry, this adds directly to the final matrix, decreasing the number of allocations needed. 

I've run the model two times, before and after, to be sure the raw results are exactly the same.

I've also use Chairmarks to do a naive benchmark, just to have an idea of the improvement. The following is for a single `run_model` call:
```
## Previously
9.890 s (5060673 allocs: 2.831 GiB, 6.76% gc time, without a warmup)

## Now
6.659 s (5060363 allocs: 2.819 GiB, 8.64% gc time, without a warmup)
```

### DIFF
--- a/src/ecosystem/corals/growth.jl
+++ b/src/ecosystem/corals/growth.jl
@@ -570,10 +570,6 @@ function settler_DHW_tolerance!(
 end
 
 """
-    fecundity_scope!(fec_groups::Array{Float64, 2}, fec_all::Array{Float64, 2},
-                     fec_params::Array{Float64}, C_cover_t::Array{Float64, 2},
-                     k_area::Array{Float64})::Nothing
-
     fecundity_scope!(
         fec_groups::AbstractMatrix{T},
         fec_params::AbstractMatrix{T},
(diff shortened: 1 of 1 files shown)


DanTanAtAims:
The results seems to match with main

```console
Base Commit (2751001d0cb083aa00f04fc7dc0c15cf5f99248c) Time: 68.1209s
Commit c6b7f3b41b4013fb4f30f4e4527be123ff45b396 vs Base 2751001d0cb083aa00f04fc7dc0c15cf5f99248c: MATCH. Time: 40.0541s
Commit 1fb691461239260501b58d98b8892655dfd0c486 vs Base 2751001d0cb083aa00f04fc7dc0c15cf5f99248c: MATCH. Time: 29.1367s
Commit a570f2378b5bb511d64112afe6ef5acbc9428065 vs Base 2751001d0cb083aa00f04fc7dc0c15cf5f99248c: MATCH. Time: 38.2289s
```

DanTanAtAims (approved):
Looks good to me. Results align with main. After #1034 is merged, this branch will need to be rebased and then merged.

PR #1031, opened 2026-01-29
Add Allee effect

Zapiano:
Implementation follows Kevin's model (which is base on Mumby, 2025). The plots below present a comparison for Moore Reef cluster locations with and without Allee Effect. 

| Without heat stress | With heat stress |
| ------------------- | ---------------- |
| allee_effect_no_heat | allee_effect_heat |

Should be merged after #1027 .


Zapiano:
@ConnectedSystems This is ready to be merged (once approved).

PR #1037, opened 2026-02-06
Fix sampling tests (and others)

Zapiano:
- Fix bug caused by mismatch between color types (`RGBA{Float32}` and `RGBA{Float64}`)
- Adds calls to GC in `test/data_loading.jl`. An error was occurring there, likely caused by unclosed file handles between calls. 
- Fix `test/sampling.jl` after introduction of reactive intervention strategy parameters. The sampling method sometimes changes the values of these parameters to zero, even when though some of them have lower sampling bound greater than zero. In those cases, these parameters are not being tested.

### DIFF
--- a/test/sampling.jl
+++ b/test/sampling.jl
@@ -28,7 +28,11 @@ end
         not_cw_mask =
             ms.component .∉
             [("SeedCriteriaWeights", "FogCriteriaWeights", "decision.DepthThresholds")]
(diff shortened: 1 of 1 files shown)


ConnectedSystems:
I suggest adding a prefix to the "TEST" flag as these are environment variables and cam be set by any program. Leads to confusion/conflict. Other env flags start with "ADRIA_"

PR #1074, opened 2026-04-14
Fix combine_results

Zapiano:
As in title.


Zapiano:
@ConnectedSystems FYI

PR #1076, opened 2026-04-14
bump ADRIAIndicators version.

DanTanAtAims:
Use Tabular Acropora of diameter 95cm for maximum shelter volume in the calculations of relative shelter volume.

### DIFF
--- a/src/metrics/metrics.jl
+++ b/src/metrics/metrics.jl
@@ -11,7 +11,7 @@ using
 
 using YAXArrays
 using ADRIA:
-    DataCube, ZeroDataCube, axes_names, axis_labels, axis_index
+    DataCube, ZeroDataCube, axes_names, axis_labels, axis_index, functional_group_names
 using ADRIA: n_sizes, group_indices
 
 using FLoops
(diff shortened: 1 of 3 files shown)


Zapiano:
@DanTanAtAims it looks like the checks have failed.

src/metrics/metrics.jl
```diff
@@ -773,13 +777,19 @@ function _relative_shelter_volume(
     RSV::YAXArray = ZeroDataCube(
         (:timesteps, :groups, :sizes, :locations, :scenarios), size(X), X.properties
     )
+
+    tab_acro_idx::Int64 = findfirst(functional_group_names() .== :tabular_Acropora)
+    reference_params::Tuple{Float64, Float64, Float64} = (
+        95.0, pa_params[tab_acro_idx, n_sizes, 1], pa_params[tab_acro_idx, n_sizes, 2]
```
Zapiano, line 783: @DanTanAtAims 
I understand this `95.0` is the mean diameter? 
We should document this (at the minimum, add a comment) and where this particular value comes from.

Zapiano (approved):

src/metrics/metrics.jl
Zapiano: I've created an issue to deal with this on a separate PR: #1078

PR #1079, opened 2026-04-19
Split intervention capacity

Zapiano:
Allow splitting aquaculture (seed) and larval methods (mc) intervention capacities across different sets of locations. 

The figure below validates this new feature for both interventions with capacity split 40% to reefs on the north and 60% on the south. In each row, the left panel shows a locations (dots) with dot size proportional to the mean number of corals deployed across scenarios over all timesteps and functional groups. The right panel shows, for each scenario, a stacked bar of the total corals actually logged, alongside a black diamond marking the input budget (N_seed × seed_years or N_mc_settlers × mc_years). Dashed reference lines indicate the expected per-set budget share (weight × total budget).

image


Zapiano:
@ConnectedSystems FYI

PR #1086, opened 2026-05-07
Outcome mapping documentation

josephguillaume:
As part of work on scenario discovery, I've reviewed and added interpretation to the outcome mapping functionality.

Corrections:

- The method does not actually calculate factor regions. Through the use of bootstrapping, it actually calculates the confidence interval of the mean.
https://github.com/open-AIMS/ADRIA.jl/blob/341a867a7c99e41b4c6058d44f7c7584fd9575e6/src/analysis/sensitivity.jl#L805
- The example does not use percentiles. Min-max normalisation is used but the result is just the same variable rescaled to [0,1]. These do not represent percentiles of the underlying data points.

Clarifications

- Added API description of missing arguments
- Added `s_tac` and `foi` definition based on the original test
https://github.com/open-AIMS/ADRIA.jl/blob/d0ecf31c835896f81f94a9f20091da44d2434ef3/test/analysis.jl#L298

Added
- Emphasised selection of outcome and factors of interest as a key step in the scenario discovery process
- Interpretation in terms of scenario discovery

Out of scope

- Rerunning the example. Mismatched figure labels are therefore not fixed.
- Real world example - I have emphasised that this is based on the test data.

### DIFF
--- a/docs/src/usage/analysis.md
+++ b/docs/src/usage/analysis.md
@@ -536,17 +536,35 @@ save("rsa.png", rsa_fig)
 
 ### Outcome mapping
 
-A monte-carlo filtering approach similar to Regional Sensitivity Analysis.
+As the name implies, outcome mapping aims to identify the relationship between model outputs and the region of factor space that led to those outputs.
+Similarly to Regional Sensitivity Analysis, it does this by filtering scenarios that match a specific outcome.
 
-As the name implies, outcome mapping aids in identifying the relationship between model
-outputs and the region of factor space that led to those outputs.
+This implementation then calculates the mean value of an outcome as a function of a factor.
+This is a form of scenario discovery that summarises scenarios matching an outcome in terms of values of a factor and the average value of one outcome.
+
+This example aims to identify DHW and wave scenarios and interventions which lead to top outcomes for coral cover. Note that it uses a test dataset rather than real data.
 
 ```julia
+# As outcome, we are looking at total coral cover, averaged over time
+s_tac = ADRIA.metrics.scenario_total_cover(rs)
 mean_s_tac = dropdims(mean(s_tac, dims=1), dims=1)
+
+# Factors of interest to investigate
+foi = [
+    :dhw_scenario, # DHW scenarios 1-50 in the test dataset
+    :wave_scenario, # Wave scenarios 1-50 in the test dataset
+    # Interventions
+    :N_seed_TA, # Number of seeded Tabular Acropora deployed per intervention event
+    :N_seed_CA, # Number of seeded Corymbose Acropora deployed per intervention event
+    :fogging, # Fogging effectiveness on a scale of 0-1
+    :SRM # Reduction in DHW obtained by shading
+]
 
 tf = Figure(size=(1600, 1200))  # size of figure
 
-# Indicate factor values that are in the top 50 percentile
+# Indicate factor values that are in the top half of the range
+#   mean_s_tac is normalised to [0,1], so 0.5 is half-way between the minimum and maximum values
+# Split each factor into 20 bins
 tac_top_50 = ADRIA.sensitivity.outcome_map(rs, mean_s_tac, x -> any(x .>= 0.5), foi; S=20)
 ADRIA.viz.outcome_map!(
     tf[1, 1],
@@ -556,7 +574,7 @@ ADRIA.viz.outcome_map!(
     axis_opts=Dict(:title => "Regions which lead to Top 50th Percentile Outcomes", :ylabel => "TAC [m²]")
 )
 
-# Indicate factor values that are in the top 30 percentile
+# Indicate factor values that are in the top 30% of the range
 tac_top_30 = ADRIA.sensitivity.outcome_map(rs, mean_s_tac, x -> any(x .>= 0.7), foi; S=20)
 ADRIA.viz.outcome_map!(
     tf[2, 1],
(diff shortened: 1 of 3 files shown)


docs/src/usage/analysis.md
```diff
@@ -536,17 +536,35 @@ save("rsa.png", rsa_fig)
...
 
 ```julia
+# As outcome, we are looking at total coral cover, averaged over time
+s_tac = ADRIA.metrics.scenario_total_cover(rs)
 mean_s_tac = dropdims(mean(s_tac, dims=1), dims=1)
 
+# Factors of interest to investigate
+foi = [
+    :dhw_scenario, # DHW scenarios 1-50 in the test dataset
+    :wave_scenario, # Wave scenarios 1-50 in the test dataset
```
Zapiano, line 555:
```suggestion
    :dhw_scenario, # DHW scenario indexes
    :wave_scenario, # Wave scenarios indexes
```
Number of DHW or wave scenarios depends on the domain data loaded, so better to be less specific

docs/src/usage/analysis.md
Zapiano, line 584:
```suggestion
    axis_opts=Dict(:title => "Regions that lead to outcomes higher than 30%", :ylabel => "TAC [m²]"))
```

docs/src/usage/analysis.md
```diff
@@ -556,7 +574,7 @@ ADRIA.viz.outcome_map!(
     axis_opts=Dict(:title => "Regions which lead to Top 50th Percentile Outcomes", :ylabel => "TAC [m²]")
```
Zapiano, line 574:
```suggestion
    axis_opts=Dict(:title => "Regions that lead to outcomes higher than 50%", :ylabel => "TAC [m²]")
```

docs/src/usage/analysis.md
```diff
@@ -536,17 +536,35 @@ save("rsa.png", rsa_fig)
...
+    :N_seed_TA, # Number of seeded Tabular Acropora deployed per intervention event
+    :N_seed_CA, # Number of seeded Corymbose Acropora deployed per intervention event
+    :fogging, # Fogging effectiveness on a scale of 0-1
+    :SRM # Reduction in DHW obtained by shading
+]
+
 tf = Figure(size=(1600, 1200))  # size of figure
 
-# Indicate factor values that are in the top 50 percentile
+# Indicate factor values that are in the top half of the range
```
Zapiano, line 565:
```suggestion
# Indicate factors that lead to an outcome (`mean_s_tac`) is at or above 50%
```

docs/src/usage/analysis.md
```diff
@@ -556,7 +574,7 @@ ADRIA.viz.outcome_map!(
     axis_opts=Dict(:title => "Regions which lead to Top 50th Percentile Outcomes", :ylabel => "TAC [m²]")
 )
 
-# Indicate factor values that are in the top 30 percentile
+# Indicate factor values that are in the top 30% of the range
```
Zapiano, line 577:
```suggestion
# Indicate factors that lead to an outcome (`mean_s_tac`) is at or above 30%
```

docs/src/usage/analysis.md
```diff
@@ -570,6 +588,40 @@ save("outcome_map.png", tf)
 
 ![Outcome mapping](../assets/imgs/analysis/outcome_map.png)
 
+The figure shows the mean values of total coral cover (TAC) obtained as a function of each factor value, across scenarios in which outcomes in the top 50% or 30% of the range are obtained. The ribbon shows the estimated 95% confidence interval around the mean.
```
Zapiano, line 591: When I read "top 50% or 30%" it sounds like we're talking about 50% or 70% percentile, when in fact we are plotting all scenarios that have outcome above or equal 50% or 30%. I think it would be better to make this more clear.

docs/src/usage/analysis.md
```diff
@@ -570,6 +588,40 @@ save("outcome_map.png", tf)
...
+There is less uncertainty in the mean for the top 30% because there were fewer scenarios in this group. The top 30% scenarios are a subset of the scenarios in the top 50%.
+
+While this example uses a test data package rather than real data, we can still interpret the results:
+
+- DHW scenario:
+  - These are categorical values representing 50 different scenarios. There is no specific order to the scenarios - they are just interpreted individually.
+  - Note that as `S=20`, only 20 bins are shown - multiple scenarios have been aggregated.
+  - All 50 scenarios are represented in the top 50% of the range - it is possible to get results in the top half of the range in all scenarios. Coral cover is higher in scenarios ~20-25, ~35 and ~40, and lower in scenarios ~15 and ~45. To interpret this, we would need to look at the definition of those scenarios.
+  - In the top 30% of the range, only some scenarios are represented. For the DHW scenarios that are missing, none of the combinations of factors sampled are able to achieve results in the top 30% of the range.
+- Wave scenario: categorical variable representing each scenario in the test data package
```
Zapiano, line 602: Since this is a more general info, maybe it would be better to add a section in the beginning of the Analysis with the list of factors and a brief explanation (something similar to Table 1 of CoralBlox paper)?

docs/src/usage/analysis.md
```diff
@@ -570,6 +588,40 @@ save("outcome_map.png", tf)
...
+  - Only scenarios with at least 0.1 fogging effectiveness are in the top 30% of the range.
+- SRM: 
+  - As expected, coral cover increases with shade (decreases with DHW) in both groups of scenarios.
+  - In the top 30% of the range, the scenario with the lowest level of shade still has about 4.5 DHW reduction.
+  - In the top 50% of the range, the scenario with the lowest level of shade still has about 2.5 DHW reduction.
+  - There are very few scenarios in the top 50% with shade giving reductions of less than 3 DHW - there is very little uncertainty in the mean.
+
+For this test dataset, according to this analysis:
+
+1) **Ensuring conditions for success**: Shade (SRM) is dominating the analysis. To be in the top 30% of the range, shade with at least ~4.5 DHW reduction is necessary.  Fogging with an effectiveness of at least 0.1 is also needed.
```
Zapiano, line 618:
```suggestion
1) **Ensuring conditions for success**: Shade (SRM) is dominating the analysis. To have an outcome above 30% of the range, shade with at least ~4.5 DHW reduction is necessary.  Fogging with an effectiveness of at least 0.1 is also needed. 
```

docs/src/usage/analysis.md
```diff
@@ -570,6 +588,40 @@ save("outcome_map.png", tf)
...
+- SRM: 
+  - As expected, coral cover increases with shade (decreases with DHW) in both groups of scenarios.
+  - In the top 30% of the range, the scenario with the lowest level of shade still has about 4.5 DHW reduction.
+  - In the top 50% of the range, the scenario with the lowest level of shade still has about 2.5 DHW reduction.
+  - There are very few scenarios in the top 50% with shade giving reductions of less than 3 DHW - there is very little uncertainty in the mean.
+
+For this test dataset, according to this analysis:
+
+1) **Ensuring conditions for success**: Shade (SRM) is dominating the analysis. To be in the top 30% of the range, shade with at least ~4.5 DHW reduction is necessary.  Fogging with an effectiveness of at least 0.1 is also needed. 
+2) **Avoiding failure**: In some DHW scenarios, none of the sampled interventions are able to achieve performance in the top 30% of the range. Whether this is a problem depends on what those scenarios represent.
```
Zapiano, line 619: Same as above

Zapiano (approved):
@josephguillaume Thanks for the work! I've added a few comments for your consideration.

docs/src/usage/analysis.md
josephguillaume: To be able to interpret the results we do actually need to know what domain data is loaded, which is why I wanted to be more specific.
Given it's the test dataset, I understand scenarios 1-50 don't actually have a specific meaning - but usually they would, and it is important to know they don't in this case.

How about:
```suggestion
    :dhw_scenario, # All DHW scenarios in the domain data, here scenarios 1-50 in the test dataset
    :wave_scenario, # All wave scenarios in the domain data, here scenarios 1-50 in the test dataset
```

docs/src/usage/analysis.md
josephguillaume: There's an issue that we need to be clear about 30% of what. It's not necessarily 30% coral cover given it's a  min-max normalisation.
My current attempt was to say top 50% or 30% of the range.
An alternative would be to say top 30% of the maximum. That's not technically correct if the minimum is not zero, but might still make more sense?

docs/src/usage/analysis.md
josephguillaume: How about:
```suggestion
- Wave scenario: similar to DHW, this is a categorical factor with each scenario interpreted individually
```

My intent was to describe how it would be interpreted without giving an interpretation given that the current version of ADRIA does not model effect of waves.

docs/src/usage/analysis.md
ConnectedSystems:
```suggestion
The figure shows the mean values of total absolute cover (TAC) obtained as a function of each factor value, across scenarios in which outcomes in the top 50% or 30% of the range are obtained. The ribbon shows the estimated 95% confidence interval around the mean.
```

I think what you have here is fine. I wouldn't worry about agonising over the wording so much here as the method used to identify the scenario outcome of interest is user defined. It's clear (to me at least, for the purpose of demonstration) what the intent is.

ConnectedSystems (approved):
This all looks good to me, thanks!

ConnectedSystems:
>The method does not actually calculate factor regions. Through the use of bootstrapping, it actually calculates the confidence interval of the mean.

Agreed - I mentioned the need to revert back to plotting scattergram-style plots (or similar) - the initial intention here was to limit the amount of data being passed back and forth for visualization purposes (e.g., cloud-based workflows where the visual elements are serialised to JSON and passed client-side), but I'm not so concerned about that any more.

PR #1087, opened 2026-05-11
Add support for loading calibrated coral parameters

Zapiano:
Calibrated parameter loading: load_domain now accepts a calib_params_fn::String="" keyword argument pointing to a calibrated parameters NetCDF file (produced by the CoralBlox calibration workflow). When provided, the Coral and GrowthAcceleration structs are rebuilt with the calibrated values before the domain model is constructed. 
Uses Base.invokelatest to correctly resolve the newly defined constructors across Julia's world age boundary introduced by the internal eval calls.

NetCDF format: The expected file contains six variables — linear_extension, mb_rate, dist_mean (shaped functional_group × size_class), linear_extension_scale, mb_rate_scale (shaped functional_group × cb_calib_group), and growth_acceleration (shaped cb_calib_group × accel_param).


Zapiano:
@ConnectedSystems FYI

PR #1088, opened 2026-05-11
Rule induction documentation

josephguillaume:
As part of work on scenario discovery, I've added interpretation to the rule induction functionality.

Clarifications: 

- The `target_clusters` function is key here and I hadn't understood what it does. I have tried to clarify the API docs.

Added:

- Interpretation of the rule induction plots, including in terms of density and coverage of the rules obtained.

Out of scope:

- Temporal variability is not an intuitive performance metric and I'm not sure why it's the default metric for `target_clusters`, but I think the interpretation still works for the time being.
- I suspect that the 1% default size limit in `target_clusters` results in 1 out of the 6 clusters being selected. I haven't investigated this.
- I didn't perform a detailed review of the implementation of SIRUS given that it uses an external package.

### DIFF
--- a/docs/src/usage/analysis.md
+++ b/docs/src/usage/analysis.md
@@ -467,8 +467,11 @@ save("tsc_map.png", tsc_map_fig)
 
 ### Rule Induction (using Series Clusters)
 
-After clustering, it is possible to target some specific scenarios based on each cluster
-median outcome temporal variability:
+The SIRUS Rule Induction algorithm ([Bénard et al. 2021](https://doi.org//10.1214/20-EJS1792)) can be used for scenario discovery by summarising scenarios in terms of binary rules, 
+i.e. thresholds below/above which a factor will lead to a specified outcome.
+
+For this example, we cluster scenarios with similar total cover, and then focus on those with high temporal variability in total cover. We explore what intervention characteristics lead to high temporal variability.
+
 
 ```julia
 # Find Time Series Clusters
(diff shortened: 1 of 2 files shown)


ConnectedSystems:
> The target_clusters function is key here and I hadn't understood what it does.

From memory, it is a user-provided "rule" that delineates the behavioral and non-behavioral outcomes.

>Temporal variability is not an intuitive performance metric and I'm not sure why it's the default metric for target_clusters, but I think the interpretation still works for the time being.

This is because the primary use case for SIRUS originally was to identify the scenario time series of interest as relying on a summary statistic can be misleading

> I suspect that the 1% default size limit in target_clusters results in 1 out of the 6 clusters being selected. I haven't investigated this.

Hmmm, thanks for the flag!

ConnectedSystems (approved):

josephguillaume:
> From memory, it is a user-provided "rule" that delineates the behavioral and non-behavioral outcomes.

`find_scenarios` has a user provided rule/function. `target_clusters` just picks the top clusters according to the user supplied metric. That's effectively a rule too, just a less flexible one.

> This is because the primary use case for SIRUS originally was to identify the scenario time series of interest as relying on a summary statistic can be misleading

No objection to that. From a scenario discovery perspective, it's still saying that scenarios will be selected based on their temporal variability, and the results show this is probably not what an analyst would want. High temporal variability is likely associated with large declines. Low temporal variability seems to be associated with cases that are persistently low and never improve. In both cases other metrics would capture these behaviours more directly.

PR #1093, opened 2026-05-18
Simplified interface for results-only visualization

ConnectedSystems:
Improve compatibility of the visualization method when plotting results from ReefModEngine.jl, starting only with the scenario time series method.

Enables use patterns like:

```julia
s_tac = ADRIA.metrics.scenario_total_cover(rs)
ADRIA.viz.scenarios(s_tac)
```

### DIFF
--- a/ext/AvizExt/theme.jl
+++ b/ext/AvizExt/theme.jl
@@ -7,6 +7,8 @@ const COLORS::Dict{Symbol,Union{Symbol,String}} = Dict(
     :RCP45 => :darkblue,
     :RCP60 => :seagreen,
     :RCP85 => :orangered,
+    :scenarios => :steelblue,
+    :interventions => :steelblue,
     :counterfactual => :red,
     :unguided => :lawngreen,
     :guided => :dodgerblue,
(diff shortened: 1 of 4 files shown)


Zapiano:
I'm not sure if this is expected behaviour or not, but using ADRIA results I get a different plot uwing `scenarios(rs)` and `scenarios(rs, outcomes)`:

```
samples = ADRIA.sample(dom, 32)
rs = ADRIA.run_scenarios(dom, samples, "45")
stac = ADRIA.metrics.scenario_total_cover(rs)
```

`ADRIA.viz.scenarios(rs, stac)`
image

`ADRIA.viz.scenarios(stac)`
image

ConnectedSystems:
Ah, with ReefModEngine.jl I am doing:

```julia
scen_groups = ReefModEngine.scenario_groups(result_store)
f = ADRIA.viz.scenarios(s_tac, scen_groups)
```

I guess for ADRIA, we should have an equivalent helper inside the results-only function to do that...

Zapiano (approved):

Zapiano:
> I guess for ADRIA, we should have an equivalent helper inside the results-only function to do that...

Yes, we would need to save the scenario groups in the outcomes YAXArrays metadata to be retrieved by the viz function later.

PR #1102, opened 2026-05-29
Optimise I/O, fix data race in scenario runner, and shrink log storage

Zapiano:
- Shrink log storage: switched cover logs to UInt16 and load logs to Float64, and moved scenario metadata to column headers only. This reduces ResultSet disk footprint significantly for large runs.
- Fix data race in parallel scenario runner: replaced thread-local functional group buffers with a Channel-based pool so each thread checks out and returns its own buffer safely. Previously, concurrent writes to shared buffers produced silently wrong results under Threads.@threads.
- Optimise combine_results: parallelised all Zarr reads across result sets with Threads.@threads, switched outcome writes to a single batched 4-D write per result set (eliminating repeated read-modify-write cycles on shared chunks), and replaced the O(N²) sim-constants validation with a first-vs-last check. Together, these cuts substantially reduce wall time when combining large numbers of result sets.
- Increase DHW heat-tolerance upper bound.

### DIFF
--- a/src/ecosystem/corals/Corals.jl
+++ b/src/ecosystem/corals/Corals.jl
@@ -2,7 +2,7 @@ using DataFrames
 import ModelParameters: Model
 
 # Upper bound offset to use when re-creating critical DHW distributions
-const HEAT_UB = 8.0
+const HEAT_UB = 24.0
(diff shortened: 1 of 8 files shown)


Zapiano:
@ConnectedSystems FYI

PR #1123, opened 2026-06-16
fix: support dynamic year axis in Plotly backend

DanTanAtAims:
The Plotly backend in ADRIAviz always hardcoded a 2025 base year offset for the x-axis, shifting actual calendar years (like 2015–2100) by 10 years to 2025–2110.

This PR updates `_year_ticks` to check if the axis values are relative timesteps (e.g. 1 to n) or actual calendar years:
- **Relative timesteps (e.g. < 1900):** Default to the 2025 base year offset and issue a warning log.
- **Calendar years (e.g. >= 1900):** Use the dataset's coordinate years directly on the x-axis.

This is an image using PlotlyJS, noting that other x-axis is probably more readable as I am having some difficulties getting wsl to work with backends.

testing


ConnectedSystems (approved):
Will merge once I have things sorted in the target branch

PR #1168, opened 2026-08-04
Load calibrated coral parameters into ReefMod and RME domains

Zapiano:
## Summary

`calib_params_fn` was only honoured by `ADRIADomain`, so ReefMod and RME datapackages always fell back to ADRIA's default coral parameters. All three domain loaders now accept it, sharing a single loading path in `src/io/calib_params.jl`.

Also adds `dist_std` to the parameters read from a calibration file. The CoralBlox calibration writes it per functional group and size class, and ADRIA consumes it that way, but the loader was silently dropping it.

```julia
dom = ADRIA.load_domain(
    RMEDomain, path, "45"; calib_params_fn=".../calibrated_params.nc"
)
```

## Backwards compatibility
Non-breaking, hence the patch bump to 0.17.1. A minor bump would have locked ADRIAviz and ADRIAanalysis out — both pin ADRIA = "0.17", which under Julia's compat semantics excludes 0.18. calib_params_fn defaults to "" everywhere, so existing call sites are unaffected.

## Testing
Full suite passing locally. test/calib_params_loading.jl covers dist_std (including that it stays resolved per size class) and loading ReefModDomain / RMEDomain with calibrated parameters. The mock fixture gains dist_std; the old fixture is kept as mock_calibrated_params_no_dist_std.nc to cover the fallback for calibration files that predate the parameter.

## Also
Docstrings across the three ExtInterface files had drifted from their signatures — missing arguments, wrong return types, and two timeframe defaults documented as (2022, 2100) that are actually nothing. Corrected.


Zapiano:
@ConnectedSystems FYI

